PR #333, opened 2022-09-03
Improve the web viewer UI [WIP]

ethanweber:
**Features for this PR:**
- [x] remove dat.gui
- [x] keep Leva for now but move to a panel
- [x] create panels, including one for creating camera paths (as a template)
- [x] moved all CSS to one file `index.scss`
- [x] make the camera field of view property work
- [x] make scene elements easy to toggle off
- [x] modal that pops up when websocket_url isn't specified in the URL

**This is what the UI looks like at the moment:**
image

### DIFF
--- a/nerfactory/viewer/server/server.py
+++ b/nerfactory/viewer/server/server.py
@@ -95,9 +95,6 @@ class WebSocketHandler(tornado.websocket.WebSocketHandler):  # pylint: disable=a
         print("opened:", self, file=sys.stderr)
         self.bridge.send_scene(self)
 
-    def on_close(self):
-        print("WebSocket closed")
-
     async def on_message(self, message: bytearray):  # pylint: disable=invalid-overridden-method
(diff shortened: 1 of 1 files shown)


smart4654154:
Thank you for developing the viewer,
Now I'm changing the range of the mouse scroll wheel(https://github.com/nerfstudio-project/nerfstudio/issues/2980)
This should be  simple. it seems that only min and max need to be changed.
But I can't find the path to the web viewer UI file in nerfstudio1.0.0
The web viewer UI file seems to be in nerfstudio \ nerfstudio \ viewer_legacy \ app \ src \ modules \ Scene \ Scene.jsx, but my modifications have no effect. After deleting the web viewer UI file(nerfstudio \ nerfstudio \ viewer_legacy \ app), the view can still run.
Can you tell me the correct web viewer UI file path and the file to modify the scroll wheel range?. Thank you very much.

PR #342, opened 2022-09-05
Auto orient poses

tancik:
Automatically choose up direction when using a "nerfactory" dataset.
There are two methods for orientation,
pca: computes PCA of location and aligns eigenvectors with axis
up: sets the average up direction to up

### DIFF
--- a/nerfactory/cameras/utils.py
+++ b/nerfactory/cameras/utils.py
@@ -18,7 +18,7 @@ Camera transformation helper code.
 """
 
 import math
-from typing import Tuple
+from typing import Literal, Tuple
 
 import numpy as np
 import torch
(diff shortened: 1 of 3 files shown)


jingyibo123:
Question @tancik : 
Apart from more convenient visualization, are there performance-related purposes for auto-orient the poses? 

p.s. Thanks for the great work and the explanation

tancik:
Purely convenience. However there is a chance that performance will change a little. If the alignment of the geometry (walls, floor), is closely aligned to the grid, the hash encoding may produce better results. However this is not something we optimize for.

IshanArya:
Hi @tancik , I've been reading through the auto orient code to try to understand how the poses are aligned. After reading the above discussion, I am confused.

Do the camera poses not have to be aligned? And if so, how can the position of the cameras relate to each other if each camera is operating in its own world space?

I may be misunderstanding something, but from what I read, each pose comes from the camera extrinsic, which transforms world coordinates to camera coordinates. So how do the cameras know where they are relative to each other?

Thanks in advance for any clarification!

brentyi:
> I may be misunderstanding something, but from what I read, each pose comes from the camera extrinsic, which transforms world coordinates to camera coordinates. So how do the cameras know where they are relative to each other?

For automatic orientation, all of the cameras are treated as a single rigid group and translated/rotated together, so they won't move relative to each other. You can also think of this as re-defining the world coordinate system.

PR #550, opened 2022-09-26
Snap to cameras

origamiman72:
Quality of life improvement: added camera selection and snapping for training cams. Doesn't work for render cams yet!

cameraSelection


origamiman72:
Last commit adds request from #695 , you can rotate the camera so that "up" looks correct and then transform the coordinate system to match within the camera panel.

tancik (approved):
LGTM

PR #586, opened 2022-09-29
tensorf

terrancewang:


nerfstudio/fields/tensorf_field.py
```diff
@@ -0,0 +1,138 @@
...
+            )
+        else:
+            self.B = nn.Linear(in_features=self.color_encoding.get_out_dim(), out_features=appearance_dim, bias=False)
+
+        self.field_output_rgb = RGBFieldHead(in_dim=self.mlp_head.get_out_dim(), activation=nn.Sigmoid())
+
+    def get_density(self, ray_samples: RaySamples):
+        positions = SceneBox.get_normalized_positions(ray_samples.frustums.get_positions(), self.aabb)
+        if self.spatial_distortion is not None:
+            positions = self.spatial_distortion(positions)
```
tancik, line 89: Have you tested with spatial distortions? If not maybe remove it for now. I think there is probably a bug here since the positions are normalized in the line above.

nerfstudio/models/tensorf.py
```diff
@@ -0,0 +1,282 @@
...
+
+        # the callback that we want to run every X iterations after the training iteration
+        def reinitialize_optimizer(
+            self, training_callback_attributes: TrainingCallbackAttributes, step: int  # pylint: disable=unused-argument
+        ):
+            resolution = self.upsampling_steps.pop(0)
+
+            # upsample the position and direction grids
+            # TODO(ethan): ask Brent how to get typing to work on this... the Encoding base class type
+            # in NeRFField is causing the issue
```
tancik, line 121: Is this still an issue?

nerfstudio/data/dataparsers/blender_dataparser.py
```diff
@@ -88,7 +88,9 @@ def _generate_dataparser_outputs(self, split="train"):
 
         # in x,y,z order
         camera_to_world[..., 3] *= self.scale_factor
-        scene_box = SceneBox(aabb=torch.tensor([[-1.5, -1.5, -1.5], [1.5, 1.5, 1.5]], dtype=torch.float32))
+        scene_box = SceneBox(
+            aabb=torch.tensor([[-0.6585, -1.1833, -0.4651], [0.6636, 1.1929, 1.0512]], dtype=torch.float32)
```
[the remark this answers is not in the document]
terrancewang: Oh sorry yeah forgot to remove that. The tighter box improves results by about 2 PSNR, but in the original paper I think they list a method to tighten the bounding box mid-training that I didn't implement.

nerfstudio/models/tensorf.py
terrancewang: I think this should be fine with the new TensoRFField

nerfstudio/fields/tensorf_field.py
terrancewang: Yeah tested with scene contraction and it seems to not affect PSNR but slightly improves rays / sec on the lego scene. Probably not acting as intended with the normalization

terrancewang: Yeah tried it with scene contraction and no change to PSNR, probably not doing what it's intended to with the normalization

tancik (approved):
LGTM!

PR #610, opened 2022-10-02
Camera Extrinsics Optimization

akristoffersen:
Pose optimization, with the first technique being added being BARF.

Currently only works on every model without custom cuda code (everything besides Instant-NGP)

### DIFF
--- a/nerfstudio/cameras/camera_optimizers.py
+++ b/nerfstudio/cameras/camera_optimizers.py
@@ -69,9 +69,9 @@ class CameraOptimizer(nn.Module):
     ) -> TensorType["num_cameras", 3, 4]:
         """Indexing into camera adjustments.
         Args:
-            indices (TensorType["num_cameras"]): indices of Cameras to optimize.
+            indices: indices of Cameras to optimize.
         Returns:
-            TensorType["num_cameras", 3, 4]: Tranformation matrices from optimized camera coordinates coordinates
+            Tranformation matrices from optimized camera coordinates coordinates
             to given camera coordinates.
         """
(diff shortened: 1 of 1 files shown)


f-dy:
Hi @akristoffersen, the docstring for camera_optimizers.py says "Pose and Intrinsics Optimizers", but it seems like only pose is optimized. Did you try optimizing the intrinsics as well, as in Nerf--?

akristoffersen:
hi @f-dy, we never ended up trying camera intrinsics optimization. If you check https://github.com/nerfstudio-project/nerfstudio/blob/686e995dd6e8d65daf121b4ee51e963c67173fcd/nerfstudio/cameras/cameras.py#L314, there is some code for optimizing distortion parameters, though it is not fully implemented.

Feel free to experiment with this yourself, or feel free to add a feature request PR to see if someone would like to try to implement this.

f-dy:
Well we're already having major issues with the current (pose-only) optimization, which we found was worsening results whenever the SfM was good enough. Decreasing the camera LR (x0.1) fixes that, but then the cameras are almost not optimized.
How was the default LR set?
Did you also observe that it's hard to get a camera LR that works fine with all scenes (at least when starting from ground truth camera poses)?

akristoffersen:
We did notice a drop in all metrics when using pose optimization on only the training poses, though the rendered results were clearly sharper. This was mitigated if the evaluation poses were optimized at test-time. The LR may also require further tuning, we had tuned the LR back when pose optimization was added, but surely could use retuning as the nerfacto model has had additional development.

I would not expect camera pose optimization to improve the scene if the poses are already pretty good or GT such as in synthetic scenes.

Issue #579, opened 2022-09-29
dataparser interfering with --vis list


ethanweber:
So this is a minimal command to run the viewer with different data `nf-train proposal --vis viewer nerfactory-data --pipeline.datamanager.dataparser.data-directory data/nerfactory/desolation --pipeline.datamanager.dataparser.downscale-factor 4`. However, it throws the error `nf-train proposal: error: argument --vis: invalid choice: 'nerfactory-data' (choose from 'viewer', 'wandb', 'tensorboard')` because it thinks `nerfactory-data` is part of the `--vis` list. Any ideas here @brentyi?

PR #628, opened 2022-10-03
use literal instead of list for --vis

ethanweber:
- now we can only specify one of viewer, wandb, or tensorboard. we don't think the user will want to use more than one very often. if they do, they can hack around things to change logic or submit a PR fix
- fixes issue #579 because there was no good way around this issue as of now


RafailFridman:
Hi!
I'm a user that wants to have the viewer and wandb together :) 
Could you please help me/ point in the direction of what to change in the code to allow this?

PR #648, opened 2022-10-04
Cameras to TensorDataclass

jake-austin:
Cameras is now a tensordataclass and is about as generalized as I think we can get before launch. Updated tensordataclass to make this happen, keeping a list of fields that will have custom tensor sizes before the batch dimensions. Had to bug fix some issues with TensorDataclass in general even before making these modifications.

ns-train and ns-eval appear to work just like normal, performance seems to also be at normal

The one bug I can't fix is that the input images don't show up on the viewer anymore, and I'm not sure how to fix this at this point.

Will do one last sweep tomorrow to make sure that nobody is accessing any of the fields in an outdated way, but should be fine given that everything works as normal pending the viewer bug


nerfstudio/cameras/camera_paths.py
```diff
@@ -76,11 +76,11 @@ def get_spiral_path(
     else:
         raise ValueError("Only one of radius or radiuses must be specified.")
 
-    up = camera.camera_to_worlds[0, :3, 2]  # scene is z up
-    focal = torch.min(camera.fx[0], camera.fy[0])
+    up = camera.flatten().camera_to_worlds[0, :3, 2]  # scene is z up
```
ethanweber, line 79: I don't think flatten and ravels are needed in this function.

nerfstudio/cameras/rays.py
```diff
@@ -135,7 +135,7 @@ class RayBundle(TensorDataclass):
     """Unit ray direction vector"""
     pixel_area: TensorType[..., 1]
     """Projected area of pixel a distance 1 away from origin"""
-    camera_indices: Optional[TensorType[..., 1]] = None
+    camera_indices: Optional[TensorType[..., "num_camera_batches"]] = None
```
ethanweber, line 138: I think this can stay 1.

nerfstudio/data/utils/dataloaders.py
```diff
@@ -48,13 +48,15 @@ def __init__(
         num_images_to_sample_from: int = -1,
         num_times_to_repeat_images: int = 0,
         device: Union[torch.device, str] = "cpu",
+        collate_fn=default_collate,
```
ethanweber, line 51: Can update the docstring with this change.

tests/cameras/test_cameras.py
```diff
@@ -1,12 +1,19 @@
...
 
 import torch
 
 from nerfstudio.cameras.cameras import Cameras
 from nerfstudio.cameras.rays import RayBundle
 
+# # Required imports for the performance cameras debugging
+# from time import perf_counter
+# import numpy as np
+
```
ethanweber, line 15: remove

tests/cameras/test_cameras.py
```diff
@@ -22,5 +29,260 @@ def test_pinhole_camera():
...
+        camera_to_worlds[1],
+        fx_ys[0],
+        fx_ys[0],
+        cx_ys[0],
+        cx_ys[0],
+        h_ws[0],
+        h_ws[0],
+        distortion_params[0],
+        camera_types[0],
+    )
```
ethanweber, line 82: delete

tests/cameras/test_cameras.py
```diff
@@ -22,5 +29,260 @@ def test_pinhole_camera():
...
+    fx_ys = [
+        10.0,
+        torch.tensor(1).float() * 10,
+        torch.ones(batch_size) * 10,
+        torch.ones(batch_size, 1) * 10,
+        torch.ones((batch_size, batch_size)) * 10,
+    ]
+    h_ws = [
+        None,
+        800,
```
ethanweber, line 50: move these values to top

tests/cameras/test_cameras.py
```diff
@@ -22,5 +29,260 @@ def test_pinhole_camera():
...
+    batch_size = 2
+    # pylint: disable=unnecessary-comprehension
+    c2w_flat = torch.eye(4)[:3, :]
+    camera_to_worlds = [
+        c2w_flat,
+        torch.stack([c2w_flat] * batch_size),
+        torch.stack([torch.stack([c2w_flat] * batch_size)] * batch_size),
+    ]
+    fx_ys = [
+        10.0,
```
ethanweber, line 42: move to top

tests/cameras/test_cameras.py
```diff
@@ -22,5 +29,260 @@ def test_pinhole_camera():
...
+    _ = c2_dist[torch.tensor([0]), torch.tensor([0])]
+
+    assert c0.shape == ()
+    assert c1[...].shape == torch.Size([batch_size])
+
+    assert _check_cam_shapes(c0, ())
+    assert _check_cam_shapes(c1, (2,))
+    assert _check_cam_shapes(c2_dist, (2, 2))
+
+    assert c0.generate_rays(0).shape == (800, 800)
```
ethanweber, line 141: move to top

tests/cameras/test_cameras.py
```diff
@@ -22,5 +29,260 @@ def test_pinhole_camera():
...
+    c1.generate_rays(0)
+    c1.generate_rays(torch.tensor([0, 1]).unsqueeze(-1))
+
+    # Make sure rays generated are same when distortion params are identity (all zeros) and None
+    assert c2.shape == c2_dist.shape
+
+    c2_rays = c2.generate_rays(torch.tensor([0, 0]))
+    c_dist_rays = c2_dist.generate_rays(torch.tensor([0, 0]))
+    assert _check_dataclass_allclose(c2_rays, c_dist_rays)
+    assert c2_rays.shape == (800, 800)
```
ethanweber, line 152: move to top

ethanweber:
Checklist
- [ ] remove ravel from the code
- [ ] speed up the test cases (they are taking over 4 minutes locally compared to 30 sec on main)
- [ ] make the code work for variable height and width

nerfstudio/utils/tensor_dataclass.py
```diff
@@ -90,9 +103,13 @@ def _get_dict_batch_shapes(self, dict_: Dict) -> List:
             The batch shapes of all tensors in the dictionary.
         """
         batch_shapes = []
-        for v in dict_.values():
+        for k, v in dict_.items():
             if isinstance(v, torch.Tensor):
-                batch_shapes.append(v.shape[:-1])
+                if isinstance(self._field_custom_dimensions, dict) and k in self._field_custom_dimensions.keys():
+                    # pylint: disable=unsubscriptable-object
+                    batch_shapes.append(v.shape[: -self._field_custom_dimensions[k]])
```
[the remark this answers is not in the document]
jake-austin: Just to make sure it's on record even though we talked offline about this: should be the final n dimensions that we are looking to exclude, so we index until the nth element from the end, just like when the field is a tensor with only 1 data dimension, we do v.shape[-1] (see the else case for this if statement).

lkskstlr:
Hi @jake-austin, @tancik

Thank you all for the amazing package and I know that your TODO lists are probably overflowing like crazy. I just wanted to ask what the timeline on this PR is and if it already works with adapted `transforms.json` (see https://github.com/nerfstudio-project/nerfstudio/issues/868). If not, no problem :)

ethanweber:
Hey @lkskstlr, we are refactoring the dataloader and then we will merge PR #961. It's quite a big change, so we just want to make sure it doesn't break anything. I'll close this PR since #961 implements everything here and has a test with Phototourism data.

PR #671, opened 2022-10-05
Installation Docs Updates

ravirahman:
Overall awesome work!

In this PR.

1. Fixed a small typo in the README.md
2. Added commends to upgrade pip and setuptools into the "install from source" sections

I wasn't able to actually run anything because I didn't have a GPU

Other comments -- feel free to take or leave :)

1. It doesn't look like conda is used for any dependencies, so I'd remove that from the README.md, and instead recommend to use pip / virtualenv directly.
2. Add maximum versions to any range dependencies in the pyproject.toml (e.g. h5py, appdirs, dcargs, ...) or use strict version pinning. Also, `requests` should have a version.
3. All the `python scripts/train.py` commands don't work -- getting errors along the lines of:
> `train.py: error: unrecognized arguments: --config-name=graph_instant_ngp.yaml`
4.  Thoughts on renaming the default branch to `main`? See https://github.com/github/renaming.
5. If you have a docker image with everything preinstalled, can you release that? It'd be a really nice one-liner to have to train an example, start the visualiezer, etc...

6. https://docs.nerf.studio/en/latest/quickstart/viewer_quickstart.html is blank

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ For more complex changes (e.g. running with your own data/ setting up a new NeRF
 We reccomend using conda to manage dependencies. Make sure to install [Conda](https://docs.conda.io/en/latest/miniconda.html) before preceding.
 
 ```bash
-conda create --name nerfstudio -y python=3.8;
+conda create --name nerfstudio -y python=3.8
 conda activate nerfstudio
 python -m pip install --upgrade pip
 ```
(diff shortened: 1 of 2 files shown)


ethanweber (approved):
Thanks for the feedback. We will take this into consideration and update!

tancik (approved):

PR #693, opened 2022-10-05
Fix viewer slider stepping past min/max limits

brentyi:
Just fixing two corner cases I noticed:
1. If you press the "next frame" button when there are no cameras added, the slider currently goes negative
1. If you press the "prev frame" button after adding two cameras with the slider at 0, the viewer currently crashes


tancik (approved):
Awesome, thanks for the fixes!

PR #697, opened 2022-10-05
Remove outdated viewer logging

tancik:
Remove logging in terminal for connecting to localhost. This is no longer necessary as users do not host the viewer.


ethanweber (approved):
LGTM!

PR #699, opened 2022-10-05
max number of iterations logic

ethanweber:
- [x] Pause training and keep rendering after done training, when using the viewer
- [x] Updated max number of iterations to be 30K for nerfacto and instant-ngp


tancik (approved):
LGTM

PR #719, opened 2022-10-06
Reduced `tyro` overhead + helptext formatting

brentyi:
Just released `tyro` 0.3.23, which does some more aggressive caching and reduces the config overhead significantly. On my machine, parsing the full config library now takes ~0.15 seconds each time the `ns-train` script is run (down from ~0.45).
It's dramatically lower for the other scripts. (on the order of 1e-4 seconds)

A few changes:
- Bumped `tyro` dependency.
- Removed ANSI sequences from subcommand helptext. (these are unfortunately no longer supported)
- Copied the install scripts walkthrough from the docs to the help message + made some minor typo fixes and edits.

image


tancik (approved):
Awesome, thanks!

PR #726, opened 2022-10-07
Pose Opt Doc typo

akristoffersen:


brentyi (approved):
thanks for catching this!!

PR #744, opened 2022-10-09
bump nerfacc to 0.2.0. bug fix for depth vis.

liruilong940607:
- Bump nerfacc version to 0.2.0
- Fix depth visualization when the samples is packed.

NerfAcc 0.2.0 v.s. 0.1.3: Slight optimization on the library side. More util functions are supported. 

Screen Shot 2022-10-09 at 11 28 09 AM

image (6)

The huge performance improvements in NerfAcc 0.2.0 comes mainly from the better logic for occupancy grid updating. Not sure how to incorporate that which seems to need some refactor of the nerfstudio code. Left a comment in the code.

### DIFF
--- a/nerfstudio/model_components/ray_samplers.py
+++ b/nerfstudio/model_components/ray_samplers.py
@@ -477,7 +477,7 @@ class VolumetricSampler(Sampler):
             starts = torch.ones((1, 1), dtype=starts.dtype, device=rays_o.device)
             ends = torch.ones((1, 1), dtype=ends.dtype, device=rays_o.device)
 
-        ray_indices = nerfacc.unpack_to_ray_indices(packed_info).long()
+        ray_indices = nerfacc.unpack_info(packed_info)
         origins = rays_o[ray_indices]
         dirs = rays_d[ray_indices]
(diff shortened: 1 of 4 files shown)


liruilong940607:
Another cool thing in nerfacc=0.2.0, is that you can set an `alpha_thre` to `ray_marching`,  which explicitly skips the empty space. This is more accurate than using occupancy grid to do the skipping, as the occupancy grid stores the **maximum** value for each cell. For contracted occupancy grid where a cell can represent a large space, this value can be quite off than the actual opacity queried from the radiance field.

Simplly setting `alpha_thre=1e-2` to `ray_marching` gives noticeable speedup & performance boost (white is with alpha_thre)
Screen Shot 2022-10-09 at 11 45 02 AM

liruilong940607:
All the unbounded results shown on the nerfacc webpage are using alpha_thre=1e-2. So I think that can be safely set as default. Worth to note the occupancy grid also have by default 1e-2 threshold so `alpha_thre` is kinda like double check using network's outputs

tancik (approved):
LGTM, thanks for the work in nerfacc!

PR #747, opened 2022-10-10
Google Colab support with viewer

ethanweber:
- [x] ability to connect to viewer bridge server running on a different machine
- [x] create ns-bridge-server command
- [x] use ngrok to let people run the server locally and expose publicly
- [x] removed `logging` and now use `rich`
- [x] here is the colab link https://colab.research.google.com/drive/1l-QDYEcv759kUNJQrJ8z1CI8o1iIiU-l?usp=sharing

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,6 @@
 
 <img src="https://user-images.githubusercontent.com/3310961/194017985-ade69503-9d68-46a2-b518-2db1a012f090.gif" width="52%"/> <img src="https://user-images.githubusercontent.com/3310961/194020648-7e5f380c-15ca-461d-8c1c-20beb586defe.gif" width="46%"/>
 
-
 - [Quickstart](#quickstart)
 - [Learn more](#learn-more)
 - [Supported Features](#supported-features)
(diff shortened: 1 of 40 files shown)


ethanweber:
Webrtc is connecting but not transmitting any information right now. I suspect it has something to do with [asyncio](https://docs.python.org/3/library/asyncio.html) being used improperly.

README.md
```diff
@@ -78,7 +77,7 @@ For more complex changes (e.g., running with your own data/setting up a new NeRF
 We recommend using conda to manage dependencies. Make sure to install [Conda](https://docs.conda.io/en/latest/miniconda.html) before proceeding.
 
 ```bash
-conda create --name nerfstudio -y python=3.8
+conda create --name nerfstudio -y python=3.8.12
```
tancik, line 80: Can we avoid having such a specific version?

docs/quickstart/google_colab.md
```diff
@@ -0,0 +1,24 @@
+# Google Colab
+
+Here we briefly explain how to get our repo working in a Google Colab.
+
+## Viewer suport
+
+To get the viewer working, you have to do some steps outside the notebook.
+
+#### Installing on Mac M1
```
tancik, line 9: This isn't specific to M1

tancik (approved):

PR #748, opened 2022-10-10
Terrance/camera controls

terrancewang:


terrancewang:
two modes of moving the camera, first person and orbit. toggle between them with either the button or r. in first person, move camera horizontally with wasd, vertically with qe, and turn camera with arrow keys. in orbit, orbit using wasd, zoom in and out using q and e.

terrancewang:
also let me know if any of the move speed / controls seem weird, or if the button names First Person Controls and Orbit Controls seem clear enough.

ethanweber (commented):
Can you update the documentation in the "Getting Started" panel for how to use?

ethanweber:
Is there a way to make it so that when you go forward with w it actually goes forward in view space rather than world space?

terrancewang:
> Is there a way to make it so that when you go forward with w it actually goes forward in view space rather than world space?

Oh yeah I think that would be a good change. will look into it

tancik:
The keyboard controls are great! The mouse controls seem to have changed though, they don't mimic what was existing in the main branch.

terrancewang:
> The keyboard controls are great! The mouse controls seem to have changed though, they don't mimic what was existing in the main branch.

hmm that's weird, are the mouse controls just inversed now? I'll check it out later today

tancik:
> 
They arn't reversed. Rather the pivot point is not at the camera. Moving side-to-side is very slow.
In the example video, this PR is on the left and main is on the right. I first rotate both, then I move the camera in both.

https://user-images.githubusercontent.com/3310961/199856829-59aa540f-e179-4175-a8a1-f133d5bfb056.mov

tancik (approved):
Nice!

Issue #695, opened 2022-10-05
Add manual coordinate frame rotation in viewer.


tancik:
If the auto-orientation methods in the dataparser fails, the scene make look rotated in the viewer. We should add controls in the UI to manually correct for these offsets.

tancik:
Added

PR #749, opened 2022-10-10
Added up vector offset

origamiman72:
Addresses #695 , you can move the camera so that "up" looks correct and then press [Render->Reset Up Direction] and the grid will shift so that up matches the camera. Also adds a vector to display the local up direction once button is pressed.


ethanweber (approved):
Looks good! We can leave #695 open to further improve.

PR #766, opened 2022-10-13
Reduce memory usage of nerfacto model

tancik:
Storing all of the eval sample info during eval would cause spikes in memory. By only storing during training this PR is able to reduce memory usage 2x while also increasing the number of eval chunks by 2x.

### DIFF
--- a/nerfstudio/viewer/server/viewer_utils.py
+++ b/nerfstudio/viewer/server/viewer_utils.py
@@ -639,7 +639,6 @@ class ViewerState:
         check_thread = CheckThread(state=self)
         render_thread = RenderThread(state=self, graph=graph, camera_ray_bundle=camera_ray_bundle)
 
-        oom = False
         with TimeWriter(None, None, write=False) as vis_t:
             check_thread.start()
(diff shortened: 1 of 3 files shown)


nerfstudio/viewer/server/viewer_utils.py
```diff
@@ -639,25 +639,23 @@ def _render_image_in_viewer(self, camera_object, graph: Model, is_training: bool
...
         with TimeWriter(None, None, write=False) as vis_t:
             check_thread.start()
             render_thread.start()
             try:
                 render_thread.join()
                 check_thread.join()
             except IOChangeException:
-                pass
+                del camera_ray_bundle
+                torch.cuda.empty_cache()
```
ethanweber, line 650: You found these changes to help?

tancik: They don't seem to hurt. They clear memory once in awhile.

ethanweber (approved):
ns-render memory is at 2966 MiB for me now. Nice!

PR #773, opened 2022-10-13
Update first_nerf.md

krahets:
Add an example of resuming training, which is a common usage.

### DIFF
--- a/docs/quickstart/first_nerf.md
+++ b/docs/quickstart/first_nerf.md
@@ -55,6 +55,12 @@ Run a nerfacto model with different data and port.
 ns-train nerfacto --vis viewer --data data/nerfstudio/poster --viewer.websocket-port 7007
 ```
 
+Run a nerfacto model and load the latest checkpoint to resume training.
+
+```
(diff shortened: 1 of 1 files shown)


docs/quickstart/first_nerf.md
```diff
@@ -55,6 +55,12 @@ Run a nerfacto model with different data and port.
 ns-train nerfacto --vis viewer --data data/nerfstudio/poster --viewer.websocket-port 7007
 ```
 
+Run a nerfacto model and load the latest checkpoint to resume training.
+
+```
+ns-train nerfacto --vis viewer --data data/nerfstudio/poster --trainer.load_dir {base_dir}/nerfstudio_models
```
tancik, line 61: Maybe we should be more specific

ie,
```
ns-train nerfacto --data data/nerfactory/posterv2/ --trainer.load_dir outputs/data-nerfstudio-poster/nerfacto/{timestamp}/nerfstudio_models
```

tancik (approved):
Thanks, this will helpful for other getting started!

docs/quickstart/first_nerf.md
krahets: Thx! It is clearer. The change has been committed.

PR #776, opened 2022-10-14
Add PyTorch jit-ing to the undistort computation for speed

kerrj:
About 6ms faster per step (somewhere 10-15% speedup overall)


tancik (approved):
Awesome speedup!

PR #777, opened 2022-10-14
Fix tab completion for PyPI installs, add `py.typed`

brentyi:
1. `ns-install-cli` was formerly breaking on machines where `nerfstudio` was installed via PyPI.
2. Need a `py.typed` for type checking if folks are using `nerfstudio` as a library: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/


tancik (approved):
LGTM

PR #778, opened 2022-10-14
Respect custom timestamps passed to `ns-train`

corbt:
When using `ns-train` as part of a pipeline, it's convenient for the output directory to be deterministic. This should be achievable by passing the `--timestamp` argument to `ns-train`, but that argument is currently being overwritten. This PR fixes the logic to respect a custom user-provided `--timestamp` value.


tancik (approved):
LGTM, thanks for the update!

PR #785, opened 2022-10-16
Refine intrinsics with COLMAP

tancik:
Refining the intrinsics in colmap helps improve the final reconstruction sharpness.

Before:
image
After:
image


ethanweber (approved):
LGTM

Issue #789, opened 2022-10-16
Camera view jumps when using side panel


tancik:
**Describe the bug**
Clicking buttons anywhere in the control panel can cause the camera to jump.

**To Reproduce**
1. Zoom in very close to the training data
2. Click on the right panel anywhere
3. The camera jumps

**Expected behavior**
The camera should not jump. I'm guessing the training image is "selected" and when it sees a "click" event it fires.

**Screenshots**
https://user-images.githubusercontent.com/3310961/196059592-67454ca8-9b87-485e-b0bd-8c124a61ab34.mp4

mcallisterdavid:
Ah, sorry about these bugs. This one is an easy fix at least

Issue #788, opened 2022-10-16
Don't view from training view when images are hidden.


tancik:
**Describe the bug**
When the training images are hidden, they can still be selected for viewing from their position.

**To Reproduce**
1. Select hide images button
2. Click where a training image was
3. The Camera jumps position

**Expected behavior**
The camera should not move.

mcallisterdavid:
Gotcha, taking a look

PR #791, opened 2022-10-16
[VIEWER] Camera Select Bug Fixes

mcallisterdavid:
Fixes for #788 and #789. Visibility updates are shallow, so scene objects often have stale visible booleans. This adds a simple traversal up the parent chain to check visibility


tancik (approved):
LGTM, Thanks for the quick fix!

PR #794, opened 2022-10-17
Speedup get_ray_sample

tancik:
Small tweaks to get a 5% speedup on nerfacto. Found with @kerrj


ethanweber (approved):
Nice improvements. You should add @kerrj as a coauthor when you make the merge?

PR #799, opened 2022-10-18
Multi-gpu improvements

ethanweber:
[DDP](https://pytorch.org/tutorials/beginner/dist_overview.html) seems to be working, but it only really makes sense for long or compute-heavy training runs. Here is a small test for only 1K iterations with vanilla-nerf.

```python
# 3 GPU (3x LR)
ns-train vanilla-nerf --machine.num-gpus 3 --vis wandb --trainer.max-num-iterations 1000 --optimizers.fields.optimizer.lr 15e-4 --trainer.steps-per-eval-batch 100 --trainer.steps-per-eval-image 100 --logging.local-writer.max-log-size=0

# 1 GPU (1x LR)
ns-train vanilla-nerf --machine.num-gpus 1 --vis wandb --trainer.max-num-iterations 1000 --optimizers.fields.optimizer.lr 5e-4 --trainer.steps-per-eval-batch 100 --trainer.steps-per-eval-image 100 --logging.local-writer.max-log-size=0
```

image

### DIFF
--- a/docs/quickstart/first_nerf.md
+++ b/docs/quickstart/first_nerf.md
@@ -152,3 +152,23 @@ ns-eval --load-config={PATH_TO_CONFIG} --output-path=output.json
 ```
 
 We also provide a train and evaluation script that allows you to do benchmarking on the classical Blender dataset (see our [benchmarking workflow](../developer_guides/debugging_tools/benchmarking.md)).
+
+## Multi-GPU Training
+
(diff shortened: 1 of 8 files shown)


tancik (approved):
LGTM

ethanweber:
Note that further experiments are needed to verify that the code works as expected. The key would be to find a high learning rate where the model diverges on 1 GPU but works with multiple GPUs.

PR #800, opened 2022-10-18
Add more controls for orienting poses

tancik:
Adding the following three options for the nerfstudio dataset:

`--orientation-method {pca, up, none}` (can now be `none`)
`--center-poses {bool}`
`--auto-scale-poses {bool}`


ethanweber (approved):
LGTM

PR #802, opened 2022-10-18
Support python 3.7

tancik:
We aren't using much 3.8 functionality. This PR lowers the requirement to >=3.7.3 (TorchTyping breaks if you go lower)


ethanweber (approved):
LGTM. Maybe worth updating the README and installation docs.

Issue #798, opened 2022-10-18
Add a "load camera path" option to viewer


tancik:
**Is your feature request related to a problem? Please describe.**
Currently there is no way to load an existing camera path into the viewer.

**Describe the solution you'd like**
We should provide a button in the viewer under the camera path tab that enables users to load existing camera paths.

PR #804, opened 2022-10-19
Add ability to load path from file

origamiman72:
Addresses #798 , allows camera path jsons to be uploaded. Not backwards compatible unless we want old camera path files to load a keyframe for every frame in the video.


tancik (approved):
LGTM!

PR #809, opened 2022-10-19
Mesh export

ethanweber:
## CLI interface

For poisson surface reconstruction
```bash
ns-export poisson --load-config config.yml --output-dir exports
```

For TSDF
```bash
ns-export tsdf --load-config config.yml --output-dir exports
```

Texturing, starting from a mesh file
```bash
python scripts/texture.py --load-config config.yml --input-ply-filename filename.ply --output-dir exports
```

## Method

**UV mapping step**
We have two ways to texture the mesh. Either (1) with [xatlas](https://github.com/jpcy/xatlas) unwrapping or (2) with a unique rectangle per mesh.

For (2), this is what the UV texture image look like. Each mesh face has a unique location.
Screen Shot 2022-11-17 at 8 32 37 PM

**Texturing step**
Now we have a UV mapping per triangle. We take the center of every pixel in the texture image, and we find the corresponding XYZ position and normal vector with barycentric interpolation. Then we render a short ray from the NeRF to set the RGB value.

## Example

Example with the plane scene, with a specified bounding box
```python
ns-export tsdf --load-config outputs/data-nerfstudio-plane/nerfacto/2022-11-18_124018/config.yml --output-dir exports/plane --downscale_factor 8 --resolution 128 --use-bounding-box True --bounding-box-min -0.5 -0.55 -0.25 --bounding-box-max 0.5 0.55 0.15 --px_per_uv_triangle 8
```

## TODOs

- [x] turn off predicting normals by default
- [x] throw descriptive error when running poisson without normal prediction on
- [x] chunking for texturing needs to be smaller (eg 10) to avoid memory issues
- [x] add total times to progress bars
- [x] fix documentation (installation details, etc.)
- [x] argument to save out the point cloud when using poisson meshing (since it's a lot of memory)
- [x] cuda clear cache after point cloud creation
- [x] finish message and summary (#points, # faces, ...)
- ~~[ ] fix `downscale_factor` logic for the TSDF code~~


ksnzh (approved):
The error is caused by open3d 0.16.0 having no prebuilt python3.8 package before Oct 27. Now I have tried and it's okey.

kevinddchen:
Oh, is the aim of this PR to create the mesh from the input data? Not from the NeRF density field itself, as done here https://github.com/ashawkey/torch-ngp/blob/4ae16168f69b005bf4c375ff424ef9d81bc397d7/nerf/utils.py#L571?

tancik:
> https://github.com/ashawkey/torch-ngp/blob/4ae16168f69b005bf4c375ff424ef9d81bc397d7/nerf/utils.py#L571

The goal is to extract a mesh from an existing NeRF model. We will probably support different method of extraction (such as from density fields). However our first method will extract a point cloud and convert it to a mesh, we have found that this often leads to better results.

kevinddchen:
I see. I am very interested in mesh from NeRF model. I think mesh from density field is kinda hard, and doesn't look that great most of the time.

I've seen other papers try other representations which seem to be better for surfaces (e.g. SDF? output alphas instead of density?) and I'd love to try those out in `nerfstudio` to see how they are.

Maybe even textured mesh? 👀

scripts/mesh.py
```diff
@@ -0,0 +1,93 @@
...
+    Annotated[ExportTSDFMesh, tyro.conf.subcommand(name="tsdf")],
+    Annotated[ExportPoissonMesh, tyro.conf.subcommand(name="poisson")],
+    Annotated[ExportMarchingCubesMesh, tyro.conf.subcommand(name="marching-cubes")],
+]
+
+
+def entrypoint():
+    """Entrypoint for use with pyproject scripts."""
+    tyro.extras.set_accent_color("bright_yellow")
+    tyro.cli(Commands).main()
```
brentyi, line 89: A structure like this seems like it might be (slightly) simpler:

```python
MesherConfig = Union[
    Annotated[TSDFMesherConfig, tyro.conf.subcommand(name="tsdf")],
    Annotated[PoissonMesherConfig, tyro.conf.subcommand(name="poisson")],
    Annotated[MarchingCubesMesherConfig, tyro.conf.subcommand(name="marching-cubes")],
]

def main(load_config: Path, output_dir: Path, mesher_config: MesherConfig) -> None:
        _, pipeline, _ = eval_setup(load_config)
        mesher = mesher_config.setup()
        mesher.export_mesh(pipeline, output_dir=output_dir)

tyro.cli(main)
```

tancik: I guess the purpose of this convoluted logic is to allow the following syntax
`ns-export tsdf --load-config XX --output-dir XX`

brentyi: Ah clever, makes sense!

scripts/exporter.py
```diff
@@ -85,25 +159,368 @@ def main(self) -> None:
...
+
+        # with CONSOLE.status("Saving Mesh", spinner="pipe") as status:
+        CONSOLE.print("Saving Mesh")
+        o3d.io.write_triangle_mesh(str(output_dir / "mesh.ply"), mesh)
+        print("\033[A\033[A")
+        CONSOLE.print("[bold green]:white_check_mark: Saving Mesh")
+
+
+@dataclass
+class ExportTextureMesh(Exporter):
```
cpheinrich, line 340: Just fyi if you want to do proper texturing I'd suggest using https://github.com/nmoehrle/mvs-texturing -- there are no python bindings so you'd have to ask people to build from source and then point to the build folder, but will probably give you the best results and pretty fast

devernay: Be aware that MVS Texturing sometimes has strange color shifts.
Meshroom can also do texturing, but the the texturing part is a bit harder to run.

cpheinrich: Ok, fwiw, I've never had any issues with color shifts and I've used this a lot. The texturing quality on openMVS is slightly higher: https://github.com/cdcseacave/openMVS, although there are more third party dependencies so if you only need texturing then mvs-texturing is an easier integration. They are all based on this paper: https://link.springer.com/content/pdf/10.1007/978-3-319-10602-1_54.pdf

ethanweber: Thanks @cpheinrich for the pointers!

devernay: Maybe you're right @cpheinrich, this issue was solved since https://github.com/nmoehrle/mvs-texturing/issues/24
However, this is still not optimizing the reprojected images as NeRF does (and tonemapping mode makes it a bit more unreal).

What would better fit in nerfstudio is something where we would get the texture from NeRF (even though it's lower resolution):
1. produce the mesh using some method (can be density or pointcloud or something else)
2. create UVs for the mesh (eg using xatlas)
3. render a geometry image (a render of XYZ positions in UV space) -> this gives you XYZ for each texture point
4. compute texture by evaluating nerf at each 3D point, maybe averaging over directions that are on the outer side of the surface

Before step 4 we could fine-tune the NeRF, as done in [Surface NeRF aka SNeRF](https://arxiv.org/abs/2108.04886) - this is called SNeRF (fixed) in the paper.

Step 4 could be replaced by something where we compute another representation for a surface lightfield, eg spherical harmonics,

ethanweber: @cpheinrich @devernay

I updated the description of the PR with what we are doing for the first pass. Our UV texture image is a "texture atlas" since every face maps to a unique location. We then color the texture image with NeRF.

Averaging over directions (step 4. suggested by @devernay) will likely give better results.

Happy to hear thoughts on this approach and what can be improved. Thanks!

This is the result I'm getting with TSDF + this type of texturing.
https://user-images.githubusercontent.com/16965789/202813416-9d6eeed7-6e7f-4f0d-b2ad-356a460c1787.mov

nerfstudio/exporter/texture_utils.py
```diff
@@ -0,0 +1,275 @@
...
+        + nearby_normals[..., 1, :] * w1[..., None]
+        + nearby_normals[..., 2, :] * w2[..., None]
+    ).to(device)
+    # normalize the direction vector to make it a unit vector
+    directions = torch.nn.functional.normalize(directions, dim=-1)
+
+    origins = origins - offset * directions
+    pixel_area = torch.ones_like(origins[..., 0:1])
+    camera_indices = torch.zeros_like(origins[..., 0:1])
+    nears = torch.zeros_like(origins[..., 0:1])
```
tancik, line 193: I forget, will the near plane get clipped to a larger value elsewhere in the code?

nerfstudio/exporter/texture_utils.py
```diff
@@ -0,0 +1,275 @@
...
+    w1 = get_parallelogram_area(p, v2, v0) / area
+    w2 = get_parallelogram_area(p, v0, v1) / area
+
+    # TODO: clean this up
+    # compute the length of the rays we want to render
+    # we can make a good guess for this by looking at the average distance between vertices on the triangle faces
+    neary_vertices_dev = nearby_vertices.to(device)
+    offset = torch.mean(torch.norm(neary_vertices_dev[..., 0, :] - neary_vertices_dev[..., 1, :], dim=-1))
+
+    origins = (
```
tancik, line 177: Are these not on GPU already?

nerfstudio/exporter/texture_utils.py
```diff
@@ -0,0 +1,275 @@
...
+    # create the .obj file
+    lines_obj = ["# Generated with nerfstudio", "mtllib material_0.mtl", "usemtl material_0"]
+    lines_obj = [line + "\n" for line in lines_obj]
+    file_obj = open(output_dir / "mesh.obj", "w", encoding="utf-8")  # pylint: disable=consider-using-with
+    file_obj.writelines(lines_obj)
+
+    # write the geometric vertices
+    # move vertices back to cpu
+    vertices = vertices.cpu().numpy()
+    for i in tqdm(range(len(vertices))):
```
tancik, line 237: Use rich, not tqdm. I don't think tqdm is a dependency

nerfstudio/exporter/texture_utils.py
```diff
@@ -0,0 +1,275 @@
...
+    vertex_normals = vertex_normals.cpu().numpy()
+    for i in tqdm(range(len(vertex_normals))):
+        normal = vertex_normals[i]
+        line = f"vn {normal[0]} {normal[1]} {normal[2]}\n"
+        file_obj.write(line)
+
+    # write the faces
+    # move faces back to cpu
+    faces = faces.cpu().numpy()
+    for i in tqdm(range(len(faces))):
```
tancik, line 261: No tqdm

nerfstudio/exporter/tsdf_utils.py
```diff
@@ -0,0 +1,336 @@
...
+    weights: TensorType["xdim", "ydim", "zdim"]
+    """TSDF weights for each voxel."""
+    colors: TensorType["xdim", "ydim", "zdim", 3]
+    """TSDF colors for each voxel."""
+    voxel_size: TensorType[3]
+    """Size of each voxel in the TSDF. [x, y, z] size."""
+    origin: TensorType[3]
+    """Origin of the TSDF [xmin, ymin, zmin]."""
+    truncation_margin: float = 5.0
+    """Margin for truncation."""
```
tancik, line 58: Can you elaborate

nerfstudio/exporter/tsdf_utils.py
```diff
@@ -0,0 +1,336 @@
...
+from nerfstudio.exporter.exporter_utils import Mesh, render_trajectory
+from nerfstudio.pipelines.base_pipeline import Pipeline
+
+CONSOLE = Console(width=120)
+
+
+@dataclass
+class TSDF:
+    """
+    Class for creating TSDFs.
```
tancik, line 42: Can you add more details

nerfstudio/exporter/tsdf_utils.py
```diff
@@ -0,0 +1,336 @@
...
+        zdim = torch.arange(volume_dims[2])
+        grid = torch.stack(torch.meshgrid([xdim, ydim, zdim], indexing="ij"), dim=0)
+        voxel_coords = origin.view(3, 1, 1, 1) + grid * voxel_size.view(3, 1, 1, 1)
+
+        # initialize the values and weights
+        values = -torch.ones(volume_dims.tolist())
+        weights = torch.zeros(volume_dims.tolist())
+        colors = torch.zeros(volume_dims.tolist() + [3])
+
+        # TODO: move to device
```
tancik, line 110: Why not initialize all these values on device?

nerfstudio/exporter/tsdf_utils.py
```diff
@@ -0,0 +1,336 @@
...
+        return TSDF(voxel_coords, values, weights, colors, voxel_size, origin)
+
+    def get_mesh(self) -> Mesh:
+        """Extracts a mesh using marching cubes."""
+
+        device = self.values.device
+
+        # run marching cubes on CPU
+        tsdf_values_np = self.values.clamp(-1, 1).cpu().numpy()
+        vertices, faces, normals, _ = measure.marching_cubes(tsdf_values_np, level=0, allow_degenerate=False)
```
tancik, line 121: Here is a CUDA version of marching cubes: https://github.com/tatsy/torchmcubes

tancik: Arg, I see the install isn't straightforward

nerfstudio/exporter/tsdf_utils.py
```diff
@@ -0,0 +1,336 @@
...
+    pipeline: Pipeline,
+    output_dir: Path,
+    downscale_factor: int = 2,
+    depth_output_name: str = "depth",
+    rgb_output_name: str = "rgb",
+    resolution: Union[int, List[int]] = field(default_factory=lambda: [256, 256, 256]),
+    batch_size: int = 10,
+    use_bounding_box: bool = True,
+    bounding_box_min: Tuple[float, float, float] = (-1.0, -1.0, -1.0),
+    bounding_box_max: Tuple[float, float, float] = (1.0, 1.0, 1.0),
```
tancik, line 270: Missing docstrings

pyproject.toml
```diff
@@ -88,6 +90,7 @@ ns-render = "scripts.render:entrypoint"
 ns-export = "scripts.exporter:entrypoint"
 ns-dev-test = "scripts.github.run_actions:entrypoint"
 ns-bridge-server = "nerfstudio.viewer.server.server:entrypoint"
+ns-mesh = "scripts.mesh:entrypoint"
```
tancik, line 93: remove

scripts/exporter.py
```diff
@@ -85,25 +92,159 @@ def main(self) -> None:
...
+
+        # create the point cloud
+        point_cloud = []
+        for i, _ in enumerate(depth_images):
+            ray_bundle = cameras.generate_rays(camera_indices=i)
+            points = ray_bundle.origins + ray_bundle.directions * depth_images[i]
+            point_cloud.append(points)
+        point_cloud = torch.stack(point_cloud)
+        point_cloud = point_cloud.reshape(-1, 3)
+        color_cloud = torch.from_numpy(np.stack(color_images)).reshape(-1, 3)
```
tancik, line 195: This should be replaced with the generate point cloud function. (plus the stuff after this, I couldn't select it all in github)

tancik (commented):
Nice work!

nerfstudio/exporter/texture_utils.py
ethanweber: Good catch, this was a bug. I updated the SceneCollider classes.

nerfstudio/exporter/texture_utils.py
ethanweber: They are. Thanks.

devernay:
Nice work!

Are you aware that the [NVDiffRast license](https://github.com/NVlabs/nvdiffrast/blob/main/LICENSE.txt) would taint the nerfstudio license, and make it usable commercially only by NVidia (at least that's what I understand from 3.3, since nerfstudio becomes a derivative work). How easy would it be to switch to [PyTorch3D](https://github.com/facebookresearch/pytorch3d) (BSD), or even [neural_renderer](https://github.com/daniilidis-group/neural_renderer) (MIT)?

ethanweber:
@devernay thanks for the comments! We agree with you. I've removed nvdiffrast from the repo and am using PyTorch instead without additional libraries. It's a bit slower but I had to do this to resolve some edge artifacts that nvdiffrast had when sampling texture coordinates not contained within triangle faces.

machenmusik:
FYI - with https://github.com/nerfstudio-project/nerfstudio/pull/809/commits/eca7027daf0cab8fc4bd14ab37b018169835d3a1 I am seeing errors trying to load prior checkpoints trained with 0.1.11 into viewer with ns-train...
```
RuntimeError: Error(s) in loading state_dict for VanillaPipeline:
        Missing key(s) in state_dict: "_model.field.mlp_pred_normals.params", "_model.field.field_head_pred_normals.net.weight", "_model.field.field_head_pred_normals.net.bias".
```

nerfstudio/exporter/tsdf_utils.py
ethanweber: Possibly should be switched to that way.

machenmusik:
> FYI - with [eca7027](https://github.com/nerfstudio-project/nerfstudio/commit/eca7027daf0cab8fc4bd14ab37b018169835d3a1) I am seeing errors trying to load prior checkpoints trained with 0.1.11 into viewer with ns-train... but ns-export commands do work.
> 
> ```
> RuntimeError: Error(s) in loading state_dict for VanillaPipeline:
>         Missing key(s) in state_dict: "_model.field.mlp_pred_normals.params", "_model.field.field_head_pred_normals.net.weight", "_model.field.field_head_pred_normals.net.bias".
> ```

Looks like turning on predict-normals by default caused this; 
older datasets can be loaded by specifying `--pipeline.model.predict-normals False`

ethanweber:
@machenmusik Good catch. We need to predict normals for poisson meshing to work well. We're still debating whether to default to predicting normals or not going forward.

docs/quickstart/export_geometry.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Install extra dependencies
+
+If you want to texture the mesh with `--unwrap-method xatlas`, then you need to install the following package shown below. We use PyTorch to rasterize the mesh and interpolate vertices and normals in the texture image.
+
+## Exporting a mesh
+
+> Poisson surface reconstruction
+
+    We default to using Poisson surface reconstruction implemented in Open3D, followed by higher resolution texturing by using a texture image and querying the NeRF. See the steps below to train your own model and export as a mesh.
```
tancik, line 13: Lets default to TSDF

docs/quickstart/export_geometry.md
```diff
@@ -0,0 +1,53 @@
...
+
+    1. Train nerfacto with network settings that predict normals.
```
tancik, line 15: Add admonition stating that this will only work with models that predict normals, ie nerfacto.

docs/quickstart/export_geometry.md
```diff
@@ -0,0 +1,53 @@
...
+> TSDF Fusion
+
+    Run the folowing command to see non-default export commands. For example, `ns-export tsdf` will run TSDF Fusion on depth and color images from NeRF, followed by texturing. This will work for models that don't have predicted normals.
+
+    ```python
+    ns-export tsdf --load-config CONFIG.yml --output-dir OUTPUT_DIR
+    ```
+
+## Exporting a point cloud
+
```
tancik, line 36: Add

docs/quickstart/export_geometry.md
```diff
@@ -0,0 +1,53 @@
+# Export geometry
+
+Here we document how to export point clouds and meshes from nerfstudio. The main command you'll be working with is `ns-export`. Our point clouds are exported as `.ply` files and the textured meshes are exported as `.obj` files.
+
+## Install extra dependencies
+
+If you want to texture the mesh with `--unwrap-method xatlas`, then you need to install the following package shown below. We use PyTorch to rasterize the mesh and interpolate vertices and normals in the texture image.
+
```
tancik, line 8: Missing

nerfstudio/exporter/exporter_utils.py
```diff
@@ -0,0 +1,256 @@
...
+    rgbs = torch.cat(rgbs, dim=0)
+
+    pcd = o3d.geometry.PointCloud()
+    pcd.points = o3d.utility.Vector3dVector(points.float().cpu().numpy())
+    pcd.colors = o3d.utility.Vector3dVector(rgbs.float().cpu().numpy())
+
+    ind = None
+    if remove_outliers:
+        CONSOLE.print("Cleaning Point Cloud")
+        pcd, ind = pcd.remove_statistical_outlier(nb_neighbors=20, std_ratio=10.0)
```
tancik, line 181: Can you make std_ratio an argument

nerfstudio/exporter/texture_utils.py
```diff
@@ -0,0 +1,462 @@
...
+from nerfstudio.exporter.exporter_utils import Mesh
+from nerfstudio.pipelines.base_pipeline import Pipeline
+
+CONSOLE = Console(width=120)
+
+
+def get_parallelogram_area(
+    p: TensorType["bs":..., 2], v0: TensorType["bs":..., 2], v1: TensorType["bs":..., 2]
+) -> TensorType["bs":...]:
+    """Given three 2D points, return the area defined by the parallelogram. I.e., 2x the triangle area."""
```
tancik, line 45: Add args and returns to docstring for all function in this file

nerfstudio/exporter/texture_utils.py
```diff
@@ -0,0 +1,462 @@
...
+    Returns:
+        texture_coordinates: Tensor of texture coordinates for every face.
+        origins: Tensor of origins for every pixel.
+        directions: Tensor of directions for every pixel.
+    """
+
+    # pylint: disable=import-outside-toplevel
+    # pylint: disable=unused-variable
+    # pylint: disable=too-many-statements
+    import xatlas
```
tancik, line 223: Why not top level import?

nerfstudio/exporter/texture_utils.py
```diff
@@ -0,0 +1,462 @@
...
+    #     pixel_area=pixel_area,
+    #     camera_indices=camera_indices,
+    #     nears=nears,
+    #     fars=fars,
+    # )
+    # CONSOLE.print("Creating texture image by rendering with NeRF...")
+    # with torch.no_grad():
+    #     outputs = pipeline.model.get_outputs_for_camera_ray_bundle(camera_ray_bundle)
+    # texture_image_np = outputs["rgb"].cpu().numpy()
+    # media.write_image(str(output_dir / "material_0.png"), texture_image_np)
```
tancik, line 395: Any reason to keep around?

scripts/exporter.py
```diff
@@ -85,25 +91,174 @@ def main(self) -> None:
...
+    """Whether to use a bounding box for the TSDF volume."""
+    bounding_box_min: Tuple[float, float, float] = (-1, -1, -1)
+    """Minimum of the bounding box, used if use_bounding_box is True."""
+    bounding_box_max: Tuple[float, float, float] = (1, 1, 1)
+    """Minimum of the bounding box, used if use_bounding_box is True."""
+    texture_method: Literal["tsdf", "nerf"] = "nerf"
+    """Method to texture the mesh with. Either 'tsdf' or 'nerf'."""
+    px_per_uv_triangle: int = 4
+    """Number of pixels per UV triangle."""
+    unwrap_method: Literal["xatlas", "custom"] = "xatlas"
```
tancik, line 117: Documentation for this is needed somewhere

scripts/exporter.py
```diff
@@ -85,25 +91,174 @@ def main(self) -> None:
...
+            self.resolution,
+            self.batch_size,
+            use_bounding_box=self.use_bounding_box,
+            bounding_box_min=self.bounding_box_min,
+            bounding_box_max=self.bounding_box_max,
+        )
+
+        # possibly
+        # texture the mesh with NeRF and export to a mesh.obj file
+        # and a material and texture file
```
tancik, line 147: Make 2 lines

machenmusik:
@ethanweber thanks.  I would note that I tried with a nerfacto trained checkpoint, so it seems that poisson will only work for nerfacto when predict-normals was true during training.  

So if the idea is to change the default for nerfacto and others to true going forward, and to default to tsdf so the issue will only occur when explicitly trying the combination that won't work, it should be fine to just have a slightly more helpful error message (e.g. you didn't train this one with predict-normals, so use something else like tsdf instead)

machenmusik:
(It appears that changing the predict-normals default to true also means that older nerfacto output checkpoints need to be loaded with `--pipeline.model.predict-normals False` -- even if one isn't going to export with poisson -- so it might be better to revert that change for backward compatibility IMO.)

ethanweber:
I agree with your comments @machenmusik and will make the changes to have better defaults and documentation. Thanks for the thoughts!

machenmusik:
Thanks @ethanweber!  Upon further reflection, it seems needlessly cumbersome to force the user to specify the correct predict-normals setting when there's only one way it will work for ns-train checkpoint loading (and it's not obvious which it should be), so I created issue #1063 in case that behavior is easy to improve.

PR #817, opened 2022-10-21
DNeRF Data Parser

JulianKnodt:
This adds the ability to actually load the D-NeRF dataset, which another PR adds. This technically should be merged after that one, but they are independent. It also mostly is the same as `BlenderDataParserConfig` so I'm not sure if they should be merged together.

### DIFF
--- a/nerfstudio/cameras/cameras.py
+++ b/nerfstudio/cameras/cameras.py
@@ -81,6 +81,7 @@ class Cameras:
         camera_type: Optional[
             Union[TensorType["num_cameras"], int, List[CameraType], CameraType]
         ] = CameraType.PERSPECTIVE,
+        times: Optional[TensorType["num_cameras"]] = None,
     ):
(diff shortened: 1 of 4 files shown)


tancik:
I think it is fine to keep it a separate file for now.

ethanweber:
Thanks for your PRs @JulianKnodt! Just wanted to let you know we are currently traveling in Tel Aviv for the conference ECCV so we may be slow to reply for another week. Thanks for making additions to the codebase!

JulianKnodt:
@ethanweber no worries, enjoy Tel Aviv and ECCV!

JulianKnodt:
Assuming you're back from ECCV, let me know what else needs to be updated for this to be merged! If you want me to combine this with the D-NeRF PR I can also do that.

tancik (approved):
LGTM, Thanks!

PR #825, opened 2022-10-22
Add temporal distortion and D-NeRF model

JulianKnodt:
Rather than recreate an entire new model, blend Dynamic-NeRF into the existing vanilla nerf model.
It also adds a temporal distortion which can be applied in the middle of the Vanilla NeRF pipeline.

Not sure whether the best way to incorporate this approach is to use this, or if it makes sense to incorporate a way to add distortion directly into each model. This should be possible to add into `TensoRF` directly without modification (and also so I don't have to wait so long for convergence of the model), and am curious what the results of that would be like. If wanted, I can make changes to add a flag for that instead of creating an entirely separate model.

### DIFF
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -36,6 +36,7 @@ from nerfstudio.data.dataparsers.dnerf_dataparser import DNeRFDataParserConfig
 from nerfstudio.data.dataparsers.friends_dataparser import FriendsDataParserConfig
 from nerfstudio.data.dataparsers.nerfstudio_dataparser import NerfstudioDataParserConfig
 from nerfstudio.engine.optimizers import AdamOptimizerConfig, RAdamOptimizerConfig
+from nerfstudio.field_components.temporal_distortions import TemporalDistortionKind
 from nerfstudio.models.base_model import VanillaModelConfig
 from nerfstudio.models.instant_ngp import InstantNGPModelConfig
 from nerfstudio.models.mipnerf import MipNerfModel
@@ -194,7 +195,11 @@ method_configs["dnerf"] = Config(
     method_name="dnerf",
     pipeline=VanillaPipelineConfig(
         datamanager=VanillaDataManagerConfig(dataparser=DNeRFDataParserConfig()),
-        model=VanillaModelConfig(_target=NeRFModel, enable_temporal_distortion=True),
+        model=VanillaModelConfig(
+            _target=NeRFModel,
+            enable_temporal_distortion=True,
+            temporal_distortion_params={"kind": TemporalDistortionKind.DNERF},
+        ),
     ),
     optimizers={
         "fields": {
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -32,6 +32,7 @@ from nerfstudio.configs.base_config import (
 from nerfstudio.data.datamanagers.base_datamanager import VanillaDataManagerConfig
 from nerfstudio.data.datamanagers.semantic_datamanager import SemanticDataManagerConfig
 from nerfstudio.data.dataparsers.blender_dataparser import BlenderDataParserConfig
+from nerfstudio.data.dataparsers.dnerf_dataparser import DNeRFDataParserConfig
 from nerfstudio.data.dataparsers.friends_dataparser import FriendsDataParserConfig
 from nerfstudio.data.dataparsers.nerfstudio_dataparser import NerfstudioDataParserConfig
 from nerfstudio.engine.optimizers import AdamOptimizerConfig, RAdamOptimizerConfig
@@ -53,6 +54,7 @@ descriptions = {
     "semantic-nerfw": "Predicts semantic segmentations and filters out transient objects.",
     "vanilla-nerf": "Original NeRF model. (slow)",
     "tensorf": "tensorf",
+    "dnerf": "Dynamic-NeRF model. (slow)",
 }
 
 method_configs["nerfacto"] = Config(
@@ -159,7 +161,11 @@ method_configs["vanilla-nerf"] = Config(
         "fields": {
             "optimizer": RAdamOptimizerConfig(lr=5e-4, eps=1e-08),
             "scheduler": None,
-        }
+        },
+        "temporal_distortion": {
+            "optimizer": RAdamOptimizerConfig(lr=5e-4, eps=1e-08),
+            "scheduler": None,
+        },
     },
 )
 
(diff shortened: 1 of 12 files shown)


JulianKnodt:
This should be ready for review! I know y'all busy but just wanted to bump it in case you missed it. I should also probably join the discord or smth so I can directly ask what the desired outcome state of this kind of thing would be.

tancik (commented):
This looks good, added a couple of comments. Can you include some results in the PR description. Assuming those look good, we can go ahead and merge!

tancik:
> This should be ready for review! I know y'all busy but just wanted to bump it in case you missed it. I should also probably join the discord or smth so I can directly ask what the desired outcome state of this kind of thing would be.

Feel free to ping me on the discord server.

nerfstudio/fields/vanilla_nerf_field.py
```diff
@@ -89,11 +93,15 @@ def __init__(
     def get_density(self, ray_samples: RaySamples):
         if self.use_integrated_encoding:
             gaussian_samples = ray_samples.frustums.get_gaussian_blob()
+            if self.temporal_distortion is not None:
+                positions = self.temporal_distortion(gaussian_samples, ray_samples.times)
```
tancik, line 97: This will break because the `guassian_samples` are means and covariances, not just positions.

JulianKnodt:
This should be ready for review again, but really I should post pictures of convergence first.

nerfstudio/configs/method_configs.py
```diff
@@ -32,6 +32,8 @@
 from nerfstudio.data.datamanagers.base_datamanager import VanillaDataManagerConfig
 from nerfstudio.data.datamanagers.semantic_datamanager import SemanticDataManagerConfig
 from nerfstudio.data.dataparsers.blender_dataparser import BlenderDataParserConfig
+
+# from nerfstudio.data.dataparsers.dnerf_dataparser import DNeRFDataParserConfig
```
tancik, line 36: Remove

nerfstudio/configs/method_configs.py
```diff
@@ -159,7 +161,11 @@
         "fields": {
             "optimizer": RAdamOptimizerConfig(lr=5e-4, eps=1e-08),
             "scheduler": None,
-        }
+        },
```
tancik, line 164: I think maybe just make a new config for dnerf. It will make it more visible.

nerfstudio/fields/tensorf_field.py
```diff
@@ -83,7 +83,9 @@ def __init__(
         self.field_output_rgb = RGBFieldHead(in_dim=self.mlp_head.get_out_dim(), activation=nn.Sigmoid())
 
     def get_density(self, ray_samples: RaySamples):
-        positions = SceneBox.get_normalized_positions(ray_samples.frustums.get_positions(), self.aabb)
+        sample_positions = ray_samples.frustums.get_positions()
```
tancik, line 86: These changes in tensorf can probably be removed.

nerfstudio/models/base_model.py
```diff
@@ -51,6 +52,11 @@ class ModelConfig(InstantiateConfig):
     eval_num_rays_per_chunk: int = 4096
     """specifies number of rays per chunk during eval"""
 
+    enable_temporal_distortion: bool = False
```
tancik, line 55: I'm thinking we might not want to add this to the base model. If you want to add distortions to a model, one also needs to modify the model file. I worry that users may see this option in the cli and believe that enabling it will cause it to just work (but this is not the case if the model doesn't do anything with it). Thoughts on just requiring devs to add these few lines to their models if they want to use temp distortion?

JulianKnodt: that is probably much safer in terms of user understanding and correctness, since it does need to be added per model. It would be nice to get it on all the models, but realistically probably not going to happen for a bit

JulianKnodt:
image
Picture of it slowly converging (albeit artifacts but I'm not sure if they're present in the original approach)

I also noticed if I zoom out very far, nothing will appear because outside of the cameras' viewpoints the temporal distortion is not well defined. I may include a component for AABB checks unless it would be outside of the scope.

nerfstudio/models/base_model.py
JulianKnodt: I've moved this specific config into the `VanillaConfig`, but it is still checked in the base `Model` so all models will have `temporal_distortion = None` unless it is set by a dev in the config (and presumably used in the model). The use of `getattr` might rely more on dynamic typing depending on taste, but I think it is fitting for this circumstance.

tancik: I think this places the flag at a higher level than it should be. For example, now if you type `--help` on any model you get the following
image
(This is for nerfacto)

It looks like you have the flag in the VanillaModelConfig, can this one be used
image

JulianKnodt: I think having it in `--pipeline.model.<temporal-distortion>` is where it was previously, but will try to move it deeper so it only appears for vanilla nerf.

JulianKnodt: Currently the parameter is shown here 
image

I realize there was an additional `temporal-distortion` which was unused which was shown to all models.

nerfstudio/models/tensorf.py
```diff
@@ -194,7 +194,7 @@ def populate_modules(self):
             self.collider = AABBBoxCollider(scene_box=self.scene_box)
 
     def get_param_groups(self) -> Dict[str, List[Parameter]]:
-        param_groups = {}
+        param_groups = super().get_param_groups()
```
tancik, line 197: This isn't necessary, right?

nerfstudio/models/base_model.py
```diff
@@ -103,13 +106,22 @@ def populate_modules(self):
...
+
     def get_param_groups(self) -> Dict[str, List[Parameter]]:
         """Obtain the parameter groups for the optimizers
 
         Returns:
             Mapping of different parameter groups
         """
+        param_groups = {}
+        if self.temporal_distortion is not None:
+            param_groups["temporal_distortion"] = list(self.temporal_distortion.parameters())
```
tancik, line 123: For the same reason as the configs, I don't think we want to include this distortion logic in the base model. If someone added temporal distortions in their model that were on by default, thus not needed a flag like `enable_temporal_distortion` this code would fail.

tancik (commented):
Getting close!

JulianKnodt:
Image from Tensorboard:
(Ground Truth, Coarse, Fine)
image
image


Copy-Paste from [D-NeRF paper](https://arxiv.org/pdf/2011.13961.pdf)
image

Eval Images Metrics/fine_psnr of current implementations
Eval Images Metrics_fine_psnr

Currently the results of this implementation on the Lego D-NeRF dataset appear to be better, not sure exactly why, but could be because of the fourier features instead of positional encoding.

tancik (approved):
LGTM. Thanks for working through all the changes! Feel free to merge (and let me know if you don't have permission to)

JulianKnodt:
@tancik I still don't have permissions to do so, I think it's probably collaborators on the repo or members of the org who have permissions

image

PR #830, opened 2022-10-24
Fix completion install script for bash

brentyi:
It seems like #747 added a few lines for initializing zsh completions, but these commands will cause errors in bash (`autoload: command not found`, `compinit: command not found`, etc).

I just:
- Removed these lines from the bash install.
- Added a condition to only run `compinit` when needed. `compinit` is quite slow (often 100s of milliseconds), so it's nice to avoid running it if it's not needed.
- Made the completion uninstall process more general.

### DIFF
--- a/scripts/completions/install.py
+++ b/scripts/completions/install.py
@@ -133,46 +133,52 @@ def _update_rc(
     """
 
     # Install or uninstall `source_line`.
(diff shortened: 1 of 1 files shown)


tancik (approved):
LGTM, thanks!

PR #831, opened 2022-10-24
Pathlib Python 3.7 Compatibility Fix

akristoffersen:
Small fix. the `missing_ok` optional arg in `Path.unlink` is only for Python 3.8+, so this makes it compatible with 3.7.


tancik (approved):
Thanks for fixing. Good to know that the process data script now works with 3.7

PR #832, opened 2022-10-25
Fix skipping non-existent images

kevinddchen:
Some COLMAP datasets have missing images, like [fox](https://github.com/NVlabs/instant-ngp/tree/master/data/nerf/fox). This PR fixes the logic so that missing images are correctly ignored.

Also included is a slight modification which should still work for all filesystems.

### DIFF
--- a/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
+++ b/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 
 import math
 from dataclasses import dataclass, field
-from pathlib import Path, PureWindowsPath
+from pathlib import Path, PurePath
 from typing import Optional, Type
 
 import numpy as np
@@ -80,12 +80,9 @@ class Nerfstudio(DataParser):
         num_skipped_image_filenames = 0
 
         for frame in meta["frames"]:
-            if "\\" in frame["file_path"]:
-                filepath = PureWindowsPath(frame["file_path"])
-            else:
-                filepath = Path(frame["file_path"])
+            filepath = PurePath(frame["file_path"])
             fname = self._get_fname(filepath)
-            if not fname:
+            if not fname.exists():
                 num_skipped_image_filenames += 1
             else:
                 image_filenames.append(fname)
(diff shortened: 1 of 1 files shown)


nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -80,12 +80,9 @@ def _generate_dataparser_outputs(self, split="train"):
         num_skipped_image_filenames = 0
 
         for frame in meta["frames"]:
-            if "\\" in frame["file_path"]:
-                filepath = PureWindowsPath(frame["file_path"])
-            else:
-                filepath = Path(frame["file_path"])
+            filepath = PurePath(frame["file_path"])
```
tancik, line 83: Thanks for the simplification

tancik (approved):
Thanks for the improvements!

kevinddchen:
@tancik this one is also ready to go. Can you merge? Thanks.

Also, I feel bad bugging you each time I need to merge. Is it possible to have write permission to this repo? Thanks :)

kevinddchen:
Thanks!

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
terrancewang: This change actually doesn't work in the case where the data is processed on a Windows machine and trained / loaded on a Linux machine. The file paths in the dataset will be Windows file paths, but will be assumed to be a Linux file path by PurePath, because that's the OS of the machine loading the data. In this case, I think you still need to specify when the path loaded in is a Windows filepath with PureWindowsPath.

kevinddchen: Hmmm interesting. Can you elaborate how this fails?

I always thought that `frame["file_path"]` is a relative path, like `images/0000.jpg`, so that the full resolved path would be `self.config.data / frame["file_path"]`.

I admit I can't really test any Windows stuff since im on Mac, and I didn't really understand the failure for Windows in the first place.

terrancewang: yeah, it was a with the process_data script because it would write out windows paths to the transforms.json file if run on Windows, which then couldn't be processed by a PurePosixPath. I think just changing process_data to always write posix paths should fix the issue as well and leave the code more clear.

kevinddchen: OK, I see your new PR. Thanks for the fix!

Issue #821, opened 2022-10-21
Training with semantic-nerfw is broken after #750 commit.


vahidEttehadi:
I cannot run "ns-train semantic-nerfw friends-data --data data/friends/Friends-monica_apartment" after v0.1.3 commit. it shows the below error:

  File "/mnt/07b9581f-566a-42d8-9478-9d2e5fd40e78/Projects/nerfstudio/nerfstudio/models/semantic_nerfw.py", line 107, in populate_modules
    for _ in range(self.config.num_proposal_network_iterations):
AttributeError: 'SemanticNerfWModelConfig' object has no attribute 'num_proposal_network_iterations'

kevinddchen:
This should be fixed with #838. Please re-open if you still have this issue.

PR #838, opened 2022-10-26
Fix semantic NeRF config bug

dolhasz:
Seems that there has been an update to the property naming in the `NerfactoModelConfig` class, which wasn't propagated into `semantic_nerfw.py`.

The culprit was `num_proposal_network_iterations` changing to `num_proposal_iterations`.


kevinddchen:
This fixes #821

tancik (approved):
LGTM, thanks for the bugfix!

PR #841, opened 2022-10-27
docs(scripts/eval.py) Update comments

ArjunSharda:
Updates the comments in scripts/eval.py as some in my opinion can be organized better and gramatically better.


tancik (approved):
Thanks for the updates

PR #842, opened 2022-10-27
Terrance/process data fix

terrancewang:


terrancewang:
don't have an insta360 dataset to test on, but I made the same fix and it should work

tancik (approved):
LGTM

PR #847, opened 2022-10-29
Add control for resampling images

kevinddchen:
~~This PR changes the default value of `--pipeline.datamanager.train-num-images-to-sample-from` to 300 to avoid OOM when caching train data for large datasets.~~ Also, the configuration `--pipeline.datamanager.train-num-times-to-repeat-images` is added which controls how often to resample images from the total dataset, if it is too large to cache.

~~The Record3D data parser's `--max_dataset_size` configuration is now bypassed, since the handling of cached data is done better in the `DataLoader`.~~

I also add some console debug logging, to warn users when not all the training images are being used.

Looking for feedback on this PR!

This finishes the remaining tasks in #833


nerfstudio/data/datamanagers.py
```diff
@@ -340,9 +354,6 @@ def setup_eval(self):
             num_workers=self.world_size * 4,
         )
 
-        # TODO: eval dataloader should be separate from train
-        self.iter_eval_dataloader = iter(self.eval_image_dataloader)
-
```
kevinddchen, line 345: Is this code dead?

nerfstudio/data/datamanagers.py
```diff
@@ -240,12 +244,18 @@ class VanillaDataManagerConfig(InstantiateConfig):
     """Specifies the dataparser used to unpack the data."""
     train_num_rays_per_batch: int = 1024
     """Number of rays per batch to use per training iteration."""
-    train_num_images_to_sample_from: int = -1
+    train_num_images_to_sample_from: int = 300
```
kevinddchen, line 247: I change the default value here to handle large datasets better to avoid OOM.

If anyone does not like this, I can revert to the default value.

nerfstudio/data/utils/dataloaders.py
```diff
@@ -61,7 +63,13 @@ def __init__(
 
         self.cached_collated_batch = None
         if self.cache_all_images:
+            CONSOLE.log(f"Caching all {len(self.dataset)} images from the dataset")
```
tancik, line 66: Use `CONSOLE.print` for these console statements.

nerfstudio/data/datamanagers.py
```diff
@@ -294,9 +304,11 @@ def __init__(
     def setup_train(self):
         """Sets up the data loaders for training"""
         assert self.train_dataset is not None
+        CONSOLE.log("Setting up training dataset...")
```
tancik, line 307: Use CONSOLE.print for all the console logging.

nerfstudio/data/datamanagers.py
tancik, line 247: I'm worried that it may not be intuitive if the default behavior doesn't used all of the training data. Perhaps instead we could default use all the images, but print a warning to the user if they are using more that 500 images that they can specify a subset.

nerfstudio/data/dataparsers/record3d_dataparser.py
```diff
@@ -50,9 +50,8 @@ class Record3DDataParserConfig(DataParserConfig):
     """Scene scale."""
     orientation_method: Literal["pca", "up"] = "up"
     """The method to use for orientation"""
-    max_dataset_size: int = 300
-    """Max number of images to train on. If the dataset has
-    more, images will be sampled approximately evenly."""
+    max_dataset_size: int = -1
+    """This option has been deprecated. Use --pipeline.datamanager.train-num-images-to-sample-from instead."""
```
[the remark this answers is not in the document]
tancik: We have plans to deprecate the record3d_dataparser by converting record3d data to the nerfstudio format using `ns-process-data`.

tancik (commented):
Thanks for the updated. Left a few comments.

nerfstudio/data/datamanagers.py
kevinddchen: OK sounds good. I will revert the changes to default values in this PR, so that the additions will be opt-in instead.

nerfstudio/data/dataparsers/record3d_dataparser.py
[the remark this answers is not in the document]
kevinddchen: OK, i have reverted this change

nerfstudio/data/utils/dataloaders.py
```diff
@@ -61,7 +63,15 @@ def __init__(
 
         self.cached_collated_batch = None
         if self.cache_all_images:
+            CONSOLE.print(f"Caching all {len(self.dataset)} images from the dataset.")
+            if len(self.dataset) > 500:
+                CONSOLE.print("Warning: If you run out of memory, try reducing the number of images to sample from.")
```
tancik, line 68: Can you add "[bold yellow]Warning: If..."

tancik (approved):
One small change. Feel free to merge after fixing. Thanks!

kevinddchen:
@tancik Thanks for taking the time to give feedback! :)

nerfstudio/data/utils/dataloaders.py
kevinddchen: Sure! Maybe should learn more about this console... 😅

tancik: It's a great library - https://github.com/Textualize/rich

PR #860, opened 2022-10-31
always list paths as posix paths

terrancewang:


tancik (approved):
LGTM. Feel free to merge after testing it.

PR #863, opened 2022-10-31
Viewer visualization crashes when using DDP

thomasw21:
`check_main_thread` would return None in non-main processes, causing the script to crash. We now make sure to call that function only on `local_rank==0`


tancik (approved):
LGTM, thanks for the fix!

PR #865, opened 2022-10-31
Check main thread when always rendering.

thomasw21:
Related to: https://github.com/nerfstudio-project/nerfstudio/pull/863

After training the `always_render` method relies on `viewer_state` to be not None, that's only True in the main process


tancik:
Feel free to merge.

PR #872, opened 2022-11-01
[WIP] Record3D Dataparser moved to ns-process-data script.

akristoffersen:


scripts/process_data.py
```diff
@@ -801,6 +880,64 @@ def main(self) -> None:
...
+    """
+
+    data: Path
+    """Path the data, either a video file or a directory of images."""
+    output_dir: Path
+    """Path to the output directory."""
+    num_downscales: int = 3
+    """Number of times to downscale the images. Downscales by 2 each time. For example a value of 3
+        will downscale the images by 2x, 4x, and 8x."""
+    verbose: bool = False
```
tancik, line 900: We should specify a target number of images. otherwise these datasets get quite large, right?

akristoffersen: I was thinking this would be better put in the nerfacto dataparser. Since then you could scale the number of training images you use without having to go back to the process step.

tancik: Hmm, that is a good point. The one thing I'm worried about his how we set defaults. Because if a user explicitly processes one of other "image" or "videos" type dataset to have lots of images, it may be counterintuitive if not all of the images are used due to a default set in the dataparser.

akristoffersen: I could see either way-- Maybe we could just have a reasonably high default count in the nerfacto dataparser, like 1k or something.

tancik: Does record3d perform well with 1k images?

akristoffersen: I can check-- I doubt it hurts performance but it might just take a long time for everything to load in and doesn't give much gains

tancik (approved):
LGTM, feel free to merge after addressing minor comment.

PR #875, opened 2022-11-01
Revert change to pure paths

### LINKED ISSUE #874: wrong path format in sf_street/transforms.json
**Describe the bug**

image paths in `sf_street/transforms.json` are badly formatted, triggering an error during data loading.
E.g. `images\\frame_00001.png` should be `images/frame_00001.png`

**To Reproduce**

`ns-download-data --dataset nerfstudio --capture sf_street` and look at `sf_street/transforms.json`

kevinddchen:
This reverts a change made in #832 to support `transforms.json` containing relative paths made in Windows. Thanks to @terrancewang for pointing this issue out.

In particular, the `sf_street` dataset contains a filename `images\\frame_00001.png`. This must be converted into `images/frame_00001.png` in POSIX systems.

This resolves #874 .

### DIFF
--- a/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
+++ b/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 
 import math
 from dataclasses import dataclass, field
-from pathlib import Path, PurePath
+from pathlib import Path, PurePath, PureWindowsPath
 from typing import Optional, Type
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


terrancewang:
Sorry actually I think we should keep your original change to PurePaths, I made a change in the data processor that should always write POSIX paths to transforms.json, regardless of the OS. I think your change makes the code more readable and clean!

tancik:
@terrancewang update was here. - #860

kevinddchen:
OK. We should update the datasets in Google Drive then. But what about the people who already downloaded the dataset?

Maybe a compromise is to add a deprecation warning telling users to re-download the dataset, and that in the future these Windows paths will not be supported.

PR #878, opened 2022-11-02
add correct none check to distortion params

nikmo33:
If distortion params are provided and not None, then this check fails with 
`RuntimeError: Boolean value of Tensor with more than one value is ambiguous`


nikmo33:
ns-dev-test

tancik (approved):
LGTM. Thanks for the PR!

tancik:
Can you run `ns-dev-test` to fix the formatting errors

nikmo33:
> Can you run `ns-dev-test` to fix the formatting errors

Thanks! Done!

PR #881, opened 2022-11-02
disallow fars <= nears

decrispell:
If a camera ray does not intersect the bounding box, `AABBBoxCollider` can return intervals with negative far and/or far < near, which in turn causes trouble downstream.  This is a simple fix to ensure far is always > near.


tancik (approved):
LGTM, thanks for the fix.

Issue #880, opened 2022-11-02
Semantic segmentation vs Semantic thing / class


nikmo33:
Currently looking at the use of semantic segmentation - there seems to not really be a good use case to split the segmentation into thing / stuff classes. And Given the assumption that a 3d point can only belong to a single semantic class this should always be true?
I was looking into integrating semantic nerf into my use case and find that the splitting into things / stuff seems necessary and adds added complexity, whereas I would like to just load the semantic labels as is and have a single head to perform semantic segmentation. It could add an additional head on top of thing/stuff, but this seems un-necessary and adds more complexity to the code.
What do people think about changing the structure of the Semantics class to the following - 

```
@dataclass
class Semantics:
    """Dataclass for semantic labels."""

    filenames: List[Path]
    """filenames to load "stuff"/background data"""
    classes: List[str]
    """class labels for semantic data"""
    colors: torch.Tensor
    """color mapping for semantic classes"""
    is_thing: List[bool] = []
    """Optional flag to indicate if the class is a thing - Can be used for usecases that require this? """
   class_loss_weights: List[float] = []
    """Loss weighting for each class, if empty then uniform weighting is assumed"""

```
This should be flexible enough to support the current use case and allows simplifying the model code quite a bit? Im happy to self-assign this to myself as well if there is some consensus on this design decision. Thanks!

nikmo33:
I have outlined a proposal in this PR - https://github.com/nerfstudio-project/nerfstudio/pull/886 . This is without the is_thing and loss_weight above, but can be added in pretty trivially if required

ethanweber:
Hey @nikmo33, I just replied to your PR! We can discuss there. 🙂

PR #886, opened 2022-11-03
[WIP] nikhil/simplify-semantics

nikmo33:
This is a proposal meant to show a potential solution for https://github.com/nerfstudio-project/nerfstudio/issues/880
Im not quite sure how to integrate the current friends semantics into the current format, so just to see that the whole pipeline works as expected im only using the semantic thing labels. Mainly confused because the following assertion fails for me- 
assert (thing_semantics[stuff_semantics == 0] == 0).sum() == 0
assert (stuff_semantics[thing_semantics == 0] == 0) .sum() == 0
Any guidance on this ? How can this be the case - I thought every pixel can only belong to a single class.


nerfstudio/data/dataparsers/base_dataparser.py
```diff
@@ -62,6 +56,7 @@ class DataparserOutputs:
     scene_box: SceneBox = SceneBox()
     """Scene box of dataset. Used to bound the scene or provide the scene scale depending on model."""
     semantics: Optional[Semantics] = None
+
```
ethanweber, line 59: You can remove this newline.

ethanweber:
Thanks for the PR. I think it's fine to switch to a single semantics class as you suggest. The current code works for panoptic segmentation from [detectron2](https://github.com/facebookresearch/detectron2/tree/main/configs/COCO-PanopticSegmentation).

When thing semantics == 0, it is the stuff class. When stuff semantics == 0, it is the thing class. You can see this by looking at the `panoptic_classes.json` file.

```json
{
    "stuff": [
        "thing",
        "banner",
        "blanket",
        ...
    ],
    ....
    "thing": [
        "stuff",
        "person",
        "bicycle",
        ...
    ],
    ....
}
```

ethanweber (commented):
Looks good to me. Thanks for simplifying it! If you wanted to continue improving this part of the codebase, you could modify [this logic](https://github.com/nerfstudio-project/nerfstudio/blob/0afcd1774b7cddef5abdab2babe6582fc21e639e/nerfstudio/data/dataparsers/friends_dataparser.py#L194) and [this logic](https://github.com/nerfstudio-project/nerfstudio/blob/0afcd1774b7cddef5abdab2babe6582fc21e639e/nerfstudio/data/utils/datasets.py#L83-L87). Let me know if you're interested and we can discuss further!

nerfstudio/data/dataparsers/base_dataparser.py
nikmo33: > Thanks for the PR. I think it's fine to switch to a single semantics class as you suggest. The current code works for panoptic segmentation from [detectron2](https://github.com/facebookresearch/detectron2/tree/main/configs/COCO-PanopticSegmentation).
> 
> When thing semantics == 0, it is the stuff class. When stuff semantics == 0, it is the thing class. You can see this by looking at the `panoptic_classes.json` file.
> 
> ```json
> {
>     "stuff": [
>         "thing",
>         "banner",
>         "blanket",
>         ...
>     ],
>     ....
>     "thing": [
>         "stuff",
>         "person",
>         "bicycle",
>         ...
>     ],
>     ....
> }
> ```

Sorry Ethan I updated my original description. 
Basically it is not always the case that when stuff == 0, thing != 0.

nikmo33:
> Looks good to me. Thanks for simplifying it! If you wanted to continue improving this part of the codebase, you could modify [this logic](https://github.com/nerfstudio-project/nerfstudio/blob/0afcd1774b7cddef5abdab2babe6582fc21e639e/nerfstudio/data/dataparsers/friends_dataparser.py#L194) and [this logic](https://github.com/nerfstudio-project/nerfstudio/blob/0afcd1774b7cddef5abdab2babe6582fc21e639e/nerfstudio/data/utils/datasets.py#L83-L87). Let me know if you're interested and we can discuss further!

Yeah that is an area that would be nice to improve. It would be good to discuss a solution for this before I go about changing this logic?

nerfstudio/data/dataparsers/base_dataparser.py
nikmo33: And I am confused about how to handle this case for the current friends dataset with the proposed changes

ethanweber: For the friends dataset, I think we can just remove the stuff classes for now. Can you incorporate that change into this PR to avoid that code breaking?

ethanweber:
> > Looks good to me. Thanks for simplifying it! If you wanted to continue improving this part of the codebase, you could modify [this logic](https://github.com/nerfstudio-project/nerfstudio/blob/0afcd1774b7cddef5abdab2babe6582fc21e639e/nerfstudio/data/dataparsers/friends_dataparser.py#L194) and [this logic](https://github.com/nerfstudio-project/nerfstudio/blob/0afcd1774b7cddef5abdab2babe6582fc21e639e/nerfstudio/data/utils/datasets.py#L83-L87). Let me know if you're interested and we can discuss further!
> 
> Yeah that is an area that would be nice to improve. It would be good to discuss a solution for this before I go about changing this logic?

Agreed. Are you on discord? We could chat there if you are interested in this.

nerfstudio/data/dataparsers/base_dataparser.py
nikmo33: Thats basically what Ive already done so it should be good to go in that case!

ethanweber (approved):
LGTM!

PR #910, opened 2022-11-06
nikhil/nan-training-debug

nikmo33:
Got to this function be tracing nans back from model outputs up the stack. Found that before the call to 
`coord_stack = camera_utils.radial_and_tangential_undistort(coord_stack, distortion_params)` the coord_stack param is not nan and is nan after. When debugging within this function, I had to switch of torch scripting to insert a breakpoint. This caused the issue to go away, and none of the downstream breakpoints were hit either. When going to reference function in comment they do not seem to script this function. Not sure what within scripting is causing this to break

To reproduce the old error uncomment the torch.jit.script - and you should hit the breakpoint

### DIFF
--- a/nerfstudio/cameras/camera_utils.py
+++ b/nerfstudio/cameras/camera_utils.py
@@ -338,7 +338,7 @@ def _compute_residual_and_jacobian(
     return fx, fy, fx_x, fx_y, fy_x, fy_y
 
 
-@torch.jit.script
+# @torch.jit.script
 def radial_and_tangential_undistort(
     coords: torch.Tensor,
     distortion_params: torch.Tensor,
--- a/nerfstudio/cameras/cameras.py
+++ b/nerfstudio/cameras/cameras.py
@@ -298,7 +298,9 @@ class Cameras:
 
         if distortion_params is not None:
             coord_stack = camera_utils.radial_and_tangential_undistort(coord_stack, distortion_params)
-
+            if torch.isnan(coord_stack).sum():
+                nan_indices = torch.nonzero(torch.isnan(coord_stack), as_tuple=False)
+                import ipdb; ipdb.set_trace()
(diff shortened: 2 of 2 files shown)


tancik:
Great find! It does seem to fix the NAN issue on the redwood2 scene. The only downside is that there is a non-negligible performance hit (I'm seeing that training is ~10% slower). Lets investigate this a bit more before merging, I would be nice if we didn't need to remove jitting. Maybe @kerrj has an idea, he originally added the jitting.

kerrj:
A quick idea would be changing `eps` in that function to something larger (maybe 1e-4 or something), it's possible lines 373 and 374 are exploding when the denominator is too small. If that doesn't fix it, pinpointing which line exactly is causing nans would be helpful, but that might be tricky inside a jitted function.

kerrj:
Tested to 30k iterations with eps=1e-3 and it seems to work

nikmo33:
> Tested to 30k iterations with eps=1e-3 and it seems to work

Yeah this seems like a better solution - it works for me as well.

tancik (approved):
Seems reasonable to me.

kerrj:
Would be good to do a quality comparison to make sure it doesn't affect results

nikmo33:
> Would be good to do a quality comparison to make sure it doesn't affect results. 
Screenshot 2022-11-07 at 11 41 23


Here is what I get when testing on the poster dataset - 
orange is eps 1e-3 and blue 1e-9 - seems fine to me
Screenshot 2022-11-07 at 11 41 28

tancik:
Tested on a fisheye dataset. Seems to work fine.

gs14iitbbs:
> Yeah this seems like a better solution - it works for me as well, Ive reverted the jit and updated the eps value

I tried on another dataset (gardenspheres) - the eps change delayed the happening of  'nan', but still occurs in later iterations.

PR #932, opened 2022-11-10
Do not apply activation for last MLP layer

kevinddchen:
This fixes `activation` being applied on the last layer:

```
... -> linear -> `activation` -> `output_activation` -> (output tensor)
```

Also replacing tuple-lookup with set-lookup.


tancik (approved):
Thanks for the fix!

PR #933, opened 2022-11-10
Add missing activation

tancik:
When using the MLP defaults it does not set the output activation. The current models assume a relu activation has been applied.


kevinddchen (approved):
Oh damn! What a cascade of changes!

PR #947, opened 2022-11-11
Streamline Viewer in Colab, added more checks before training

akristoffersen:


tancik (approved):
Thanks for making these changes. Hopefully it will be less confusing to users.

PR #957, opened 2022-11-14
add support for per-image mask filenames in transforms.json

decrispell:
Add support for per-image mask filenames via a `"mask_path"` key in each element of `"frames"` in `transforms.json`.   The `"mask_path"` key must be present for all frames or zero frames.


nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -171,6 +182,7 @@ def _generate_dataparser_outputs(self, split="train"):
             image_filenames=image_filenames,
             cameras=cameras,
             scene_box=scene_box,
+            mask_filenames=mask_filenames,
```
ethanweber, line 185: `mask_filenames if len(mask_filenames) > 0 else None`

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -87,6 +88,10 @@ def _generate_dataparser_outputs(self, split="train"):
             else:
                 image_filenames.append(fname)
                 poses.append(np.array(frame["transform_matrix"]))
+            if "mask_path" in frame:
+                mask_filepath = PurePath(frame["mask_path"])
+                mask_fname = self._get_fname(mask_filepath)
```
ethanweber, line 93: So I don't think `self._get_fname` will work if downscale_factor != 1. This is specific for the downscaled RGB images. Were you able to test this?

ethanweber (commented):
Awesome thanks for the PR! I left a few comments. Also feel free to add documentation [here](https://github.com/nerfstudio-project/nerfstudio/blob/b5c34f0154a3358a8483120b63988a7176659a36/docs/quickstart/custom_dataset.md) for this feature so others can benefit more clearly from your work.

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
ethanweber: Right, agreed with Matt. Instead of an "images" folder, it can be called "masks".

decrispell: Added the option to specify the folder prefix to `_get_fname` - The masks will be searched for in `masks_X`, where `X` is the downscale level.

nikmo33: Another option here (I do this in my local fork for custom data) is to support image / mask / semantic downscaling in the dataset. It is common to only store images at a single resolution so probably makes sense to support that use case?

decrispell: I like that idea a lot @nikmo33 - In my opinion it's beyond the scope of this PR, though.

plusk01:
is this PR ready to be merged?

decrispell:
> is this PR ready to be merged?

This is more of a question for the project maintainers, but it is ready to be merged as far as I'm concerned.

tancik (approved):
LGTM
There is now support to do downscaling during dataloading. This would let us downscale if the downscaled versions don't exist.

PR #966, opened 2022-11-16
Add Support for Equirectangular Images

kevinddchen:
This PR adds support for equirectangular images. The ray directions should be correctly created based on the image type, but the pixel sampler needs to be manually set by a CLI config. Since the cameras in a dataset can technically have multiple types (e.g. equirectangular and perspective), I could not think of a nice way to deal with everything at the same time. In the end, the `EquirectangularPixelSampler` isn't really needed, since the inverse CDF `acos(1 - 2x)` for mapping to a sphere is already so similar to that of the uniform distribution.

Special thanks for @TheQuantumFractal for initial steps with this PR :)

Things not done yet:
- [x] Equirect Pixel sampler

Things for a later PR:
- Handle masks with equirectangular images.


scripts/process_data.py
```diff
@@ -227,7 +231,8 @@ def convert_insta360_to_images(
         else:
             CONSOLE.print("[bold red]Can't satify requested number of frames. Extracting all frames.")
 
-        vf_cmds.append(f"crop=iw*{crop_percentage}:ih*{crop_percentage}")
+        # vf_cmds.append(f"crop=iw*({crop_percentage}):ih*({crop_percentage})")
+        vf_cmds.append("v360=dfisheye:equirect:ih_fov=190:iv_fov=190:yaw=-90")
```
kevinddchen, line 235: @kerrj This will change the current behavior of insta360

scripts/process_data.py
```diff
@@ -726,8 +884,12 @@ class ProcessImages:
     """Path the data, either a video file or a directory of images."""
     output_dir: Path
     """Path to the output directory."""
-    camera_type: Literal["perspective", "fisheye"] = "perspective"
+    camera_type: Literal["perspective", "fisheye", "equirectangular"] = "perspective"
     """Camera model to use."""
+    sfm_method: Literal["colmap", "OpenSfM"] = "OpenSfM"
```
kevinddchen, line 889: @kerrj Should we use `colmap` by default?

scripts/process_data.py
```diff
@@ -812,10 +994,14 @@ class ProcessVideo:
     """Path the data, either a video file or a directory of images."""
     output_dir: Path
     """Path to the output directory."""
-    num_frames_target: int = 300
+    num_frames_target: int = 150
     """Target number of frames to use for the dataset, results may not be exact."""
-    camera_type: Literal["perspective", "fisheye"] = "perspective"
+    camera_type: Literal["perspective", "fisheye", "equirectangular"] = "perspective"
     """Camera model to use."""
+    sfm_method: Literal["colmap", "OpenSfM"] = "OpenSfM"
```
kevinddchen, line 1001: @kerrj Similar to above, should we revert to `300` and use `colmap`?

scripts/process_data.py
```diff
@@ -944,28 +1149,28 @@ def main(self) -> None:
...
             with CONSOLE.status("[bold yellow]Saving results to transforms.json", spinner="balloon"):
-                num_matched_frames = colmap_to_json(
-                    cameras_path=colmap_dir / "sparse" / "0" / "cameras.bin",
-                    images_path=colmap_dir / "sparse" / "0" / "images.bin",
+                num_matched_frames = opensfm_to_json(
+                    cameras_path=opensfm_dir / "sparse" / "0" / "cameras.bin",
+                    images_path=opensfm_dir / "sparse" / "0" / "images.bin",
                     output_dir=self.output_dir,
-                    camera_model=CAMERA_MODELS["fisheye"],
+                    camera_model=CAMERA_MODELS["equirectangular"],
```
kevinddchen, line 1173: @kerrj Should we revert to the original for now? Or should we be using equirect images as the de-facto standard for insta360 from here on?

kevinddchen (commented):
Some comments I thought needed addressing before merging

scripts/process_data.py
kevinddchen: Also, the `install_checks.check_colmap_installed()` checks are redundant if we are using OpenSfM

kevinddchen:
@kerrj Can I have some help cleaning up the process data script to pass linting? Since you're the original author.

scripts/process_data.py
tancik: Yes, colmap should be default

scripts/process_data.py
tancik: Yes, keep the same defaults

scripts/process_data.py
tancik: Let's keep the standard methods for now. We can change the standard in a different PR.

nerfstudio/fields/nerfacto_field.py
```diff
@@ -112,7 +112,7 @@ def __init__(
         self.use_semantics = use_semantics
 
         num_levels = 16
-        max_res = 1024
+        max_res = 2048
```
tancik, line 115: Don't include

tancik (commented):
Thanks for picking this up. Thoughts of splitting this into two PRs, one for OpenSFM and one for Equirectangular? Also want to loop in @akristoffersen @TheQuantumFractal as they have been looking into equirectangular images.

kevinddchen:
OK, I have reverted all changes to `scripts/process_data.py` and kept just the equirect image contributions. I will actually just finish off the remaining task for this PR before opening it back up for review.

nerfstudio/data/pixel_samplers.py
```diff
@@ -92,3 +92,66 @@ def sample(self, image_batch: Dict):
...
+    num_images, image_height, image_width, _ = batch["image"].shape
+
+    # only sample within the mask, if the mask is in the batch
+    if "mask" in batch:
+        # TODO(kevinddchen): implement this
+        raise NotImplementedError("Masking not implemented for equirectangular images.")
+    else:
+        num_images_rand = torch.rand(num_rays_per_batch, device=device)
+        # sample phis according to pdf(x) = sin(x) / pi, but normalized to the range [0, 1]
+        phi_rand = torch.acos(1 - 2 * torch.rand(num_rays_per_batch, device=device)) / torch.pi
```
akristoffersen, line 120: would it maybe be more intuitive to write this as calculate a random direction uniform over a unit sphere, then convert this to a pixel in the image? Can't seem to find where this math was found from

kevinddchen: I agree its doing something less intuitive.

Just making sure I'm thinking about this right: sampling uniformly over the sphere means picking `phi` and `theta`. `theta` is uniformly distributed over `[0, 2 pi]`, no confusion there. `phi` is distributed over `[0, pi]` with weight `sin(phi)`, since that is the differential area element. So the PDF is proportional to `sin(phi)`.

Here we are doing something strange, since we are using inverse transform sampling to draw random samples using samples drawn from the uniform distribution over [0, 1].

Would you like this documented better? Or is there a better way and I'm overcomplicating?

akristoffersen: Sampling over the unit sphere uniformly can be done by normallizing a vector made by 3 samples from a standard normal distribution:

> xyz = torch.normal(mean=torch.zeros((3,)))
> sample_xyz = xyz / xyz.norm()

It might be good to add documentation on similar to how we are doing here with the other camera types: https://docs.nerf.studio/en/latest/nerfology/model_components/visualize_cameras.html#

I haven't touched this part of the docs yet, but from what I understand its all notebook based so modifying one of the other graphics shouldn't be too different-- but let me know if I can help with that!

kevinddchen: Hmmm yeah, thats much more intuitive. I'll use that instead then.

I'll take a look at the docs and see if I understand what is going on there, haha.

kevinddchen: Hmmmm ok, going back to this, I did some thinking and while the 3d gaussian approach is more intuitive, its a lot harder to convert a point on the unit sphere back to coordinates in the equirectangular image.

To figure out the azimuthal angle, we will need some kind of `atan2(y, x)` but also condition on the signs of `x` and `y` and adding/subtracting multiples of `pi` to get things into the correct range.

I'm confident the current implementation is correct, so if you don't mind, can we go with this then? Its a lot simpler implementation-wise, even if the approach is less intuitive.

akristoffersen: @kevinddchen yup! That's totally good with me, just wanted to check the math was right and I hadn't seen the inverse transform sampling derivation before. I can also do a full review now and let you know if I see anything, but this all looks great so far. thanks again for the help 😀

kevinddchen: Yes please! Be super nit picky!

TheQuantumFractal: You can reference this for a derivation for sampling using the inverse CDF (http://corysimon.github.io/articles/uniformdistn-on-sphere/)

nerfstudio/cameras/cameras.py
```diff
@@ -248,7 +250,7 @@ def get_image_coords(self, pixel_offset: float = 0.5) -> TensorType["height", "w
         image_coords = torch.stack(image_coords, dim=-1) + pixel_offset  # stored as (y, x) coordinates
         return image_coords
 
-    def generate_rays(
+    def generate_rays(  # pylint: disable=too-many-statements
```
tancik, line 253: Can you add a test for this camera model (in `test/cameras/test_cameras.py`)

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -266,6 +266,9 @@ class VanillaDataManagerConfig(InstantiateConfig):
     camera_optimizer: CameraOptimizerConfig = CameraOptimizerConfig()
     """Specifies the camera pose optimizer used during training. Helpful if poses are noisy, such as for data from
     Record3D."""
+    use_equirect_sampler: bool = False
```
tancik, line 269: I need to think a bit more, but is there any way to infer whether to use this sampler given the camera_type?

TheQuantumFractal: self.train_dataset.dataparser_outputs.cameras.camera_type[0] == CameraType.EQUIRECTANGULAR.value

kevinddchen: How should we handle mixed camera types? e.g. some images equirect, some perspective. I don't think its possible at the moment, but the `Cameras` data structure allows for the possibility

kevinddchen: I see this disclaimer: https://github.com/nerfstudio-project/nerfstudio/blob/2c64ba5b6aa7c47f2e214572ad4d8c8759888b20/nerfstudio/cameras/cameras.py#L54-L56

If its OK for now to assume that all cameras in a `Cameras` take the same type, then I can do a bunch of simplifications in the pixel sampler

tancik: You could fall back to using the normal sampler if any camera isn't equirectangular, (and maybe print a warning).

kevinddchen: I am fine with this

nerfstudio/data/pixel_samplers.py
```diff
@@ -111,7 +111,7 @@ def collate_image_dataset_batch_equirectangular(batch: Dict, num_rays_per_batch:
     num_images, image_height, image_width, _ = batch["image"].shape
 
     # only sample within the mask, if the mask is in the batch
-    if "mask" in batch:
+    if "mask" in batch:  # pylint: disable=no-else-raise
```
tancik, line 114: Remove else to fix pylint error
https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-raise.html

tancik (approved):
One minor comment, but otherwise looks good to me. Thanks for adding equirectangular support!

nerfstudio/cameras/cameras.py
```diff
@@ -588,8 +590,15 @@ def _generate_rays_from_coords(
...
+        # Do not apply distortion for equirectangular images
         if distortion_params is not None:
-            coord_stack = camera_utils.radial_and_tangential_undistort(coord_stack, distortion_params)
+            mask = (self.camera_type[true_indices] != CameraType.EQUIRECTANGULAR.value).squeeze(-1)  # (num_rays)
+            coord_mask = torch.stack([mask, mask, mask], dim=0)
+            if mask.any():
+                coord_stack[coord_mask, :] = camera_utils.radial_and_tangential_undistort(
+                    coord_stack[coord_mask, :].reshape(3, -1, 2),
+                    distortion_params[mask, :],
+                ).reshape(-1, 2)
```
kevinddchen, line 601: @ethanweber I saw your refactor of `Cameras`. Before I merge this PR, can you take a quick look at this piece to make sure I'm not royally messing anything up?

ethanweber: I think it looks correct to me. The logic checks out as far as I can tell.

PR #968, opened 2022-11-16
Updating the script for downloading data

jake-austin:


scripts/downloads/download_data.py
```diff
@@ -140,43 +227,49 @@ def main(
...
+
+
+method_configs = {
+    "blender": BlenderDownload(),
+    "friends": FriendsDownload(),
+    "nerfstudio": NerfstudioDownload(),
+    "record3d": Record3dDownload(),
+    "dnerf": DNerfDownload(),
+    "phototourism": PhototourismDownload(),
+}
```
ethanweber, line 249: Do something like this instead. It's quite a bit cleaner. https://github.com/nerfstudio-project/nerfstudio/blob/009e21be248418028e6303ecde593a20ef88fa87/scripts/process_data.py#L1174-L1180

tancik (commented):
These changes will break everywhere the download data script is currently being used. These also need to be updated. From memory this include documentation and the colab.

scripts/downloads/download_data.py
```diff
@@ -94,43 +119,119 @@ def download_capture_name(save_dir: Path, dataset_name: str, capture_name: str,
...
 
 
-def download_nerfstudio(save_dir: Path, capture_name: str):
-    """Download specific captures for the nerfstudio dataset."""
-    download_capture_name(save_dir, "nerfstudio", capture_name, capture_name_to_file_id=nerfstudio_file_ids)
+@dataclass
+class NerfstudioDownload(DatasetDownload):
+    """Download the nerfstudio dataset."""
+
+    capture_name: NerfstudioCaptureName = "bww_entrance"  # type: ignore
```
brentyi, line 126: If you add a `from __future__ import annotations` to the top of this file you should be able to get rid of these `# type: ignore`s (at least for pyright/VS Code)

ethanweber (commented):
Looking better, but there are a lot of docs that need to updated still. For example, commands like `ns-download-data --dataset=nerfstudio --capture=poster`. Also looks like you renamed `--capture` to `--capture-name`, which will have to be updated as well.

ethanweber (approved):
LGTM, but that'd be great to update the PR description first.

PR #973, opened 2022-11-17
Support downsample in dataset

nikmo33:
As discussed briefly in https://github.com/nerfstudio-project/nerfstudio/pull/957 - it seems like have downsample support in the base InputDataset could be beneficial. While colmap outputs images at multiple-resolutions this may not be the case for many other applications where there are images only stored at a single resolution. This allows us to load in data in the raw format and downsample once loaded to disk

### DIFF
--- a/nerfstudio/data/datamanagers/base_datamanager.py
+++ b/nerfstudio/data/datamanagers/base_datamanager.py
@@ -266,6 +266,10 @@ class VanillaDataManagerConfig(InstantiateConfig):
     camera_optimizer: CameraOptimizerConfig = CameraOptimizerConfig()
     """Specifies the camera pose optimizer used during training. Helpful if poses are noisy, such as for data from
     Record3D."""
+    dataset_scale_factor: float = 1.0
+    """The scale factor for scaling spatial data such as images, mask, semantics
+    along with relevant information about camera intrinsics
+    """
 
 
 class VanillaDataManager(DataManager):  # pylint: disable=abstract-method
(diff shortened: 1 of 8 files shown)


nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -266,6 +266,10 @@ class VanillaDataManagerConfig(InstantiateConfig):
     camera_optimizer: CameraOptimizerConfig = CameraOptimizerConfig()
     """Specifies the camera pose optimizer used during training. Helpful if poses are noisy, such as for data from
     Record3D."""
+    dataset_camera_scale_factor: float = 1.0
```
tancik, line 269: Maybe `camera_res_scale_factor`. I worry this may be ambiguous with extrinsic scale.

tancik (commented):
This looks fine to me. I think we should still suggest that people pre-scale their images. We actually used to have this feature and found that people would upload 10mp images and everything would grind to a halt :)

nikmo33:
> This looks fine to me. I think we should still suggest that people pre-scale their images. We actually used to have this feature and found that people would upload 10mp images and everything would grind to a halt :)

Fair enough. Agree - it is probably a judgement call on which of the above methods are appropriate. Let me know if you agree that this feature in additional also makes sense

nerfstudio/data/datamanagers/base_datamanager.py
nikmo33: done

tancik (approved):
LGTM, I think having the option is generally a good thing.

PR #986, opened 2022-11-20
[WIP] Add NR-NeRF method

JulianKnodt:
[NR-NeRF](https://arxiv.org/pdf/2012.12247.pdf) is another approach to dynamic NeRF that uses rigidity and additional losses.

I don't know if dedicating a whole set of configs to dynamic approaches is good, maybe it would be good to create a subcommand and have dynamic configs under that?

I can't test it rn because I'm on a machine that has 3 GB and it seems that some tensor is being stored that leads to an eventual OOM even if I only use 50 rays per batch

Edit: turns out eval was causing OOM not mem leaking from somewhere else

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -174,6 +174,7 @@ data
 # Misc
 old/
 temp*
+*.swp
 !temporal*
 .nfs*
 external/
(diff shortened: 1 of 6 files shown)


tancik:
This has gotten stale. I'm going to close it for now.

PR #991, opened 2022-11-21
fix typo error which should be self.cameras.times

ksnzh:


ksnzh:
The old code will cause attribute error that 'Cameras' object has no attribute 'time'.

tancik (approved):
Thanks for the fix!

PR #992, opened 2022-11-21
Fix variable name error in dataloaders.py

HrsPythonix:
AttributeError: 'Cameras' object has no attribute 'time' will occur when ns-render with --traj spiral


tancik:
This was fixed in #991

PR #993, opened 2022-11-21
create rays on device when rendering

zmurez:
When using `traj=="filename"` `camera_ray_bundle` is generated on the cpu and then moved to the gpu which can be slow.


tancik (approved):
Nice! I get ~6% speedup

zmurez:
I think it depends on utilization/hardware... I got no speed up when running render on a server by itself. But when running 8 copies in parallel on the server (1 per gpu), the cpu became the bottleneck and I got 10x speedup after.

tancik:
Feel free to merge

zmurez:
> Feel free to merge

I don't have merge access.

tancik:
> I don't have merge access.

I merged for now. Sent you a new invite.

PR #994, opened 2022-11-21
implemented threadpool mechanics when loading images

dozeri83:
Hi,
I noticed that the loading of the images is not implemented with threads.

Using threadpool with 4 threads improved the loading time for me by a factor of 2-3, when the images are located locally on the machine SSD. 

If the images are located on a network drive the the loading time is better by a factor of 3-4.

This is particularly important when there are many images in the dataset and the RAM memory is "limited" ( for me 400 images could not fitt in my 48 GB memory) and the user have to use num_images_to_sample_from and num_times_to_repeat_image fields. In this case the training becomes very IO intensive ( you keep loading images throughout the training).

Thanks


tancik (approved):
LGTM, I also see a 3-4x speedup on the initial data loading

tancik:
Feel free to merge when ready.

dozeri83 (commented):
implemented threadpool mechanics when loading images.
when using threadpool with 4 threads (configurable with max_workers) 
, this improve loading time by a factor of 3-4.
I added protection code against loading too many and too less threads.

PR #998, opened 2022-11-21
TensoRF bugfix: AABB normalized_positions to [-1, 1]

JulianKnodt:
Previously, SceneBox.get_normalized_positions() returned a value in 0,1. This was fine for InstantNGP and Nerfacto, which use these values directly in an MLP. For TensoRF though, this was an error because grid sample expects values in the range `[-1,1]`. Thus, I think only 1/4th of the encoding planes were actually being used, and half of the encoding lines were.

### DIFF
--- a/nerfstudio/field_components/encodings.py
+++ b/nerfstudio/field_components/encodings.py
@@ -423,6 +423,13 @@ class TensorVMEncoding(Encoding):
         return self.num_components * 3
 
     def forward(self, in_tensor: TensorType["bs":..., "input_dim"]) -> TensorType["bs":..., "output_dim"]:
+        """Compute encoding for each position in in_positions
+
+        Args:
+            in_tensor: position inside bounds in range [-1,1],
+
+        Returns: Encoded position
+        """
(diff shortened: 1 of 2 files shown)


tancik:
Nice find. Does this change effect nerfacto / instant-ngp performance?

JulianKnodt:
I haven't checked if it affects their performance, since I was looking at TensoRF when I found this.

Does the `HashGrid` encoding expect values in `[0,1]`? If so I can just isolate it to TensoRF to the correct range.
image

tancik:
I think hashgrid expects [0,1].

JulianKnodt:
I've changed it to the TensoRF callsites, altho I browsed tcnn's readme, I didn't see a specific doc that referenced it anywhere but better to not cause random changes elsewhere.

tancik:
I just tested nerfacto with [-1,1] and it does work, but the results were worse. Lets just focus this PR on tensorf for now.

ethanweber (approved):
LGTM too. Nice find!

PR #1007, opened 2022-11-23
add render time in viewer

JulianKnodt:
Adds a slider for render time in the viewer.

Only displays when the model has temporal distortion.

Is "render time" self-explanatory or is other wording better?

### DIFF
--- a/nerfstudio/viewer/app/src/modules/ConfigPanel/ConfigPanel.jsx
+++ b/nerfstudio/viewer/app/src/modules/ConfigPanel/ConfigPanel.jsx
@@ -1,5 +1,5 @@
 import { buttonGroup, useControls } from 'leva';
-import { useContext, useEffect } from 'react';
+import { useContext, useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 
 import { WebSocketContext } from '../WebSocket/WebSocket';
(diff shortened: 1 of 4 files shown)


JulianKnodt:
image

New slider, is getting piped properly into the server, now need to figure out when to display/pipe data from the server. Not sure where to read the graph information.

nerfstudio/viewer/app/src/modules/ConfigPanel/ConfigPanel.jsx
```diff
@@ -164,6 +197,17 @@ export function RenderControls() {
         '1024px': () => setControls({ max_resolution: 1024 }),
         '2048px': () => setControls({ max_resolution: 2048 }),
       }),
+      // Dynamic NeRF rendering time
+      ...(display_render_time ? {render_time: {
+        label: 'Render Time',
```
tancik, line 202: Maybe `Render Timestep`

tancik (commented):
When changing or adding messages between the server and client we also need to update the version number - https://github.com/nerfstudio-project/nerfstudio/blob/3d70b4e61457ad077ec0ffbb5415528456517ddb/nerfstudio/viewer/app/package.json#L4
This prevents issues with backwards compatability.

tancik (approved):
LGTM, thanks for adding

Issue #979, opened 2022-11-18
How to view trained model?


HrsPythonix:
I have trained model, how can I launch viewer but not to train?

tancik:
Use the same training call, but add the following arguments, `--trainer.load_dir outputs/data-nerfstudio-{data}/{model}/{timestamp}/nerfstudio_models --viewer.start-train False`

segments-tobias:
Is it `load_dir` (as in the docs) or `load-dir` as in the CLI help?

brentyi:
> Is it `load_dir` (as in the docs) or `load-dir` as in the CLI help?

These will produce the same behavior, but the latter is preferred.

PR #1010, opened 2022-11-23
(minor) `--trainer.load_dir` => `--trainer.load-dir` in docs

brentyi:
As pointed out in https://github.com/nerfstudio-project/nerfstudio/issues/979#issuecomment-1324723023


tancik (approved):
Lgtm

PR #1020, opened 2022-11-24
config nerfacto

zmurez:
Allows Nerfacto's hashmap to be configured.

I found that increasing these values can improve the visual quality (reduces blur). As such it would be nice to be able to tune these via the command line.


HrsPythonix:
Can you share how you tune these parameters?

tancik (approved):
Lgtm

PR #1024, opened 2022-11-25
Add support for insta360 single-file format

### LINKED ISSUE #1019: ns-process-data insta360 does not support the single-file format used by the insta360 X3
**Describe the bug**
ns-process-data insta360 throws an error looking for a second file when given single-file 360 footage (both sensor's data in a single file). does not support 360 video where footage from both sensors is in the same file.

**To Reproduce**
1. Record 360 footage with a camera that records to a single file instead of two files. 
    _Note that: Insta360 X3 records to a single file (only 00) in normal video mode and only uses two files (00 and 10) for certain modes, like HDR and TimeShift._
2. Attempt to process data: `ns-process-data insta360 --data .\VID_20220925_113615_00_190.insv --output outputs`
3. See error `FileNotFoundError: Could not find VID_20220925_113615_10_190.insv`

**Expected behavior**
ns-process-data should support single-file 360 footage and an error should not be thrown by `ns-process-data insta360 --data .\VID_20220925_113615_00_190.insv --output outputs` when a second file does not exist (if the single file contains data from two sensors.

**Screenshots**
Here is an example of what the expected two-file format looks like:
Note how there are two separate 1:1 aspect ratio files with "00" and "10" in the files names. This is what ns-process-data is looking for.
image

Whereas here is what a single-file (seemingly the new standard for insta360 cameras) 360 video looks like:
Note how it's a single file, which is 2:1 aspect ratio, and this one file has "00" in the name. This should be supported.
image

**Additional context**
This single file format appears to be the new standard way that insta360 records videos, at least in the new X3 camera.

cnsumner:
This adds support for the 2:1 aspect ratio single-file format used in some recording modes of Insta360 cameras (X3, normal video mode is confirmed, other modes unsure but possible).

Removed transpose from the ffmpeg command for fetching frames because I found it yielded better results.

I am not a python dev so please critique if I've made any horrible mistakes in this code!

Resolves #1019

### DIFF
--- a/nerfstudio/process_data/insta360_utils.py
+++ b/nerfstudio/process_data/insta360_utils.py
@@ -66,6 +66,8 @@ def convert_insta360_to_images(
         video_back: Path to the back video.
         output_dir: Path to the output directory.
         num_frames_target: Number of frames to extract.
+        crop_percentage: Percentage used to calculate the cropped dimentions of extracted frames. Currently used to crop
+         out the curved portions of the fish-eye lens.
         verbose: If True, logs the output of the command.
     Returns:
         A tuple containing summary of the conversion and the number of extracted frames.
@@ -133,6 +135,8 @@ def convert_insta360_single_file_to_images(
         video: Path to the video.
         output_dir: Path to the output directory.
         num_frames_target: Number of frames to extract.
+        crop_percentage: Percentage used to calculate the cropped dimentions of extracted frames. Currently used to crop
+         out the curved portions of the fish-eye lens.
         verbose: If True, logs the output of the command.
     Returns:
         A tuple containing summary of the conversion and the number of extracted frames.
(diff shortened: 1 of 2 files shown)


nerfstudio/process_data/insta360_utils.py
```diff
@@ -123,3 +118,69 @@ def convert_insta360_to_images(
...
+    num_frames_target: int,
+    crop_percentage: float = 0.7,
+    verbose: bool = False,
+) -> Tuple[List[str], int]:
+    """Converts a video into a sequence of images.
+
+    Args:
+        video: Path to the video.
+        output_dir: Path to the output directory.
+        num_frames_target: Number of frames to extract.
```
tancik, line 135: Add arg for `crop_percentage`

tancik (commented):
Sorry for the delayed review (on holiday). I think the proposed solution in this PR works for now. I wrote some comments in #1019 for how we might better support these cameras in the future.

PR #1041, opened 2022-11-28
fix download command in readme

hturki:
FWIW I still get the following error when running the download command:

```
Access denied with the following error:

        Cannot retrieve the public link of the file. You may need to change
        the permission to 'Anyone with the link', or have had many accesses.

You may still be able to access the file from the browser:

         https://drive.google.com/uc?id=1dmjWGXlJnUxwosN6MVooCDQe970PkD-1
```

However manually downloading the file with the link at the bottom and placing the zip into the expected location seems to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ The following will train a _nerfacto_ model, our recommended model for real worl
 
 ```bash
 # Download some test data:
-ns-download-data nerfstudio --capture-name=poster
+ns-download-data --dataset nerfstudio --capture-name=poster
 # Train model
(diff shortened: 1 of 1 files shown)


ethanweber:
Hi @hturki, if you use the latest version on main, the command actually changed to `ns-download-data nerfstudio --capture-name=poster`. We changed the CLI slightly in PR #968 to make it more conducive to adding more datasets.

tancik:
The error you are running into is caused by too many people trying to download that file. The solution for that is for us to host the data somewhere other than google drive.

PR #1042, opened 2022-11-28
Allow users to skip using openrelay

tancik:
There seems to be an issue with some users if they are unable to connect to the TURN server. Seems to be related to - https://github.com/aiortc/aioice/issues/16
As a workaround, we now have the option to skip using openrelay servers suing the the command `--viewer.skip-openrelay`.


vinodrajendran001:
May I know where to add the command ```--viewer.skip-openrelay``` ?

tancik:
It can be added at the end of the command, ie,
`ns-train nerfacto --data DATA --viewer.skip-openrelay True`

PR #1055, opened 2022-11-30
Add support for RealityCapture

tancik:
Not fully working yet. Poses seem good for some datasets, but not all. Still debuging.


machenmusik:
Some quick questions
- Looking at RealityCapture export options, there are many; which CSV export option(s) are supported?
- Would it help to use non-CSV alignment export options such as Bundler v0.3?

machenmusik:
FYI, I found that this needed #1057 as well as same fix in another spot.

nerfstudio\data\dataparsers\nerfstudio_dataparser.py:215
`idx_tensor = torch.tensor(indices).type(torch.long)`

machenmusik:
(Thanks for better suggestion https://github.com/nerfstudio-project/nerfstudio/pull/1057#issuecomment-1332670064 )

machenmusik:
(Any update?)

tancik:
> (Any update?)

Sorry for the lack of updates, other tasks pushed this down the priority stack. It currently mostly works, but the scene is currently flipped and the quality seems a bit blurrier than expected. Im not sure if this is a problem with how I am processing he poses, or if the data I am testing with has issues. I am also unsure if the camera model I am using for distortions exactly matches the one that realitycapture uses. I hope to get back to this PR at somepoint, but if probably won't be anytime in the immediate future. Others are welcome to hop onto this PR to see if they can fix / answers some of the issues I've listed above.

machenmusik:
Did a quick test, something is definitely strange
- compared to Polycam, and COLMAP processing for same image set, RealityCapture values for focal length seem off
- using Polycam/COLMAP values with RealityCapture transforms, one gets much better quality, but it appears mirrored, so maybe LHS/RHS issue

nerfstudio/process_data/realitycapture_utils.py
```diff
@@ -0,0 +1,120 @@
...
+    img = np.array(Image.open(output_dir / image_filename_map[cameras["#name"][0].split(".")[0]]))
+    height, width, _ = img.shape
+
+    data["h"] = int(height)
+    data["w"] = int(width)
+
+    for i, name in enumerate(cameras["#name"]):
+        frame = {}
+        frame["file_path"] = image_filename_map[name.split(".")[0]].as_posix()
+        frame["fl_x"] = float(cameras["f"][i]) * width / 36
```
machenmusik, line 70: Note that for the output sample I tried, this and the following line needed to use height, not width.  Perhaps that is since the pictures were portrait, not landscape; not sure if the right thing to do is use max of width and height...

nerfstudio/process_data/realitycapture_utils.py
```diff
@@ -0,0 +1,120 @@
...
+        frame["p1"] = cameras["t1"][i]
+        frame["p2"] = cameras["t2"][i]
+
+        # Transform matrix to nerfstudio format. Please refer to the documentation for coordinate system conventions.
+        rot = _get_rotation_matrix(-float(cameras["heading"][i]), float(cameras["pitch"][i]), float(cameras["roll"][i]))
+
+        transform = np.eye(4)
+        transform[:3, :3] = rot
+        transform[:3, 3] = np.array([float(cameras["x"][i]), float(cameras["y"][i]), float(cameras["alt"][i])])
+        transform = np.array([[0, 1, 0, 0], [1, 0, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]) @ transform
```
machenmusik, line 89: Note that for the output sample I tried, the first two subarrays should be swapped so it is just the identity matrix

machenmusik (changes requested):
With the output sample I tried, and the changes mentioned, I was able to get a quite good nerfacto result, with the only annoyance being that I needed to rotate the view 180 degrees.

Obviously we want more than one sample to try in order to confirm.
(Unfortunately I am not at liberty to share my output sample.)

nerfstudio/process_data/realitycapture_utils.py
tancik: Let me know if you happen to try this with non portrait images. I can change an necessary.

machenmusik: With the CottageMiniature dataset (JPG_1 folder only), this yielded much better results
```
        frame["fl_x"] = float(cameras["f"][i]) * max(width, height) / 36
        frame["fl_y"] = float(cameras["f"][i]) * max(width, height) / 36
```
It was also necessary to avoid divide-by-zero from auto_scale_poses, which I will put into a small separate PR.

machenmusik: > having only 1 of 30 images chosen
For some reason, ns-process-data is copying all images but only downsampling the first one.

dkorolov:
I make some own debugging related RealityCaprure data import. We have some mess in nerfstudio related to the Orientation EXIF tag in source images and FFMPEG images autorotation behavior. This significantly influences on RealityCaprure data import.
I created a topic related to this
https://github.com/nerfstudio-project/nerfstudio/issues/1260

dkorolov:
I tested the autorotation issue on my data I can not share. But for public testing, we can use the Gingerbread Man dataset (it is free for everyone) - https://www.capturingreality.com/download/files/Calendar-gingerbread-man  
Also here are many free datasets that come with RealityCapture - https://www.capturingreality.com/free-datasets.

machenmusik (approved):

nerfstudio/process_data/realitycapture_utils.py
```diff
@@ -0,0 +1,119 @@
...
+        # TODO: Unclear how to get the principal point from RealityCapture, here a guess...
+        frame["cx"] = float(cameras["px"][i]) / 36.0 + width / 2.0
+        frame["cy"] = float(cameras["py"][i]) / 36.0 + height / 2.0
+        # TODO: Not sure if RealityCapture uses this distortion model
+        frame["k1"] = cameras["k1"][i]
+        frame["k2"] = cameras["k2"][i]
+        frame["k3"] = cameras["k3"][i]
+        frame["k4"] = cameras["k4"][i]
+        frame["p1"] = cameras["t1"][i]
+        frame["p2"] = cameras["t2"][i]
```
tancik, line 81: @machenmusik Does this help or hurt the reconstruction?

machenmusik: My private dataset and CottageMiniature both yield nice results with things as they are now, so at the moment it does not hurt.  Not sure I can say whether they help definitively.

machenmusik:
Thanks @dkorolov.  
The original issues with this PR branch appear to be resolved;
wondering if we want to clean up EXIF orientation metadata behavior in a subsequent PR.
@tancik thoughts?

machenmusik:
Some good news, looks like `-noautorotate` makes the Gingy sample process correctly.  Incoming separate PR.

machenmusik (approved):
When rebased against latest main including #1261, this PR now successfully processes data and trains against both CottageMiniature (ordinary images) and Gingy (rotated images) sample sets from RealityCapture, yielding nice looking results (at least to my eye).

tancik:
@machenmusik Thanks for helping out with this!

PR #1057, opened 2022-11-30
Fix IndexError runtime error

machenmusik:
fixes this error:

  File "C:\code\nerfstudio\nerfstudio\data\dataparsers\nerfstudio_dataparser.py", line 232, in _generate_dataparser_outputs
    distortion_params = torch.stack(distort, dim=0)[idx_tensor]
IndexError: tensors used as indices must be long, byte or bool tensors

### DIFF
--- a/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
+++ b/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
@@ -229,7 +229,7 @@ class Nerfstudio(DataParser):
                 p2=float(meta["p2"]) if "p2" in meta else 0.0,
             )
         else:
-            distortion_params = torch.stack(distort, dim=0)[idx_tensor]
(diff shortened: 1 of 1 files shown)


tancik (approved):
LGTM

machenmusik:
(Apologies for the force push; attribution was messed up, same one-liner change.)

tancik:
Do you think it would be better to just update line 215 with
`idx_tensor = torch.tensor(indices, dtype=torch.long)`
?

machenmusik:
That, or the variation in https://github.com/nerfstudio-project/nerfstudio/pull/1055#issuecomment-1332630814, would presumably work as well, is there a preference?  Happy to change the one-liner if so (and to do corresponding PR to realitycapture branch if needed)

tancik:
I think just the single edit would be good

PR #1058, opened 2022-11-30
Interpolation config for tcnn hashgrid

ydmalex:
Would like to compute the second derivative when using the tcnn hashgrid encoding as described in https://github.com/NVlabs/tiny-cuda-nn/issues/185.


tancik (approved):
LGTM!

tancik:
If you run `ns-dev-test` it will fix / give you info on the Core Test errors

ydmalex:
@tancik my bad, should be fixed now; seems like you might need to reapprove the workflows...

PR #1059, opened 2022-11-30
Add mesh export interface to viewer

ethanweber:
Adding the UI for mesh export.


machenmusik:
(This is now essentially #809 plus the two files in nerfstudio/viewer, right?)

PR #1065, opened 2022-12-01
Dnerf viewer and training fix

myaldiz:
Hi there!

I realized there are two bugs while working on `dnerf`.
First one is about training, caused by an incorrect shape of times tensor.
Second one is about the viewer. `times` return as zero when first connected to `viewer` and prevent entering the if clause.
Below two changes fixes it.

Thanks


tancik:
If you run `ns-dev-test` it will tell you what the error is. In this case it should be `is not None` instead of `!= None`

myaldiz (commented):
`is not None` change

PR #1067, opened 2022-12-01
Make friends_dataparser path separator agnostic (fixes #1066)

### LINKED ISSUE #1066: `ns-train semantic-nerfw` fails looking for images_4 png instead of jpg
**Describe the bug**
`ns-train semantic-nerfw` fails looking for image png instead of jpg

**To Reproduce**
1. Install nerfstudio.  (0.1.10, 0.1.11, latest main)
2. `ns-download-data friends` complains, so you must download from the GDrive link and extract into data/friends manually
3. `ns-train semantic-nerfw`
4. See error about PIL/Image.py -- but notice that it is looking for image png, rather than jpg which is present and specified in cameras.json

**Expected behavior**
Training should not fail.

machenmusik:
Fixes #1066 on Windows.

### DIFF
--- a/nerfstudio/data/dataparsers/friends_dataparser.py
+++ b/nerfstudio/data/dataparsers/friends_dataparser.py
@@ -116,12 +116,11 @@ class Friends(DataParser):
 
         # --- semantics ---
         if self.config.include_semantics:
+            empty_path = Path()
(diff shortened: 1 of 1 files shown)


tancik (approved):
LGTM, thanks for fixing!

Issue #1056, opened 2022-11-30
subprocess.CalledProcessError: Command '['ns-render', '--tyro-print-completion', 'bash']' returned non-zero exit status 1. While running nerfStudio from docker image


Napolean29:
[09:49:20] 🤷 .zshrc not found, skipping.                                                                 install.py:202
           🔍 Found .bashrc!                                                                              install.py:204
[09:49:21] ✔ Wrote new completion to /data/dl/nerfstudio/scripts/completions/bash/_ns-dev-test!           install.py:109
           ✔ Wrote new completion to /data/dl/nerfstudio/scripts/completions/bash/_ns-install-cli!        install.py:109
           ✔ Wrote new completion to /data/dl/nerfstudio/scripts/completions/bash/_ns-process-data!       install.py:109
[09:49:25] ✔ Wrote new completion to /data/dl/nerfstudio/scripts/completions/bash/_ns-eval!               install.py:109
           ✔ Wrote new completion to /data/dl/nerfstudio/scripts/completions/bash/_ns-download-data!      install.py:109
[09:49:29] ✔ Wrote new completion to /data/dl/nerfstudio/scripts/completions/bash/_ns-train!              install.py:109
Traceback (most recent call last):
  File "/usr/local/bin/ns-install-cli", line 8, in <module>
    sys.exit(entrypoint())
  File "/data/dl/nerfstudio/scripts/completions/install.py", line 274, in entrypoint
    tyro.cli(main, description=__doc__)
  File "/usr/local/lib/python3.8/dist-packages/tyro/_cli.py", line 125, in cli
    _cli_impl(
  File "/usr/local/lib/python3.8/dist-packages/tyro/_cli.py", line 326, in _cli_impl
    out, consumed_keywords = _calling.call_from_args(
  File "/usr/local/lib/python3.8/dist-packages/tyro/_calling.py", line 194, in call_from_args
    return unwrapped_f(*args, **kwargs), consumed_keywords  # type: ignore
  File "/data/dl/nerfstudio/scripts/completions/install.py", line 243, in main
    completion_paths = list(
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 619, in result_iterator
    yield fs.pop().result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 437, in result
    return self.__get_result()
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
[...]

PR #1068, opened 2022-12-01
More informative completion install error messages

brentyi:
I'm curious about the root cause of #1056, but it's difficult to debug without a useful error message.

To test this change, I slid an `assert False` into `scripts/render.py`:
image

Old error message just says that the subprocess failed:
image

New error message tells us why:
image


tancik (approved):
Lgtm

PR #1087, opened 2022-12-05
Update mipnerf.md

krahets:
1. Add  `--pipeline.datamanager.train-num-rays-per-batch` argument to docs for most of the users have to change the batch size to run this model.
2. Add `nerfstudio-data` argument. It seems like mipnerf set the Blender dataset as default.


ethanweber:
Thanks! I made a PR #1092 which resolves this issue.

PR #1093, opened 2022-12-05
Fix mask filenames misaligned after splitting set into train and eval

Zunhammer:
As mentioned in Discord the masks work only partly which is caused by a misalignment oif the filename indices after splitting the set into train and eval as the indices for the masks is not adjusted. Fixed this by simply applying the same indices to the masks as for the images.


tancik (approved):
Good catch! LGTM

PR #1098, opened 2022-12-07
Bug correction auto_orient_and_center_poses

fdarmon:
When using auto_orient_and_center_poses with method="none" the returned pose is 4x4 (contrary to up and pca) which triggers an assertion error later. This modification solves it.


tancik (approved):
LGTM, good catch!

PR #1108, opened 2022-12-11
change normals grad calculation

terrancewang:


tancik (approved):
LGTM. Do you see any changes in quality, speed, or other?

terrancewang:
> LGTM. Do you see any changes in quality, speed, or other?

Speed is the same and I didn't notice any diff in quality. Should be getting the same normals just using a different gradient function

PR #1109, opened 2022-12-11
Colab working with Python3.8

akristoffersen:
This is a temporary fix. I couldn't get the wheel to work so installing from the pytorch bindings on git. Slightly slower installation now.


tancik (approved):
Lgtm

Issue #1099, opened 2022-12-07
Volumetric sampling bug/problems in Instant-NGP implemtation


dozeri83:
Hi @tancik , 
****
tl;dr - there is a problem in the volumetric sampler resulting in a memory crash.
The cause of this issue is the initial sigma_fn method in the ray marching.
*****

I have been experimenting for a while in different configurations and methods in nerfstudio and stumbled upon weird issue. 
First of all I am using instant-ngp  (with AABB contraction) implementation since I can compare it visually to the original instant-ngp.

I am using blender hotdog data set with aabb 1. (on nerfstudio - scene_scale 0.5)

nerfstudio seem to crash on GPU memory because of this line
https://github.com/nerfstudio-project/nerfstudio/blob/6baf31dc9a9d81e6b44478bd31e0c0bc575493b2/nerfstudio/pipelines/dynamic_batch.py#L74
num_samples_per_batch - was very small.

the reason num_samples_per_batch was very small is because on the first iteration of the train the volumetric sampler
rejected the samples because of  sigma_fn ( the samples had zero density so it rejected all of them)

https://github.com/nerfstudio-project/nerfstudio/blob/e605b3344ac623fd96e2458ef98f8fbc7a97d718/nerfstudio/model_components/ray_samplers.py#L458

I had 2 workarounds - 
1. replaced  sigma_fn with None.
2. replaced sigma_fn with None for the first 100 iteration and then replaced it with  self.get_sigma_fn(rays_o, rays_d).

workaround 1 work best for me (uniform sampler?) with better PSNR.

BTW - nerfacto model looked very bad on the blender hotdog data set. ( but this is a different issue)

dozeri83:
Reconstruction:

ns-train instant-ngp --trainer.max-num-iterations 35000 --vis tensorboard  --experiment-name transforms_train_scaled_0_3333333 --output_dir output/hotdog --timestamp 23 --data nerf_sythetic/hotdog/  --orientation-method none --center-poses False --auto-scale-poses False --scene_scale 0.5

where I scaled the transforms file by 0.3333 (ngp scale) and used nerstudio loader:

[transforms_train_scaled_0_3333333.zip](https://github.com/nerfstudio-project/nerfstudio/files/10177371/transforms_train_scaled_0_3333333.zip)

tancik:
[...]

PR #1116, opened 2022-12-13
instant-ngp-bounded model

dozeri83:
Hi,
 
@tancik , as we discussed in #1099 , I inerested a new method based on instant-ngp (instant-ngp-boudned) that is designated for  bounded scenes.

I tried to tackle several problems in this Fork:

1. evaluation of instant-ngp-bounded on blender data set (using the benchmarking scripts)  looks way better (PSNR ranges 31-36 db)
2.  bounded real scene looks better and GUI looks similar to instant-ngp ( black background)
3. crashing bug in instant-ngp in cased of small aabb of size 1.
4. instant-ngp and nerfstudio data-parsers  can now handle any transformation file  instead of hardcoded transforms.json .
i.e --data directoy/my_trans_file.json . (backward compatible with plain directory)

There are several open issues and problems  that I encountered with thatI  want to open:

1. ignore pixels/ray that have zeros in thier alpha channel.  
  the original instnat-ngp code ignores them (mask them away) and I think this is one of the reasons that instant ngp looks  better on blender. (+2 PSNR more than us). I tried to implement this but it did not look good for some reason.
  
  2. samping bug in the volumetric sampler (#1099)  - when we use the volumetric sampler with density, the density rejected way too many good samples. I think that we can not rely fully on the density and should implement a probabilistic sampling that combines both the current density and uniform/non uniform sampling. In  the mean time I replaced the density with None. ( only relevant for the insant-ngp-bounded)

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ ns-download-data nerfstudio --capture-name=poster
 # Train model
 ns-train nerfacto --data data/nerfstudio/poster 
 ```
-or
+The last command-line can be also replaced by referring explicitly to the transformation file:
 
 ```bash
 ns-train nerfacto --data data/nerfstudio/poster/transforms.json
--- a/README.md
+++ b/README.md
@@ -131,7 +131,13 @@ The following will train a _nerfacto_ model, our recommended model for real worl
 # Download some test data:
 ns-download-data nerfstudio --capture-name=poster
 # Train model
-ns-train nerfacto --data data/nerfstudio/poster
+ns-train nerfacto --data data/nerfstudio/poster 
+```
+or
+
+```bash
+ns-train nerfacto --data data/nerfstudio/poster/transforms.json
+
 ```
 
 If everything works, you should see training progress like the following:
--- a/README.md
+++ b/README.md
@@ -131,13 +131,7 @@ The following will train a _nerfacto_ model, our recommended model for real worl
 # Download some test data:
 ns-download-data nerfstudio --capture-name=poster
 # Train model
-ns-train nerfacto --data data/nerfstudio/poster 
-```
-The last command-line can be also replaced by referring explicitly to the transformation file:
-
-```bash
-ns-train nerfacto --data data/nerfstudio/poster/transforms.json
-
+ns-train nerfacto --data data/nerfstudio/poster
 ```
 
 If everything works, you should see training progress like the following:
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -116,7 +116,7 @@ method_configs["instant-ngp"] = ExperimentConfig(
 )
 
 
-method_configs["instant-ngp-bounded"] = Config(
+method_configs["instant-ngp-bounded"] = ExperimentConfig(
     method_name="instant-ngp-bounded",
     trainer=TrainerConfig(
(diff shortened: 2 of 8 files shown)


nerfstudio/configs/method_configs.py
```diff
@@ -56,6 +60,7 @@
 descriptions = {
     "nerfacto": "Recommended real-time model tuned for real captures. This model will be continually updated.",
     "instant-ngp": "Implementation of Instant-NGP. Recommended real-time model for bounded synthetic data.",
```
tancik, line 62: Lets change this to `Recommended for real unbounded scenes`

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -258,21 +268,22 @@ def _generate_dataparser_outputs(self, split="train"):
         )
         return dataparser_outputs
 
-    def _get_fname(self, filepath: PurePath, downsample_folder_prefix="images_") -> Path:
+    def _get_fname(self, filepath: PurePath, data_dir: PurePath, downsample_folder_prefix="images_") -> Path:
         """Get the filename of the image file.
         downsample_folder_prefix can be used to point to auxillary image data, e.g. masks
+        data_dir: the directory of the data that contains the transform file
```
tancik, line 274: I realize this was an existing issue, but can you add doc for the missing arguments.

nerfstudio/models/instant_ngp.py
```diff
@@ -122,7 +122,7 @@ def populate_modules(self):
         )
 
         # renderers
-        background_color = "random" if self.config.randomize_background else colors.WHITE
+        background_color = "random" if self.config.randomize_background else colors.BLACK
```
tancik, line 125: It might be better to switch `self.config.randomize_background` to `self.config.background_color` which takes the literal `["random", "white", "black"]`

tancik (commented):
Thanks for doing a deeper dive into ngp + bounded scenes. This is an area that has not received much development attention.

README.md
```diff
@@ -131,7 +131,13 @@ The following will train a _nerfacto_ model, our recommended model for real worl
...
 ns-download-data nerfstudio --capture-name=poster
 # Train model
-ns-train nerfacto --data data/nerfstudio/poster
+ns-train nerfacto --data data/nerfstudio/poster 
+```
+The last command-line can be also replaced by referring explicitly to the transformation file:
+
+```bash
+ns-train nerfacto --data data/nerfstudio/poster/transforms.json
+
```
tancik, line 140: Let's avoid adding this to the readme, I think the readme should be as simple as possible. This info should exist in the config help text ie. - https://github.com/nerfstudio-project/nerfstudio/blob/1a70665e76bfbc71ef193cc6d3429efa582ddaff/nerfstudio/data/dataparsers/nerfstudio_dataparser.py#L50

dozeri83: @tancik ,what am I doing wrong with the Readme ? I reverted the changes and still dont pass the doc builds.
locally I pass on my station.

tancik: It seems to be a bug with readthedocs. This just started happening to other PRs. Ignore for now.

tancik: Once you add the `.json` optional comments to the dataparsers, I think this PR will be ready to go. I will need to force merge it on my end.

tancik (approved):

PR #1126, opened 2022-12-14
Changed tinycudann to wheel to work with py3.8

akristoffersen:
Speeding up installation time with a wheel. Had to go down to tinycudann v1.5, but should still work out fine.


tancik (approved):
Lgtm

PR #1135, opened 2022-12-16
adding data subset

ethanweber:
Creating a list of captures that work reasonably well. Also starting to move away from google drive.


tancik (approved):
LGTM

PR #1142, opened 2022-12-18
Support grayscale-only images in dataset

mabl:
Some datasets may only be grayscale. In that case, it is easiest to simply convert them into RGB.


tancik (approved):
LGTM

tancik:
Run `ns-dev-test` to fix the errors.

PR #1148, opened 2022-12-19
[Fix] point cloud export broken in Dockerfile

### LINKED ISSUE #1129: Error running ns-export in docker
I have cloned this repo [https://github.com/nerfstudio-project/nerfstudio] and i was trying to build the docker file within the repo (successful) and while doing `ns-export` on terminal, i get the following error which i fail to resolve: "Cannot mix incompatible Qt library (5.12.8) with this library (5.15.2)" Can you please help me resolve this issue. 

The only change i made in the docker file was to change `ENV TCNN_CUDA_ARCHITECTURES=86` to `ENV TCNN_CUDA_ARCHITECTURES=80` to get `ns-train` to work.

Zunhammer:
Rebased the docker image to one built on ubuntu 22.04 where the libraries are matching the required versions for pymeshlab. Pointcloud export is now working again.

The new image comes with some additional changes (only a side node):
Python is now version 3.10
CUDA is now version 11.7
Colmap needs to be compiled with c++14

Didn't find any issues with the new versions so far.

Fixes #1129


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Define base image.
-FROM nvidia/cudagl:11.3.1-devel
+FROM nvidia/cuda:11.7.1-devel-ubuntu22.04
```
tancik, line 2: Will this still work with windows?

tancik (commented):
Thanks for looking into this. The documentation should also be updated to reflect the new prerequisites. -nvidia/cuda:11.7.1-devel-ubuntu22.04

Dockerfile
Zunhammer: Yes, also the old image was based on Ubuntu (version 20.04), you simply need the docker environment based on WSL and Nvidia drivers as before.
Actually, this shouldn't make any difference from the user perspective.

tancik: Gotcha, (docker beginner here)

gowriaddepalli:
When will this merge in the main branch?

gowriaddepalli:
How to get version of python 3.10 + cuda 11.6 to run on google colab demo for nerfstudio? Any idea? I can change cuda 11.3 to 11.6 on google colab. Please let me know. Thanks in advance. It can help me fix ns-export issue maybe @tancik

Zunhammer:
> How to get version of python 3.10 + cuda 11.6 to run on google colab demo for nerfstudio? Any idea? I can change cuda 11.3 to 11.6 on google colab. Please let me know. Thanks in advance. It can help me fix ns-export issue maybe @tancik

Sorry, I have no experience with Colab, also I don't see this related to the Dockerfile/docker image.

Zunhammer:
> Thanks for looking into this. The documentation should also be updated to reflect the new prerequisites.

True, I updated the description and added some more hints there. The only requirement that chaned is actually the nvidia driver which needs to be compatible with CUDA 11.7.

tancik (approved):
LGTM

Issue #1149, opened 2022-12-19
Consider adding 360 video metadata to equirectangular render export MP4 files


machenmusik:
e.g. https://github.com/google/spatial-media/tree/v2.0

machenmusik:
Thanks for looking into it...

http://p.cweiske.de/621 says that exiftool doesn't put in the correct uuid atom, but that appears to be no longer the case; exiftool 12.42 can add the proper tags with `exiftool -XMP-GSpherical:Spherical="true" -XMP-GSpherical:Stitched="true" -XMP-GSpherical:StitchingSoftware=dummy -XMP-GSpherical:ProjectionType=equirectangular file.mp4` now.

So that could make https://github.com/smarnach/pyexiftool a simple option, but I am not sure whether the licensing is acceptable for the nerfstudio project.

machenmusik:
(Went a different route; edited the MP4 file the old-fashioned way.)

machenmusik:
Closed via #1151

PR #1151, opened 2022-12-20
when camera_type is equirectangular, add spherical metadata to rendered video

machenmusik:
Implements #1149.

### DIFF
--- a/scripts/render.py
+++ b/scripts/render.py
@@ -8,6 +8,7 @@ import json
 import os
 import struct
 import sys
+from contextlib import ExitStack
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import List, Optional
@@ -62,7 +63,7 @@ def _render_trajectory_video(
         render_width: Video width to render.
         render_height: Video height to render.
     """
-    CONSOLE.print("[bold green]Creating trajectory video")
+    CONSOLE.print("[bold green]Creating trajectory " + output_format)
     cameras.rescale_output_resolution(rendered_resolution_scaling_factor)
     cameras = cameras.to(pipeline.device)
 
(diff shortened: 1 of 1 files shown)


scripts/render.py
```diff
@@ -42,6 +44,7 @@ def _render_trajectory_video(
     rendered_resolution_scaling_factor: float = 1.0,
     seconds: float = 5.0,
     output_format: Literal["images", "video"] = "video",
+    camera_type: str = "",
```
tancik, line 47: Can you make this a Literal and add it to the doc string below.

tancik (commented):
Thanks for digging into the details for this functionality. It's a great addition!

scripts/render.py
machenmusik: I discovered the CameraType enum and switched to that instead.

machenmusik:
(Hang on, more changes incoming)

machenmusik:
It seemed inefficient for all rendered images to be held in memory just to encode video, so that is no longer done.

As a curious side effect, I noticed that non-filename trajectories such as spiral don't have a defined render width or height.  That didn't matter before when passing all the images at once to be encoded,, but it does now that we create the video writer before we have the first image (and its size).  The values for Nerfacto are currently the default for spiral trajectory, but if there is a better way to know what resolution should be expected, that would be preferable.

For later perhaps: it should be possible to reuse camera path files, but override the width and height for a given render command.

machenmusik:
(okay, finally got isort / black / pylint to agree...)

scripts/render.py
```diff
@@ -53,9 +59,11 @@ def _render_trajectory_video(
         rendered_resolution_scaling_factor: Scaling factor to apply to the camera image resolution.
         seconds: Length of output video.
         output_format: How to save output data.
+        camera_type: Camera projection format type.
+        render_width: Video width to render.
```
tancik, line 63: Nit: doc order should match definition order

tancik (approved):
LGTM, this restructure looks good to me. Thanks for working on this.

scripts/render.py
machenmusik: Must have been one of the isort / black / pylint fixes, forgot to update docs

tancik:
Looks good, feel free to merge. You should now have access (check your email)

machenmusik:
Done thx

PR #1156, opened 2022-12-21
work around mismatch to allow rgb depth side-by-side output

machenmusik:


machenmusik:
(You'll want to look at `Files changed` as a few superfluous merge commits seem to be in my fork.)

Example invocation:
`ns-render --load-config outputs\data\your\path\here\config.yml --traj spiral --seconds 1 --rendered-output-names rgb depth --output-path renders\render_output_rgbd_sbs.mp4`

scripts/render.py
```diff
@@ -109,6 +113,8 @@ def _render_trajectory_video(
                         )
                         sys.exit(1)
                     output_image = outputs[rendered_output_name].cpu().numpy()
+                    if rendered_output_name != "rgb" and rendered_output_names[0] == "rgb":
+                        output_image = np.concatenate((output_image,) * 3, axis=-1)
```
tancik, line 117: Can we make this more general? Ie check if the output image is 1 channel and expand if necessary? I think we should avoid output_name specific logic as much as possible

machenmusik: Open to suggestions?  I don't disagree, but at the moment the only multi-channel one seems to be `rgb` and I didn't want to overcomplicate things e.g. `depth rgb` rather than `rgb depth`.  If there is a lookup table to go from output name to number of channels, that would be useful here.

tancik: For now can you just change the if statement to something like `if output_image.shape[-1] == 1:`. Agreed that a larger refactor should eventually be done. One thing we've discussed is making the outputs a dataclass with additional info like num channels, and colormaps. But that is for a future PR.

machenmusik: Is it OK to assume rgb is in there, and so 1-channel should become 3-channel?
If not, maybe we could settle on something like this for now...
`if rendered_output_names[0] == "rgb" and output_image.shape[-1] == 1:`

machenmusik: ah, I see that 1-channel (e.g. only `depth`) doesn't work anyway.

tancik (changes requested):
LGTM

machenmusik:
(GitHub still says changes requested - feel free to approve, or just force merge)

tancik (approved):

PR #1163, opened 2022-12-23
PixelSampler refactor

akristoffersen:
moved collate methods as class methods, so sampling is only thing changed.

This is useful for future pixel sampling updates, as you don't need to write a whole new collate method if the only thing changing is the pixel sampling method.

### DIFF
--- a/nerfstudio/data/pixel_samplers.py
+++ b/nerfstudio/data/pixel_samplers.py
@@ -17,120 +17,10 @@ Code for sampling pixels.
 """
 
 import random
-from typing import Dict
+from typing import Dict, Optional, Union
 
 import torch
-
-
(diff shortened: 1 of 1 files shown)


nerfstudio/data/pixel_samplers.py
```diff
@@ -160,12 +184,12 @@ def sample(self, image_batch: Dict):
             image_batch: batch of images to sample from
         """
         if isinstance(image_batch["image"], list):
-            image_batch = dict(image_batch.items())  # copy the dictionary so we don't modify the original
-            pixel_batch = collate_image_dataset_batch_list(
+            image_batch = dict(image_batch.items())  # copy the dictioary so we don't modify the original
```
tancik, line 187: nit: "dictionary"

tancik (approved):
LGTM

PR #1175, opened 2022-12-26
TrainerConfig inherit from InstantiateConfig

jake-austin:
To get around dreamfusion needing a custom backward method and having to add a hack for it, I'm trying to make the TrainerConfig inherit from InstantiateConfig so we can create custom Trainer classes that can be hotswapped in and out in the method_configs.py file.


ethanweber:
@jake-austin Does the Trainer have new logic needed for DreamFusion or can it be done with a custom Pipeline instead?

jake-austin:
I think it's probably best to just have trainer classes that we can hotswap in tbh, since the other option in my mind would be refactoring the pipeline to have the optimizers and have a function to do one complete training step, which sort of would leave the trainer barebones and might take us a little bit further from the HuggingFace style pipeline that we currently have.

tancik:
The docs also need updated, ie https://docs.nerf.studio/en/latest/quickstart/first_nerf.html#resume-from-checkpoint-visualize-existing-run uses `--training.load-dir` cmd. There may be others.

tancik (approved):
LGTM

PR #1187, opened 2022-12-30
upgrade nerfacc to 0.3.x with CUB accel

liruilong940607:
White -- nerfacc 0.3.x.
Blue -- nerfacc 0.2.x.

nerfacc 0.3.x uses CUB to accelerate the accumulation (both fwd and bwd) when available (cuda toolkit >= 11.5). Test rays / secs got ~10% speed up. For some reason the training speed is slightly slower with nerfacc 0.3.x, which needs to be further investigated.

Note the CUB acceleration is more noticeable when the #samples per ray is less, so the gap for test rays / secs got larger when converge. But it's unclear what's the reason of the slow down during training.

Screen Shot 2022-12-30 at 11 34 16 PM
Screen Shot 2022-12-30 at 11 34 34 PM

### DIFF
--- a/nerfstudio/model_components/ray_samplers.py
+++ b/nerfstudio/model_components/ray_samplers.py
@@ -431,7 +431,7 @@ class VolumetricSampler(Sampler):
         near_plane: float = 0.0,
         far_plane: Optional[float] = None,
         cone_angle: float = 0.0,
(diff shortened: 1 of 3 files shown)


liruilong940607:
The quality is numerically the same as before.

tancik (approved):
LGTM, thanks for updating!

Issue #1051, opened 2022-11-29
Visualize / export refined camera poses


andrearama:
**Is your feature request related to a problem? Please describe.**
When using Nerfacto the camera poses are being refined, which is great since in many real-world applications the poses are inaccurate.
However, I do not see an easy solution to either visualize or save the refined camera poses. 
I might for example need this for later uses. One example being rendering a video using the estimated camera poses

**Describe the solution you'd like**
Either a flag or a button on the GUI to save the estimated poses.

**Describe alternatives you've considered**
Save them from within the pipeline myself, but it gets a bit messy

akristoffersen:
Hi! This is a great suggestion. I'll try to get something out on this in the coming week.

Saving the poses should be relatively straight forward, but adding the pose movement into the visualizer might be a bit more complicated so I can work on that next.

mackopes:
@akristoffersen Do you think it would make sense to export the optimised camera poses as part of the `ns-export` script?

mackopes:
I implemented the camera exporter in this [PR](https://github.com/nerfstudio-project/nerfstudio/pull/1191)

Issue #1017, opened 2022-11-24
Camera optimizer didn't work


Xiaxia1997:
@benji-york @corbt @erich666 @akanazawa @decrispell 

**Describe the bug**
I've tried the pose refinement for [storefront](https://drive.google.com/file/d/16b792AguPZWDA_YC4igKCwXJqW0Tb21o/view?usp=sharing) dataset and got the result as below. Noise mentioned is `--pipeline.datamanager.camera-optimizer.position-noise-std 0.15`. From the chart, I'm confused about the pose refinement can not recover psnr from 21 to 28. Moreover, the pose refinement works nothing when noise added.

| condition                   | psnr |
| --------------------------- | ---- |
| without pose refinement     | 28.3 |
| pose refinement             | 28.7 |
| add noise only              | 21.3 |
| add noise + pose refinement | 21.2 |

1. So my first question is about the pose refinemen, did I use it appropriately? Here is my command:

```
# add noise, camera-optimizer
ns-train nerfacto --data=/data1/nerfstudio/data/storefront/   --vis tensorboard --viewer.websocket-port=5082 --pipeline.model.proposal-net-args-list.0.hidden-dim 64 --pipeline.model.proposal-net-args-list.0.log2-hashmap-size 22 --pipeline.model.proposal-net-args-list.0.num-levels 16 --pipeline.model.proposal-net-args-list.0.max-res 2048 --pipeline.datamanager.camera-optimizer.mode SO3xR3 --pipeline.datamanager.camera-optimizer.position-noise-std 0.15 nerfstudio-data 
# add noise, no camera-optimizer
ns-train nerfacto --data=/data1/nerfstudio/data/storefront/   --vis tensorboard --viewer.websocket-port=5082 --pipeline.model.proposal-net-args-list.0.hidden-dim 64 --pipeline.model.proposal-net-args-list.0.log2-hashmap-size 22 --pipeline.model.proposal-net-args-list.0.num-levels 16 --pipeline.model.proposal-net-args-list.0.max-res 2048 --pipeline.datamanager.camera-optimizer.position-noise-std 0.15  --pipeline.datamanager.camera-optimizer.mode off  nerfstudio-data 
```

2. Lack of pose alignment:

[...]

PR #1191, opened 2022-12-30
Camera export

mackopes:
The feature to export camera poses has already been requested by a few people (e.g. [issue-1051](https://github.com/nerfstudio-project/nerfstudio/issues/1051)).

I would also find this feature handy.

In this PR I implemented the exporter for camera poses. Currently, it exports the camera poses as two files: `transforms_train.json` and `transforms_eval.json` for train and eval poses.

the structure of these files is
```
[
  {
    "file_path": {image file path corresponding with this camera as in the original transforms.json},
    "transform": {4x4 List of floats}
  },
  ...
]
```

Even though we can export more information in the .json files, I find the these to be sufficient (at least for my use case). User can use the `file_path` to match the exported camera poses with the input camera poses.

I am happy to add more information to the .json files if there will be need for them, but looks like right now people would like to see mainly the optimised transforms.

I tested this change with
```
$ ns-export cameras --load-config {path to my config} --output_dir cam_poses
```

### DIFF
--- a/nerfstudio/data/datasets/base_dataset.py
+++ b/nerfstudio/data/datasets/base_dataset.py
@@ -115,3 +115,6 @@ class InputDataset(Dataset):
     def __getitem__(self, image_idx: int) -> Dict:
         data = self.get_data(image_idx)
         return data
+
+    def get_image_filenames(self) -> List[Path]:
+        return self._dataparser_outputs.image_filenames
--- a/nerfstudio/data/datasets/base_dataset.py
+++ b/nerfstudio/data/datasets/base_dataset.py
@@ -117,13 +117,11 @@ class InputDataset(Dataset):
         data = self.get_data(image_idx)
         return data
 
-    def get_image_filenames(self) -> List[Path]:
-        """Returns image filenames for this dataset.
+    @property
+    def image_filenames(self) -> List[Path]:
+        """
+        Returns image filenames for this dataset.
         The order of filenames is the same as in the Cameras object for easy mapping.
-
-        Args: None
-        Returns:
-            List of image filenames.
         """
 
         return self._dataparser_outputs.image_filenames
--- a/nerfstudio/data/datasets/base_dataset.py
+++ b/nerfstudio/data/datasets/base_dataset.py
@@ -18,7 +18,8 @@ Dataset.
 from __future__ import annotations
 
 from copy import deepcopy
-from typing import Dict
+from pathlib import Path
+from typing import Dict, List
 
 import numpy as np
 import numpy.typing as npt
@@ -117,4 +118,12 @@ class InputDataset(Dataset):
         return data
 
     def get_image_filenames(self) -> List[Path]:
+        """Returns image filenames for this dataset.
+        The order of filenames is the same as in the Cameras object for easy mapping.
+
+        Args: None
+        Returns:
+            List of image filenames.
+        """
+
         return self._dataparser_outputs.image_filenames
--- a/nerfstudio/exporter/exporter_utils.py
+++ b/nerfstudio/exporter/exporter_utils.py
@@ -276,7 +276,7 @@ def collect_camera_poses_for_dataset(dataset: Optional[InputDataset]) -> List[Di
 
     frames: List[Dict[str, Any]] = []
 
-    # new cameras are in cameras, whereas image paths are in dataparser_outputs
+    # new cameras are in cameras, whereas image paths are stored in a private member of the dataset
     for idx in range(len(cameras)):
         image_filename = image_filenames[idx]
         transform = cameras.camera_to_worlds[idx].tolist()
--- a/nerfstudio/exporter/exporter_utils.py
+++ b/nerfstudio/exporter/exporter_utils.py
@@ -22,7 +22,7 @@ from __future__ import annotations
 
 import sys
 from dataclasses import dataclass
-from typing import List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 import numpy as np
 import open3d as o3d
@@ -40,7 +40,7 @@ from torchtyping import TensorType
 
 from nerfstudio.cameras.cameras import Cameras
 from nerfstudio.data.datasets.base_dataset import InputDataset
-from nerfstudio.pipelines.base_pipeline import Pipeline
+from nerfstudio.pipelines.base_pipeline import Pipeline, VanillaPipeline
 from nerfstudio.utils.rich_utils import ItersPerSecColumn
 
 CONSOLE = Console(width=120)
@@ -257,12 +257,13 @@ def render_trajectory(
             depths.append(outputs[depth_output_name].cpu().numpy())
     return images, depths
 
+
 def collect_camera_poses_for_dataset(dataset: Optional[InputDataset]) -> List[Dict[str, Any]]:
     if dataset is None:
         return []
 
     cameras = dataset.cameras
-    image_filenames = dataset._dataparser_outputs.image_filenames
+    image_filenames = dataset.get_image_filenames()
 
     frames: List[Dict[str, Any]] = []
 
@@ -270,12 +271,15 @@ def collect_camera_poses_for_dataset(dataset: Optional[InputDataset]) -> List[Di
     for idx in range(len(cameras)):
         image_filename = image_filenames[idx]
         transform = cameras.camera_to_worlds[idx].tolist()
-        frames.append({
-            "file_path": str(image_filename),
-            "transform": transform,
-        })
+        frames.append(
+            {
+                "file_path": str(image_filename),
+                "transform": transform,
+            }
+        )
 
     return frames
+
 
 def collect_camera_poses(pipeline: VanillaPipeline) -> Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]:
     """Collects camera poses for train and eval datasets.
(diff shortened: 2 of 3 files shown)


nerfstudio/exporter/exporter_utils.py
```diff
@@ -255,3 +256,44 @@ def render_trajectory(
             images.append(outputs[rgb_output_name].cpu().numpy())
             depths.append(outputs[depth_output_name].cpu().numpy())
     return images, depths
+
+def collect_camera_poses_for_dataset(dataset: Optional[InputDataset]) -> List[Dict[str, Any]]:
```
tancik, line 260: Can you add a docstring with `Args` and `Returns`

tancik (commented):
Thanks for adding! You can run `ns-dev-test` to fix the github actions error.

mackopes:
Hi @tancik,
thanks a lot for the review. I addressed all your comments and fixed the linter errors. Could you please give it 👀 one more time?

nerfstudio/data/datasets/base_dataset.py
```diff
@@ -115,3 +116,14 @@ def get_metadata(self, data: Dict) -> Dict:
     def __getitem__(self, image_idx: int) -> Dict:
         data = self.get_data(image_idx)
         return data
+
+    def get_image_filenames(self) -> List[Path]:
```
tancik, line 120: This could be a property, https://www.programiz.com/python-programming/property

If it's a property, you can remove the `Args` and `Returns`

tancik (commented):
Almost there. Added one small comment.

nerfstudio/data/datasets/base_dataset.py
mackopes: Good idea.

mackopes:
Good point you had. I updated the code as you suggested

jingyibo123:
I don't thinks this correctly exports **optimized** pose of train dataset

To do that, one should read the checkpoint and adjust dataset's poses with camera_opt state.

Correct me if I'm mistaken

idouros:
This is an extremely valuable feature, thanks! Any chance of an estimate of when we can expect to see it in a full release of a next version of nerfstudio?

tancik:
@idouros it is in the current pip version of nerfstudio.

idouros:
Excellent, thanks!

I have another question (mainly for @mackopes , but everybody else also feel free if you wish, and apologies if it sounds a bit silly):

Here it goes: if the transforms in `transforms_train.json` and `transforms_eval.json` are the camera poses, then what are the transforms in `transforms.json`? I understand that (possibly) one of the two sets describes the camera extrinsics for each camera view, but I am not sure which one.

Let me explain what I am trying to do:

I have a set of images, all of them frames from the same video (i.e. camera intrinsics are the same for all images), that I am using to train a nerf model of a scene. From that I export a point cloud. So far so good. What I want to do is to produce, for each training image, a render of the point cloud from the corresponding camera's position/pose so that it matches the training image. The steps that I thought I should follow in order to achieve that are:

1. Construct the intrinsic projection matrix (using the camera params picked up from `transforms.json`)
2. Apply the inverse of the dataparser transform to the point cloud. Then, for each image:
3. Apply the extrinsic transform (i.e. the one from the exported camera poses as per this PR) to the point cloud.
4. Apply the projection transform (from step #1 above) to get the 2D positions of the points, clip the ones outside image boundaries (info in `transforms.json`) and render the rest.

I was expecting that this would produce an image that matches the original (ok, subject to camera distortion params), but I seem to be way off. So, there is obviously something wrong in my logic that I cannot spot, let alone resolve. Any hints would be awesome. Many thanks!

idouros:
I have also added a comment here: https://github.com/nerfstudio-project/nerfstudio/issues/1017#issuecomment-1429840080 as I do not seem to be the only one trying to do this. (Note: even when I train with the camera optimization off, I still do not get much luck).

MaxChanger:
> I don't thinks this correctly exports **optimized** poses of train dataset
> 
> To do that, one should read the checkpoint and adjust dataset's poses with `'datamanager.train_camera_optimizer.pose_adjustment'`.
> 
> Correct me if I'm mistaken

Hi, @jingyibo123.  Have you made any progress? After debug checking,  I also think that the poses json of ns-export may only be the pose of recenter+rescale, not the pose optimized by network training.

jingyibo123:
I have a version working, it's not complicated, but I still need to design some sort of test to validate the output...

MaxChanger:
@jingyibo123 Hi, I found that there is an array of pose_adjustment in the saved ckpt, which can be used to transform the original pose. This might help.

mackopes:
Hello everyone (@jingyibo123, @MaxChanger , @idouros ),
I think you might be right and this indeed returns only after recenter+rescale. I will take a look at it more in depth soon!

yqdch:
> I have a version working, it's not complicated, but I still need to design some sort of test to validate the output...

Hello everyone, @jingyibo123 @mackopes @MaxChanger @tancik , is there a right way to extract the optimized poses in nerfacto now?

yqdch:
> > I have a version working, it's not complicated, but I still need to design some sort of test to validate the output...
> 
> Hello everyone, @jingyibo123 @mackopes @MaxChanger @tancik , is there a right way to extract the optimized poses in nerfacto now?

I have a worked implenmentation, just render refined pose train images on render.py script.

vijendra1125:
> I don't thinks this correctly exports **optimized** poses of train dataset
> 
> To do that, one should read the checkpoint and adjust dataset's poses with `'datamanager.train_camera_optimizer.pose_adjustment'`.
> 
> Correct me if I'm mistaken

I tried to run same training with and without camera pose optimization and exported the camera poses after training. In both cases I get same same camera poses. Based on that I also think that exported camera pose is not optimized camera pose. 
@jingyibo123 did you get the implementation to export optimized pose working?

vijendra1125:
I modified these two functions in nerfstudio/exporter/export_utils.py to get optimized pose:

```
def collect_camera_poses_for_dataset(dataset: Optional[InputDataset], camera_optimizer: Optional[CameraOptimizer]) -> List[Dict[str, Any]]:
    """Collects rescaled, translated and optimised camera poses for a dataset.

    Args:
        dataset: Dataset to collect camera poses for.

    Returns:
        List of dicts containing camera poses.
    """

    if dataset is None:
        return []

    cameras = dataset.cameras
    image_filenames = dataset.image_filenames
    if camera_optimizer is not None:
        camera_indices = IntTensor(range(len(cameras)))
        optimized_camera_to_camera_transforms = camera_optimizer(camera_indices).cpu().detach().numpy()

    frames: List[Dict[str, Any]] = []

    # new cameras are in cameras, whereas image paths are stored in a private member of the dataset
    for idx in range(len(cameras)):
        image_filename = image_filenames[idx]
        transform_c2w_3x4 = cameras.camera_to_worlds[idx]
        
        # get optimized camera pose if camera optimizer is provided
        if camera_optimizer is not None:
            transform_w2c_4x4 = np.linalg.inv(np.vstack((transform_c2w_3x4, np.array([0, 0, 0, 1]))))
            camera_to_optimized_camera_transform = np.linalg.inv(np.vstack((optimized_camera_to_camera_transforms[idx], np.array([0, 0, 0, 1]))))
            transform_optimized_c2w_4x4 = np.linalg.inv(camera_to_optimized_camera_transform@transform_w2c_4x4)
            transform_c2w_3x4 = transform_optimized_c2w_4x4[:3, :]
        
        frames.append(
            {
                "file_path": str(image_filename),
                "transform": transform_c2w_3x4.tolist(),
            }
        )

    return frames


def collect_camera_poses(pipeline: VanillaPipeline) -> Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]:
    """Collects camera poses for train and eval datasets.

    Args:
        pipeline: Pipeline to evaluate with.

    Returns:
        List of train camera poses, list of eval camera poses.
    """

    train_dataset = pipeline.datamanager.train_dataset
    assert isinstance(train_dataset, InputDataset)
    train_optimizer = pipeline.datamanager.train_camera_optimizer
    assert isinstance(train_optimizer, CameraOptimizer)

    eval_dataset = pipeline.datamanager.eval_dataset
    assert isinstance(eval_dataset, InputDataset)
    if hasattr(pipeline.datamanager, "eval_camera_optimizer"):
        eval_optimizer = pipeline.datamanager.eval_camera_optimizer
        assert isinstance(eval_optimizer, CameraOptimizer)
    else:
        eval_optimizer  = None

    train_frames = collect_camera_poses_for_dataset(train_dataset, train_optimizer)
    eval_frames = collect_camera_poses_for_dataset(eval_dataset, eval_optimizer)

    return train_frames, eval_frames
```

jingyibo123:
Unfortunately I can't make a PR because of corporate policy, Here's a snippet that's been verified to work @vijendra1125 @MaxChanger @mackopes .

```python
        train_opt_frames = train_frames
        checkpoint = torch.load(checkpoint_path)
        pose_adjustment = checkpoint['pipeline']['datamanager.train_camera_optimizer.pose_adjustment']
        from nerfstudio.cameras.lie_groups import exp_map_SO3xR3
        # from optimized to given @see camera_optimizers.py
        pose_deltas = exp_map_SO3xR3(pose_adjustment)

        for i, frame in enumerate(train_opt_frames):
            # input T from camera to world
            C2W = ...

            Copt2C = np.eye(4)
            Copt2C[:3, :] = pose_deltas[i].detach().cpu().numpy()
            # optimized T from camera to world
            Copt2W = Copt2C @ C2W
```
Hope it helps

sam598:
@vijendra1125 unfortunately your fix no longer works because of this pull

https://github.com/nerfstudio-project/nerfstudio/pull/2092

Any suggestions on how to obtain the optimizer from an already trained checkpoint?

Issue #931, opened 2022-11-10
Selecting camera in viewer isn't persistent


tancik:
**Describe the bug**
Clicking the "eye" in the camera list moves the green camera, but does not change the time step. Therefore when interacting with the scene it snaps back

https://user-images.githubusercontent.com/3310961/201157826-259340f4-baaf-4aff-9153-e7831f9d6447.mov

THE-COB:
**Disclaimer: I'm new here and this is my first issue so what I'm saying might be totally wrong/dumb**

I think this is because when the mouse moves to the viewer the green box goes to the camera that would be selected based on the playback slider. I figured the sensible thing to do would be just to change the slider position when you click the 👁️ button. The problem is I don't think you can access the right slider state element from that element.

https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/viewer/app/src/modules/SidePanel/CameraPanel/CameraPanel.jsx#L392
This is where we'd need to change the slider.

https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/viewer/app/src/modules/SidePanel/CameraPanel/CameraPanel.jsx#L458
This is where the correct slider state is defined.

There seems to be another slider state defined above the 👁️ button but when I set it to something nothing seems to happen. I'll keep messing with it though.

THE-COB:
I just looked at it again and I realized I could just pass in the slider stuff in props.

here is my PR: https://github.com/nerfstudio-project/nerfstudio/pull/1192

I think I need a maintainer to approve running workflows so I'd appreciate if someone looked at it at their earliest convenience! Obviously no rush especially cause it's New Year's! Happy Holidays everyone! @tancik (I think)

THE-COB:
Can I close this issue @tancik ?

PR #1192, opened 2022-12-31
Making camera selection on viewer persistent

THE-COB:
I basically just passed in `slider_value` and `set_slider_value` as a prop to the `CameraList` so it can change the slider.

Fix for https://github.com/nerfstudio-project/nerfstudio/issues/931


tancik:
Thanks for the fix!

PR #1196, opened 2023-01-01
Add hashmap size and num_levels to instant_ngp_field, issue 1158

DreekFire:
Added the new arguments to the end of the argument list because this would preserve compatibility with pre-existing code written using positional arguments. However, this argument order differs from that used in nerfacto.


tancik (approved):
LGTM

PR #1197, opened 2023-01-01
PIL Dependency

mcallisterdavid:
Up-to-date PIL package is required for ns-render but not in dependencies list


pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
     "nerfacc==0.3.2",
     "open3d>=0.16.0",
     "opencv-python==4.6.0.66",
+    "Pillow==9.3.0",
```
tancik, line 36: Does >=9.3.0 work?

mcallisterdavid: Sorry forgot about this, changed it to >= !!

tancik (approved):
LGTM

PR #1198, opened 2023-01-01
fixed an exception bug in case of num_samples is 0

dozeri83:
fixed an exception bug when num samples is 0 , 
"IndexError tensors used as indices must be long, byte or bool tensors"

change type to long ( was float32)


tancik (approved):
LGTM

Issue #1178, opened 2022-12-27
Add ability to render a video with dynamic changes in time. Also ability to save time as part of camera paths.


nlml:
**Is your feature request related to a problem? Please describe.**
Right now, rendering a dynamic nerf trained with the _dnerf_ method config completely ignores the temporal distortion part of this model. This is because in `nerfstudio/cameras/camera_paths.py` a `times` variable is not created and passed to the `Cameras` constructed at the end of that file.

Also there is no way to specify the times you would like to be rendered when creating a cameras path in the viewer. There should be a way to save times with each camera as part of the camera path.

**Describe the solution you'd like**
- When adding cameras in the viewer to a rendering path, there should be an option to specify a `time` value between 0 and 1 for each camera. Probably this should be a global option in the render tab (enabled only when the Render Timestep control on the Controls tab is enabled), as either your model is dynamic or it isn't. These times should then be interpolated with the same smoothing as per the cameras.
- The interpolated per-frame time values should then be exported to the `"camera_path"` list, in each camera, under a new field called "time"
- If these "time" fields are present, they should be read by `nerfstudio/cameras/camera_paths.py` and sent to the `Cameras` construction at the end of that file.

**Describe alternatives you've considered**
Had manually implemented this myself and it was working fine.

**Additional context**
Mentioned on nerfstudio discord #troubleshooting

nlml:
I am currently working on a PR for this. Should be done today.

nlml:
Fixed in #1199

PR #1199, opened 2023-01-02
Render cameras with render times for dynamic nerfs

nlml:
See https://github.com/nerfstudio-project/nerfstudio/issues/1178

Add a `--dynamic` flag to `ns-render`. If this is provided, the `get_path_from_json` will look for the field `render_time` in each camera of the `camera_path` (which comes from UI-exported .json file). If these fields are present, then when constructing the `Cameras`, these `render_time`s are used. This allows for varying timestamps when rendering from a dynamic nerf.

On the viewer side, like with the "Render Timestep" control on the Controls tab, if the `/model/has_temporal_distortion` msg is true, then controls for the Render cameras' timesteps will be visible. Using these additional controls, one can now set the `render_time`s that are exported to the camera path .json. This is acheived basically by copying the setup for FOV. Like FOV, the `render_time` can either be the same for all cameras, or specified per-camera. As per FOV, smoothness is applied to the per-camera `render_time`s.

Have tested this PR on a private dataset and it works as expected. If needed I could put together a toy dataset as a test.

Potential (minor) issues or nitpicks with this PR:
- Depending on the smoothness and the input values, the exported `render_time`s can be <0.0 or >1.0. This is probably not a huge deal and the exported values shouldn't go too far beyond this range, but perhaps we want to clip the values.
- Just a visual nitpick: the Render Time displayed in the viewer UI will show '1' or '0' instead of '1.0' and '0.0' if the entered value is exactly 1 or 0.
- I believe also now all exported `camera_path` `.json`s will contain this `render_time` field, even if the path was exported from a non-dynamic nerf. This shouldn't be a problem however as they will be ignored by `ns-render` if the `--dynamic` flag is not passed.

### DIFF
--- a/nerfstudio/viewer/app/src/modules/SidePanel/CameraPanel/CameraPanel.jsx
+++ b/nerfstudio/viewer/app/src/modules/SidePanel/CameraPanel/CameraPanel.jsx
@@ -63,6 +63,104 @@ function set_camera_position(camera, matrix) {
   mat.decompose(camera.position, camera.quaternion, camera.scale);
 }
 
+
+function RenderTimeSelector(props) {
+  const disabled = props.disabled;
+  const isGlobal = props.isGlobal;
+  const camera = props.camera;
+  const dispatch = props.dispatch;
+  const globalRenderTime = props.globalRenderTime;
+  const setGlobalRenderTime = props.setGlobalRenderTime;
+  const applyAll = props.applyAll;
+  const changeMain = props.changeMain;
+  const setAllCameraRenderTime = props.setAllCameraRenderTime;
+
+  const getRenderTimeLabel = () => {
+    if (!isGlobal) {
+      return camera.renderTime;
+    }
+    camera.renderTime = globalRenderTime
+    return globalRenderTime;
+  };
+
+  const [UIRenderTime, setUIRenderTime] = React.useState(
+    isGlobal ? globalRenderTime : getRenderTimeLabel(),
+  );
+
+  const [valid, setValid] = React.useState(true);
+
+  useEffect(
+    () => setUIRenderTime(getRenderTimeLabel()),
+    [camera, globalRenderTime],
+  );
+
+  const setRndrTime = (val) => {
+    if (!isGlobal) {
+      camera.renderTime = val;
+    } else {
+      camera.renderTime = val;
+      setGlobalRenderTime(val);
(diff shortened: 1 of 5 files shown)


nerfstudio/viewer/app/src/modules/SidePanel/CameraPanel/CameraPanel.jsx
```diff
@@ -392,6 +492,8 @@ function CameraList(props) {
                 e.stopPropagation();
                 set_camera_position(camera_main, camera.matrix);
                 camera_main.fov = camera.fov;
+                camera_main.renderTime = camera.renderTime;
+                camera_main.renderTime = camera.renderTime;
```
tancik, line 496: duplicated line

tancik (commented):
Nice!
Can you add the `--dynamic` to the render command when it is used - https://github.com/nerfstudio-project/nerfstudio/blob/3da877050df2c9b347652a5a84ad6d5d8c768650/nerfstudio/viewer/app/src/modules/RenderModal/RenderModal.jsx#L30
Also I think we should clamp the time to 0-1, as some models may require that.

nlml:
Added the `--dynamic` flag to the render cmd and render times are now clamped to [0,1] before writing to the json

tancik:
Thinking a little more about it, does it make sense not to include the time in the json if time isn't used. Then the export script can render with time only if it is present? That would remove the requirement for the user to specify `--dynamic`

nlml:
> Thinking a little more about it, does it make sense not to include the time in the json if time isn't used. Then the export script can render with time only if it is present? That would remove the requirement for the user to specify `--dynamic`

Makes sense, actually I initially thought this too but since it was my first time with React, I was just going with what seemed simplest at the time. But I've changed it now to work as you described and it actually makes the overall PR and changes much simpler. So now we only pass `times` to `Cameras` if **all** items in the camera_paths contain the `render_time` key. And these keys are only exported if `/model/has_temporal_distortion` msg is `true`.

I re-tested with both a dnerf and nerfacto (non-dynamic) model and the exported .json is either with/without the `render_time` keys, as expected, so I think we are good.

tancik (approved):
LGTM, thanks for making those changes. Sorry for not mentioning it earlier.

nlml:
No worries! Thanks!

PR #1205, opened 2023-01-05
CropBox GUI implementation

dozeri83:
This is a first pull request regarding a crop box implementation.
I chose to follow a very similar interface to the instant-ngp-method.
The main algorithmic part was in the "generate_rays" method where the nears and fars are calculate.

In order for this to work the model have to "respect" this fields in the rendering.
I only tested this for instant-ngp-bound and nerfacto and it seems to work well.

In the second part - I am planning to integrate this with the ns-render and ns-export but I don't want get too far from the main repo so I want to merge this since it is quite complete feature.

@tancik , @ethanweber

image


tancik:
Can you increment this value, this insures that people are using the latest code with the viewer - https://github.com/nerfstudio-project/nerfstudio/blob/99ab8bd669a590efcbce0ff27727b16e0403d71d/nerfstudio/viewer/app/package.json#L4

nerfstudio/cameras/cameras.py
```diff
@@ -22,13 +22,16 @@
 from typing import Dict, List, Optional, Tuple, Union
 
 import cv2
+import nerfacc
```
tancik, line 25: Im a little cautious about importing this here. Currently nerfacc is only used by instant-ngp which means that those cuda kernels were only getting build if people were using that model. Adding it here would force everyone to build those kernels (and some have had issues with that in the past). Not really sure what the best option is...

tancik:
I played around with in the viewer, it is really cool! I ran across a couple of bugs. The crop box got stuck at 0 and wouldn't let me increase the value. Playing with the scale non-intuitively changed the min max values. I think we may want to consider a different ui, the scaling, min, max is a bit confusing. What are your thought regarding something similar to what we do for defining the bounding box in the export panels? @ethanweber and I are currently traveling so it is difficult for use to contribute at the moment, but we should be able to help in a week or so. 
One other issue to consider is how the "backgound color" is handled. in nerfacto, the background color is set to the last sample's, this causes a "shell" to appear around the crop region. If cropping is used, the render should instead use black for background color.

dozeri83:
> Can you increment this value, this insures that people are using the latest code with the viewer -
> 
> https://github.com/nerfstudio-project/nerfstudio/blob/99ab8bd669a590efcbce0ff27727b16e0403d71d/nerfstudio/viewer/app/package.json#L4

change it to 22-12-17-0  ?

dozeri83:
> 

regarding the scale, my main intension was to fix the scale to the scale of the scene_aabb in "init_viewer" function, and remove it completely from the GUI
When I did this I wanted to automatically change the "x_min", "y_min" (etc) values to their new values but I was not able to do this. Maybe we should wait for your help we you arrive (enjoy the trip BTW).

dozeri83:
Regarding the "backgound color" - I also stumble upon that. 
Does the current last_sample backgound color is important for the training?
 
If so, how would you suggest changing the bg color only for the rendering? 
Should I add bg_color to "get_outputs" method ?

nerfstudio/cameras/cameras.py
dozeri83: Correct me if I am wrong but doesn't nerfacc builds the kernel only on the actual use of one of its method?
If not, we can import it dynamically in the function using importlib

https://www.geeksforgeeks.org/importlib-package-in-python/

dozeri83: just for the sport of it, I implemented the interction with aabb box in pure pytorch. 
I added tests and check that it produces the same outputs as nerfacc for random set of rays.
It is order of magnitude slower and the nerfacc (For 500 rays ~ 0.4 milisec . nerfacc ~ 0.05 milisec)  , 
but it is still pretty usable in the gui.

I define env variable "INTERSECT_WITH_NERFACC" with default of 1 -
if INTERSECT_WITH_NERFACC is set to 1, we use nerfacc with dynamic import.
if INTERSECT_WITH_NERFACC is set to 0, we can use the pytorch alternative.

users that will have problem with INTERSECT_WITH_NERFACC can be instructed to set it to 0.

ethanweber:
> > Can you increment this value, this insures that people are using the latest code with the viewer -
> > https://github.com/nerfstudio-project/nerfstudio/blob/99ab8bd669a590efcbce0ff27727b16e0403d71d/nerfstudio/viewer/app/package.json#L4
> 
> change it to 22-12-17-0 ?

You can change it to `23-01-10-0`. It's year, month, date, number.

tancik (approved):

smart4654154:
Thank you very much for your work. Can you provide some suggestions for my question? thanks #2894

smart4654154:
> > 
> 
> regarding the scale, my main intension was to fix the scale to the scale of the scene_aabb in "init_viewer" function, and remove it completely from the GUI When I did this I wanted to automatically change the "x_min", "y_min" (etc) values to their new values but I was not able to do this. Maybe we should wait for your help when you arrive (enjoy the trip BTW).

Thank you very much for your work. My question seems to be related to this job. What is the meaning of the crop scale value in the new version of viewer? I found that the value of the crop scale cannot match the coordinate values of the world coordinate system. For example, when the crop scale is set to 4, 4, 4, it is approximately equivalent to a square with a side length of 40 in the world coordinate system
image
What are the meanings of "crash max" and "crash min" in Viewer_legacy?
image

PR #1216, opened 2023-01-09
save image_coords as non-persistent buffer

hturki:
Right now image coords are being needlessly serialized as part of the checkpoint state dict


tancik (approved):
Good catch, thanks!

PR #1231, opened 2023-01-14
Fix typo COLAMP->COLMAP in colmap_utils.py

matsuren:
Fix typo COLAMP->COLMAP for console output in colmap_utils.py


matsuren:
It seems that there are more typos. I only modified documentations and comments so that it should be no runtime errors caused by this PR.

tancik (approved):
LGTM, thanks for going through and correcting all of these typos!

PR #1250, opened 2023-01-18
setting tensorf default from wandb to viewer

### LINKED ISSUE #1194: Include "--vis viewer" in sample "ns-train" command on the README
**Describe the bug**

On the [README](https://github.com/nerfstudio-project/nerfstudio), it implies the `ns-train` command will launch the web viewer by default. However when I tried it today it just launched WanB. I believe this is due to the `--vis` option being set to `wandb` by default, not `viewer` (although in the future, it'd be nice to be able to use both simultaneously.

Anyway, I got confused and I figured it might be worth updating the README to include the `--vis viewer` arg, so that the user can see the viewer while training runs. 

**To Reproduce**
Steps to reproduce the behavior: just following the first-time setup instructions on the README:
```bash
conda create --name nerfstudio -y python=3.8
conda activate nerfstudio
python -m pip install --upgrade pip
conda create --name nerfstudio -y python=3.8
conda activate nerfstudio
python -m pip install --upgrade pip
pip install nerfstudio
# I have a custom dataset, and I trained using TensoRF
ns-train tensorf --data data/engine_6_ds/
```

**Expected behavior**
When I go to https://viewer.nerf.studio/?websocket_url=ws://localhost:7007 after running `ns-train nerfacto --data data/nerfstudio/poster`, it should show the viewer is connected. 

**Screenshots**

Screen Shot 2022-12-31 at 7 51 34 PM

**Additional context**
- I just tried `nerfstudio` for the first time today. So I'm not if this could just be an issue with the latest version (I installed using `pip`).
- this was tested on Google Chrome

Thanks for creating this project! It is so helpful to have all these tools in one place.

terrancewang:
Fixes #1194

### DIFF
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -248,6 +248,9 @@ method_configs["vanilla-nerf"] = TrainerConfig(
 
 method_configs["tensorf"] = TrainerConfig(
     method_name="tensorf",
+    steps_per_eval_batch=500,
+    steps_per_save=2000,
+    max_num_iterations=30000,
     mixed_precision=False,
     pipeline=VanillaPipelineConfig(
         datamanager=VanillaDataManagerConfig(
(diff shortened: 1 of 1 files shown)


tancik (approved):
LGTM

Issue #451, opened 2022-09-16
save camera_path.json files with timestamp and in data directory


ethanweber:
Let's keep track of camera paths for our datasets. We can save them in the data folder instead of with the training run.

THE-COB:
When selecting the render button on the UI, should it default to the most recent camera_path in the data directory? Should there be another input option for that?

PR #1251, opened 2023-01-18
Camera path

THE-COB:
PR for [https://github.com/nerfstudio-project/nerfstudio/issues/451](https://github.com/nerfstudio-project/nerfstudio/issues/451)

I basically just sent the path to the data directory with a websocket the same way that the config directory is passed.

### DIFF
--- a/nerfstudio/viewer/app/src/modules/RenderModal/RenderModal.jsx
+++ b/nerfstudio/viewer/app/src/modules/RenderModal/RenderModal.jsx
@@ -28,7 +28,6 @@ export default function RenderModal(props: RenderModalProps) {
   );
 
   // react state
-  // const [filename, setFilename] = React.useState('render_output');
 
   const handleClose = () => setOpen(false);
 
--- a/nerfstudio/viewer/app/src/modules/RenderModal/RenderModal.jsx
+++ b/nerfstudio/viewer/app/src/modules/RenderModal/RenderModal.jsx
@@ -19,6 +19,17 @@ export default function RenderModal(props: RenderModalProps) {
     (state) => state.renderingState.config_base_dir,
   );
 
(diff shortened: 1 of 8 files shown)


nerfstudio/viewer/app/src/modules/RenderModal/RenderModal.jsx
```diff
@@ -19,14 +19,25 @@ export default function RenderModal(props: RenderModalProps) {
     (state) => state.renderingState.config_base_dir,
   );
 
+  const timestamp_regex = /[0-9]{4}-[0-9]{2}-[0-9]{2}_[0-9]{6}/g;
```
THE-COB, line 22: Is it alright if I get the timestamp of the directory like this? I figured it was simpler than sending the timestamp in a websocket but that might be cleaner?
I'm happy to change this
@ethanweber

nerfstudio/viewer/app/src/modules/SidePanel/CameraPanel/CameraPropPanel.jsx
```diff
@@ -43,6 +54,20 @@ export default function CameraPropPanel(props) {
 
   const [, setControls] = useControls(
     () => ({
+      path: {
+        label: 'Export Name',
+        value: export_path,
+        onChange: (v) => {
+          const valid_filename_reg = /^([a-z]|[A-Z]|[0-9]|-|_|\(|\))+$/g;
+          if(!valid_filename_reg.test(v)){
+            alert("Please only use letters, numbers, parentheses, hyphens, and underscores in path name (no spaces either)");
```
tancik, line 63: nit: "Please only use letters, numbers, parentheses, hyphens, and underscores in path name (no spaces either)" -> "Please only use letters, numbers, and hyphens)"

tancik (approved):
Minor comment, otherwise looks good!

PR #1253, opened 2023-01-18
Depth data loading for Polycam LiDAR data

ManuConcepBrito:
I was working on loading the LiDAR data from the polycam app by adding an option to the `ns-process-data polycam` interface. The API is as follows:

`ns-process-data polycam --use-depth --data 8ene,6-17p. m.-poly/ --output-dir output_data`

It searches in the keyframes folder for the depth data, scales it to match the RGB dimensions and adds it to the transforms.json. Is this still relevant with the recent [depth supervision pull request](https://github.com/nerfstudio-project/nerfstudio/pull/1173)?

### DIFF
--- a/nerfstudio/process_data/polycam_utils.py
+++ b/nerfstudio/process_data/polycam_utils.py
@@ -34,7 +34,7 @@ def polycam_to_json(
     cameras_dir: Path,
     output_dir: Path,
     min_blur_score: float = 0.0,
-    crop_border_pixels: int = 0
+    crop_border_pixels: int = 0,
 ) -> List[str]:
     """Convert Polycam data into a nerfstudio dataset.
 
@@ -98,8 +98,14 @@ def polycam_to_json(
     return summary
 
 
-def process_images(polycam_image_dir: Path, image_dir: Path, crop_border_pixels: int = 15, max_dataset_size: int = 600,
-                   num_downscales: int = 3, verbose: bool = True) -> Tuple[List[str], List[Path]]:
+def process_images(
+    polycam_image_dir: Path,
+    image_dir: Path,
+    crop_border_pixels: int = 15,
+    max_dataset_size: int = 600,
+    num_downscales: int = 3,
+    verbose: bool = True,
+) -> Tuple[List[str], List[Path]]:
     """
     Process RGB images only
 
@@ -153,9 +159,15 @@ def process_images(polycam_image_dir: Path, image_dir: Path, crop_border_pixels:
     return summary_log, polycam_image_filenames
 
 
-def process_depth_maps(polycam_depth_dir: Path, depth_dir: Path,
-                       num_processed_images: int, crop_border_pixels: int = 15, max_dataset_size: int = 600,
-                        num_downscales: int = 3, verbose: bool = True) -> Tuple[List[str], List[Path]]:
+def process_depth_maps(
+    polycam_depth_dir: Path,
+    depth_dir: Path,
+    num_processed_images: int,
+    crop_border_pixels: int = 15,
+    max_dataset_size: int = 600,
+    num_downscales: int = 3,
+    verbose: bool = True,
+) -> Tuple[List[str], List[Path]]:
     """
     Process Depth maps from polycam only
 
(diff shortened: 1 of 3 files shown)


scripts/process_data.py
```diff
@@ -597,21 +645,57 @@ def main(self) -> None:
...
-                cameras_dir=polycam_cameras_dir,
-                output_dir=self.output_dir,
-                min_blur_score=self.min_blur_score,
-                crop_border_pixels=self.crop_border_pixels,
-            )
+
+        return summary_log, polycam_image_filenames
+
+
+    def process_depth_maps(self, polycam_depth_dir: Path, depth_dir: Path, num_processed_images: int) -> List[str]:
```
tancik, line 652: Can you add a docstring

tancik, line 652: This can be moved to `process_data/polycam_utils.py`

scripts/process_data.py
```diff
@@ -550,13 +551,59 @@ def main(self) -> None:
...
+        )
+
+
+        CONSOLE.rule("[bold green]:tada: :tada: :tada: All DONE :tada: :tada: :tada:")
+
+        for summary in summary_log:
+            CONSOLE.print(summary, justify="center")
+        CONSOLE.rule()
+
+    def process_images(self, polycam_image_dir: Path, image_dir: Path) -> Tuple[List[str], List[Path]]:
```
tancik, line 594: This can be moved to `process_data/polycam_utils.py`

tancik (commented):
Giving the option to load polycam depth maps is a good idea. Thanks for working on it.

ManuConcepBrito:
I think all the comments were resolved except passing the core tests (this was developed with python 3.7 and building the docs requires >=3.8 due to the ipython version), I'll check that later.

ManuConcepBrito:
This is passing tests locally so it should be ready.

nerfstudio/process_data/polycam_utils.py
```diff
@@ -46,6 +50,7 @@ def polycam_to_json(
     Returns:
         Summary of the conversion.
     """
+    use_depth = bool(len(image_filenames) == len(depth_filenames))
```
tancik, line 53: nit: I don't think the `bool` casting necessary?

nerfstudio/process_data/process_data_utils.py
```diff
@@ -156,6 +157,54 @@ def copy_images_list(
...
+        copied_depth_map_paths.append(copied_depth_map_path)
+
+    # upscale them in place
+    with status(msg="[bold yellow] Upscaling depth maps...", spinner="growVertical", verbose=verbose):
+        upscale_factor = 2 ** POLYCAM_UPSCALING_TIMES
+        assert upscale_factor > 1
+        assert isinstance(upscale_factor, int)
+        depth_dir.mkdir(parents=True, exist_ok=True)
+        file_type = depth_dir.glob("frame_*").__next__().suffix
+        filename = f"frame_%05d{file_type}"
```
tancik, line 190: It seems like this `%05d` is unreliable. #1259 has a workaround.

tancik (commented):
Added a couple of comments.

nerfstudio/process_data/process_data_utils.py
machenmusik: Note that in the unreliable case, ffmpeg outputs a single JPEG; #1259 does have a fix, but I am double checking whether there is a ffmpeg flag that also solves it, similar to #1261, as that would be a little more efficient.

ManuConcepBrito: This should also be fixed in the crop border pixels from downscale_images then too?

machenmusik: > I am double checking whether there is a ffmpeg flag that also solves it, similar to https://github.com/nerfstudio-project/nerfstudio/pull/1261, as that would be a little more efficient.

No luck, #1259 still seems to be necessary.

> This should also be fixed in the crop border pixels from downscale_images then too?

I am not sure, I did not try using it with the problematic dataset (CottageMiniature) that required #1259

machenmusik: I think I discovered the %05d issue, CottageMiniature includes mask images, and the RealityCapture import is treating them as non-mask images... so the image format changes from 24 bit to 1 bit, ffmpeg stops processing after first frame.  I will make a new issue for that.

ManuConcepBrito: So from my understanding, there's no need to get rid of the %05d, right?

machenmusik: My opinion is that you want to take what is in main (via #1259) because there is no degradation and it makes all the datasets work.

machenmusik: (If you already have a list of the files to be processed, you can avoid the os.scandir to gather them that #1259 introduces for non-depth.)

tancik:
Great! I think once the merge conflicts are resolved, this should be ready to go.

ManuConcepBrito:
> Great! I think once the merge conflicts are resolved, this should be ready to go.

Done!

tancik (approved):
LGTM. Thanks for adding!

tancik:
`ns-dev-test` should fix the remaining issue.

ManuConcepBrito:
Sorry for that! Just pushed the black formatting changes

PR #1259, opened 2023-01-19
downsample all images

machenmusik:
CottageMiniature sample dataset was only downsampling first image correctly prior to this fix.

It would appear that trusting ffmpeg with `%05d` commands to process all images is unreliable.  There may be other areas where similar fixes are warranted.


machenmusik:
The speed of downsampling itself is identical from what I can detect; it is still doing same ffmpeg scaling.
In the problem case, because it is downsampling all the images rather than one, it may take longer overall of course, but should yield correct result :-)

tancik (approved):
LGTM

PR #1264, opened 2023-01-20
feat(ci): switch caching to `actions/setup-python@v4`

SauravMaheshkar:
Changes made by this PR can be summarized as follows:

- Set the `cache` and `cache-dependency-path` argument of `actions/setup-python` to enable `pip` caching.
- Delete the redundant `actions/cache@v2` step.

### DIFF
--- a/.github/workflows/core_code_checks.yml
+++ b/.github/workflows/core_code_checks.yml
@@ -19,10 +19,8 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.7.13'
-      - uses: actions/cache@v2
-        with:
-          path: ${{ env.pythonLocation }}
-          key: ${{ env.pythonLocation }}-${{ hashFiles('pyproject.toml') }}
+          cache: 'pip'
+          cache-dependency-path: pyproject.toml
       - name: Install dependencies
         run: |
           pip install --upgrade --upgrade-strategy eager -e .[dev]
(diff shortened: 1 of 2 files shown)


tancik:
Im a bit unfamiliar with how this caching works under the hood. I've rerun it a few times (so I think it is using the cached version) and it seems to average ~4min. Other runs of `Core Tests` seems to be ~1-1.5 min quicker. Any ideas why this would be the case?

image

image

SauravMaheshkar:
@tancik this is interesting. I don't know why this is the case. I had a look at the older versions of "Core Tests" and it seems like (setup python + caching + dependency installation) does happen quicker compared to (setup python with caching + dependency installation).

SauravMaheshkar:
Maybe the slower installation times could be because of "fetching" the packages from a cache.

tancik:
Is there an advantage to using this cache if the other method is faster? I ideally don't want to pay for more github action minutes.

SauravMaheshkar:
None that I am aware of, it just seemed best practice to cache the dependencies. 

> The action has built-in functionality for caching and restoring dependencies. It uses [toolkit/cache](https://github.com/actions/toolkit/tree/main/packages/cache) under the hood for caching dependencies but requires less configuration settings. ([_From the "Caching package dependencies" section of setup-python README_](https://github.com/actions/setup-python#caching-packages-dependencies))

If it indeed is slower than just using `actions/cache` which is what the cache implementation is based on then adding the cache parameter might not be ideal.

SauravMaheshkar:
@tancik How about I close this PR at the moment ? and maybe in the future we can get back to it if there are visible benefits.

tancik:
yea, sounds good. Sorry for not having more to add.

PR #1268, opened 2023-01-20
Bug fix for loading cameras with different distortions in nerfstudio_dataparser

matsuren:
Hi, 

I found the bug in `nerfstudio_dataparser.py` when I was working on multiple cameras with different distortion parameters.

This PR should fix `nerfstudio_dataparser` to load distortion parameters from multiple cameras correctly.

Thanks,


tancik (approved):
LGTM, wow bad bug, thanks for finding

PR #1270, opened 2023-01-20
DyCheck dataset parser

lsongx:
Dataparser for DyCheck (https://arxiv.org/abs/2210.13445) Dataset (`iphone` subset).


tancik (approved):
LGTM

Issue #1271, opened 2023-01-20
Error using `ns-render`: "TypeError: `load_pipeline()` missing 1 required positional argument: 'step'"


Mason-McGough:
I recently updated `nerfstudio` to the latest version and attempted to render the poster scene again using the following commands:

```bash
ns-download-data nerfstudio --capture-name=poster
ns-train nerfacto --data data/nerfstudio/poster
``` 

I then created a few camera poses in the UI and then clicked the "Rendering" button to generate a command. I copied the command and pasted it in a new terminal.

```bash
ns-render --load-config outputs/data-nerfstudio-poster/nerfacto/2023-01-20_183016/config.yml --traj filename --camera-path-filename data/nerfstudio/poster/camera_paths/2023-01-20_183016.json --output-path renders/data/nerfstudio/poster/2023-01-20_183016.mp4
```

This time it produced the following output:

```bash
[18:31:36] Auto image downscale factor of 2                                                 nerfstudio_dataparser.py:314
           Skipping 0 files in dataset split train.                                         nerfstudio_dataparser.py:165
           Skipping 0 files in dataset split test.                                          nerfstudio_dataparser.py:165
Loading latest checkpoint from load_dir
Traceback (most recent call last):
  File "/home/bricklayer/anaconda3/envs/nerfstudio/bin/ns-render", line 8, in <module>
    sys.exit(entrypoint())
  File "/home/bricklayer/Workspace/contrib/nerfstudio-fork/nerfstudio/scripts/render.py", line 271, in entrypoint
    tyro.cli(RenderTrajectory).main()
  File "/home/bricklayer/Workspace/contrib/nerfstudio-fork/nerfstudio/scripts/render.py", line 224, in main
    _, pipeline, _ = eval_setup(
  File "/home/bricklayer/Workspace/contrib/nerfstudio-fork/nerfstudio/nerfstudio/utils/eval_utils.py", line 105, in eval_setup
    checkpoint_path = eval_load_checkpoint(config, pipeline)
  File "/home/bricklayer/Workspace/contrib/nerfstudio-fork/nerfstudio/nerfstudio/utils/eval_utils.py", line 62, in eval_load_checkpoint
    pipeline.load_pipeline(loaded_state["pipeline"])
TypeError: load_pipeline() missing 1 required positional argument: 'step'
```

[...]

PR #1272, opened 2023-01-20
Add `step` argument in `eval_utils.py` to fix rendering bug.

Mason-McGough:
Proposed fix for [issue 1271](https://github.com/nerfstudio-project/nerfstudio/issues/1271). I was able to use `ns-render` again after this fix.


tancik (approved):
LGTM

PR #1273, opened 2023-01-21
Update Dockerfile

Scolymus:
You didn't download the nerfstudio git! Therefore it was failing running the installation RUN cd nerfstudio && python3.10 -m pip install -e . &&  cd ..

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,6 +96,8 @@ RUN python3.10 -m pip install --upgrade pip setuptools pathtools promise
 RUN python3.10 -m pip install torch==1.12.1+cu116 torchvision==0.13.1+cu116 torchaudio==0.12.1 --extra-index-url https://download.pytorch.org/whl/cu116
 # Install tynyCUDNN.
 RUN python3.10 -m pip install git+https://github.com/NVlabs/tiny-cuda-nn.git#subdirectory=bindings/torch
+# Copy nerfstudio git
+RUN git clone https://github.com/nerfstudio-project/nerfstudio.git
 
 # Copy nerfstudio folder and give ownership to user.
 ADD . /home/user/nerfstudio


tancik:
I am a novice when it comes to docker. @Zunhammer do you mind commenting on this PR?

Scolymus:
I also noticed I have errors when running ns-train. It says `OSError: Unknown compute capability. Ensure PyTorch with CUDA support is installed.`

I guess is due to the torch versions installed. We are using a FROM of a cuda 11.7 environment, but the torch versions looks like cuda 11.6. That could be the error. If cuda 11.7 is needed, I'll try to find the correct torch versions.

Zunhammer:
Hi, so the idea of this Dockerfile is to use it with the code being mounted from your local disk inside the container. by this you can edit and update the code without needing to rebuild the docker image.

If you're looking to have a ready to use solution I suggest to simply use the docker image from docker hub: https://hub.docker.com/r/dromni/nerfstudio

So, regarding the issue: as far as I can see on the first look, the mount point (line 101: ADD . /home/user/nerfstudio) needs to be moved before the install commands. I'll check this later (after work) and give feedback then.

Scolymus:
@Zunhammer oh I see! Otherwise I wanted to try `FROM nvidia/cuda:11.6.0-devel-ubuntu20.04` and `python 3.9` (since python3.10-dev does not exist in ubuntu20.04).

So now I'm using `FROM dromni/nerfstudio:0.1.15`. Shall you put a readme on using docker? I know users using docker should know, but maybe they are newbies as me :)

PS. If you make it, don't forget to tell them that you are changing the user and they may need to put `USER root:root
` after `FROM dromni/nerfstudio:0.1.15` or at least change to their user if they want to solve an error like "E: List directory /var/lib/apt/lists/partial is missing. - Acquire (13: Permission denied)"

Zunhammer:
Hi @Scolymus, in my last message, I wrote it slightly wrong (sorry, was quite busy today :) ), the idea is that after you've built the image you can either:
- run the docker image and you have the nerfstudio version inside which was on your disk during docker build
OR
- you run the docker image and mount the current folder into the container so that the "nerfstudio" folder inside the docker gets overwritten by the one on your local machine. By doing this you can apply changes easily, even during runtime of the container.

Anyway, I just rebuilt the image from the current repo (main branch) as well as release v0.1.15 and don't see any issue, did you maybe forget to add the "--gpus all" when running the container? Without this, the container cannot access the GPU and gives the same error you mentioned above (OSError: Unknown compute capability.....). Same could happen if your nvidia driver is not capable of working with CUDA 11.7 (simply update to the most recent version). Please verify and let me know if the issue still persists.

Regardless of your problem, I don't see any need for this PR. As written above, the nerfstudio folder is copied into the docker image, so nothing missing. Feel free to update the description for the docker if it's not good enough, though :)

Thanks and best regards

Scolymus:
Hi @Zunhammer, yeah, you are right. There's no need. It was my fault. About starting the docker using `FROM dromni/nerfstudio:0.1.15`, am I doing something wrong? I opened this issue https://github.com/nerfstudio-project/nerfstudio/issues/1292

Regarding the documentation, I saw now the documentation on the website. I was just looking the Readme in the github project, and the comment came from that, but it's already fine in the documentation.

PR #1274, opened 2023-01-21
intergate crop box with ns-render

dozeri83:
Hi @tancik ,
I've integrate the crop box with with the rendering, pretty simple implementation.

as for the export, I will leave this to your team, since the crop box and the clip box are colliding with each other and I dont know what to do.


tancik:
Took a slightly different approach in #1295 by saving the crop settings in the camera path json file instead of making it a command line argument. Going to close this PR for now.

PR #1277, opened 2023-01-23
nerfplayer with ngp and nerfacto

lsongx:
NeRFPlayer implementation with `instant-ngp-bounded` and `nerfacto` model

Examples:
With `nerfacto` as the base model
nerfacto

With `instant-ngp-bounded` as the base model
instantngp

### DIFF
--- a/nerfstudio/fields/nerfplayer_nerfacto_field.py
+++ b/nerfstudio/fields/nerfplayer_nerfacto_field.py
@@ -1,0 +1,409 @@
+# Copyright 2022 The Nerfstudio Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Field implementations for NeRFPlayer (https://arxiv.org/abs/2210.15947) implementation with nerfacto backbone
+"""
+
+
+from typing import Optional
+
+import numpy as np
+import torch
+from torch.nn.parameter import Parameter
+from torchtyping import TensorType
+
+from nerfstudio.cameras.rays import Frustums, RaySamples
+from nerfstudio.data.scene_box import SceneBox
+from nerfstudio.field_components.activations import trunc_exp
+from nerfstudio.field_components.embedding import Embedding
+from nerfstudio.field_components.field_heads import (
+    FieldHeadNames,
+    PredNormalsFieldHead,
+    SemanticFieldHead,
+    TransientDensityFieldHead,
+    TransientRGBFieldHead,
+    UncertaintyFieldHead,
+)
+from nerfstudio.field_components.spatial_distortions import SpatialDistortion
+from nerfstudio.field_components.temporal_grid import TemporalGridEncoder
+from nerfstudio.fields.base_field import Field
+
+try:
+    import tinycudann as tcnn
+except ImportError:
+    # tinycudann module doesn't exist
+    pass
+
+
+def get_normalized_directions(directions: TensorType["bs":..., 3]):
+    """SH encoding must be in the range [0, 1]
+
+    Args:
+        directions: batch of directions
+    """
+    return (directions + 1.0) / 2.0
+
+
+class TemporalHashMLPDensityField(Field):
+    """A lightweight temporal density field module.
+
+    Args:
+        aabb: parameters of scene aabb bounds
+        num_layers: number of hidden layers
+        hidden_dim: dimension of hidden layers
+        spatial_distortion: spatial distortion module
+    """
+
+    def __init__(
+        self,
+        aabb,
+        temporal_dim: int = 64,
+        num_layers: int = 2,
+        hidden_dim: int = 64,
+        spatial_distortion: Optional[SpatialDistortion] = None,
+        num_levels=8,
+        max_res=1024,
+        base_res=16,
+        log2_hashmap_size=18,
+        features_per_level=2,
+    ) -> None:
+        super().__init__()
+        # from .temporal_grid import test; test() # DEBUG
+        self.aabb = Parameter(aabb, requires_grad=False)
+        self.spatial_distortion = spatial_distortion
+        growth_factor = np.exp((np.log(max_res) - np.log(base_res)) / (num_levels - 1))
+
+        self.encoding = TemporalGridEncoder(
+            input_dim=3,
+            temporal_dim=temporal_dim,
+            num_levels=num_levels,
+            level_dim=features_per_level,
+            per_level_scale=growth_factor,
+            base_resolution=base_res,
+            log2_hashmap_size=log2_hashmap_size,
+        )
+        self.linear = tcnn.Network(
+            n_input_dims=num_levels * features_per_level,
+            n_output_dims=1,
+            network_config={
+                "otype": "FullyFusedMLP",
+                "activation": "ReLU",
+                "output_activation": "None",
+                "n_neurons": hidden_dim,
+                "n_hidden_layers": num_layers - 1,
+            },
+        )
+
+    # pylint: disable=arguments-differ
+    def density_fn(self, positions: TensorType["bs":..., 3], times: TensorType["bs", 1]) -> TensorType["bs":..., 1]:
+        """Returns only the density. Used primarily with the density grid.
+
+        Args:
+            positions: the origin of the samples/frustums
+            times: the time of rays
+        """
+        if len(positions.shape) == 3 and len(times.shape) == 2:
+            # position is [ray, sample, 3]; times is [ray, 1]
+            times = times[:, None]  # RaySamples can handle the shape
+        # Need to figure out a better way to descibe positions with a ray.
+        ray_samples = RaySamples(
+            frustums=Frustums(
+                origins=positions,
+                directions=torch.ones_like(positions),
+                starts=torch.zeros_like(positions[..., :1]),
+                ends=torch.zeros_like(positions[..., :1]),
+                pixel_area=torch.ones_like(positions[..., :1]),
+            ),
+            times=times,
+        )
+        density, _ = self.get_density(ray_samples)
+        return density
+
+    def get_density(self, ray_samples: RaySamples):
+        if self.spatial_distortion is not None:
+            positions = self.spatial_distortion(ray_samples.frustums.get_positions())
+            positions = (positions + 2.0) / 4.0
+        else:
+            positions = SceneBox.get_normalized_positions(ray_samples.frustums.get_positions(), self.aabb)
+        positions_flat = positions.view(-1, 3)
+        time_flat = ray_samples.times.reshape(-1, 1)
+        x = self.encoding(positions_flat, time_flat).to(positions)
+        density_before_activation = self.linear(x).view(*ray_samples.frustums.shape, -1)
+
+        # Rectifying the density with an exponential is much more stable than a ReLU or
+        # softplus, because it enables high post-activation (float32) density outputs
+        # from smaller internal (float16) parameters.
+        density = trunc_exp(density_before_activation)
+        return density, None
+
+    def get_outputs(self, ray_samples: RaySamples, density_embedding: Optional[TensorType] = None):
+        return {}
+
+
+class NerfplayerNerfactoField(Field):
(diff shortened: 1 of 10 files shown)


machenmusik:
Nice!  I've been able to reproduce results, but nerfacc wants to set up CUDA again every time... which takes quite a while.  Not sure whether that is related to changes here, or separate issue.  Do others see the same behavior?

nerfstudio/fields/nerfplayer_nerfacto_field.py
```diff
@@ -0,0 +1,409 @@
...
+
+
+class TemporalHashMLPDensityField(Field):
+    """A lightweight temporal density field module.
+
+    Args:
+        aabb: parameters of scene aabb bounds
+        num_layers: number of hidden layers
+        hidden_dim: dimension of hidden layers
+        spatial_distortion: spatial distortion module
```
tancik, line 66: Add missing arg docs

tancik (commented):
Looks great!

machenmusik:
> Thank you @machenmusik ! It needs to compile `nerfstudio\field_components\cuda` for the first run, which can take up to 20mins. (`MAX_JOBS` env for ninja may help speed up this). But after compiling, I do not need to compile later. Would you mind providing a little more details about 'setting up CUDA again every time'?

For example, I trained the mochi-high-five dataset (`ns-train nerfplayer-nerfacto ...`) to completion; before it started training, I got the usual `Setting up CUDA (This may take a few minutes the first time)` from nerfacc.

But now that I am loading that previously trained checkpoint (`ns-train nerfplayer-nerfacto --load-dir ...`), I am seeing the CUDA setup message and delay again, which I would not expect.

I do notice that I see it from NerfAcc, and then a similar message again
`nerfstudio field components: Setting up CUDA (This may take a few minutes the first time)` - 
I don't remember seeing `nerfstudio field components` previously, but not sure if that is now expected.

lsongx:
> For example, I trained the mochi-high-five dataset (`ns-train nerfplayer-nerfacto ...`) to completion on Windows 10 machine; before it started training, I got the usual `Setting up CUDA (This may take a few minutes the first time)` from nerfacc.
> 
> But now that I am loading that previously trained checkpoint (`ns-train nerfplayer-nerfacto --load-dir ...`), I am seeing the CUDA setup message and delay again, which I would not expect.
> 
> I do notice that I see it from NerfAcc, and then a similar message again `nerfstudio field components: Setting up CUDA (This may take a few minutes the first time)` - I don't remember seeing `nerfstudio field components` previously, but not sure if that is now expected.

Yes, this compilation is expected. Some more CUDA kernels are needed to run nerfplayer -- implementation can be found at `nerfstudio/field_components/cuda`. If this message appears only once on a new machine, I think that is fine.

machenmusik:
> Yes, this compilation is expected. Some more CUDA kernels are needed to run nerfplayer -- implementation can be found at `nerfstudio/field_components/cuda`. If this message appears only once on a new machine, I think that is fine.

Sure, but the issue is that it was required the second time through `ns-train nerfplayer-nerfacto` as well (which is still processing now).

lsongx:
> Sure, but the issue is that it was required the second time through `ns-train nerfplayer-nerfacto` as well (which is still processing now).

I see. Thanks for helping debug this! I think it is related to the `.so` file, which may not exist on Windows. https://github.com/nerfstudio-project/nerfstudio/blob/828d6bc893683042b3763247b65b3adeb5d5ea7e/nerfstudio/field_components/cuda/_backend.py#L30
I updated a new version. Would you mind giving it a try? Thanks again @machenmusik

machenmusik:
> > Sure, but the issue is that it was required the second time through `ns-train nerfplayer-nerfacto` as well (which is still processing now).
> 
> I see. Thanks for helping debug this! I think it is related to the `.so` file, which may not exist on Windows.
> 
> https://github.com/nerfstudio-project/nerfstudio/blob/828d6bc893683042b3763247b65b3adeb5d5ea7e/nerfstudio/field_components/cuda/_backend.py#L30
> 
> 
> Trying to solve this now...

Thanks.  I am trying the latest, still not done with first CUDA compile yet.

machenmusik:
This worked:
```
_C = None
if cuda_toolkit_available():
    if os.listdir(BUILD_DIR) != []:
        # If the build exists, we assume the extension has been built
        # and we can load it.
        print("nerfstudio field components: CUDA set up, loading (should be quick)")
        _C = load(
            name=NAME,
            sources=glob.glob(os.path.join(PATH, "csrc/*.cu")),
            extra_cflags=["-O3", "-std=c++14"],
            extra_cuda_cflags=["-O3", "-std=c++14"],
            extra_include_paths=[],
        )
    else:
        # Build from scratch. Remove the build directory just to be safe: pytorch jit might stuck
        # if the build directory exists.
        shutil.rmtree(BUILD_DIR)
        print("nerfstudio field components: Setting up CUDA (This may take a few minutes the first time)")
        _C = load(
            name=NAME,
            sources=glob.glob(os.path.join(PATH, "csrc/*.cu")),
            extra_cflags=["-O3", "-std=c++14"],
            extra_cuda_cflags=["-O3", "-std=c++14"],
            extra_include_paths=[],
        )
        print("nerfstudio field components: Setting up CUDA finished")
else:
    print("nerfstudio field components: No CUDA toolkit found. Some models may fail.")
```

Also https://github.com/KAIR-BAIR/nerfacc/pull/150

machenmusik:
Pushed the working change -- note that https://github.com/KAIR-BAIR/nerfacc/pull/150 is also needed or else nerfacc 0.3.x has same problem.  Does this still work on Linux?

lsongx:
> Pushed the working change -- note that [KAIR-BAIR/nerfacc#150](https://github.com/KAIR-BAIR/nerfacc/pull/150) is also needed or else nerfacc 0.3.x has same problem. Does this still work on Linux?

Yes. This works. Thanks!
If a user cancels the compilation in the middle, the folder is not empty. But loading a non-empty and unfinished build may have some issues (from this comment)
https://github.com/KAIR-BAIR/nerfacc/blob/58d6b4c46dfee36ab4b565ab0655b1a27ff4859f/nerfacc/cuda/_backend.py#L78

machenmusik:
> Yes. This works. Thanks! If a user cancels the compilation in the middle, the folder is not empty. But loading a non-empty and unfinished build may have some issues (from this comment) https://github.com/KAIR-BAIR/nerfacc/blob/58d6b4c46dfee36ab4b565ab0655b1a27ff4859f/nerfacc/cuda/_backend.py#L78

Library suffixes are generally .so, .dll, or .a - right?  Maybe checking for those would work for both Linux and Windows (and perhaps Mac as well).

lsongx:
> Library suffixes are generally .so, .dll, or .a - right? Maybe checking for those would work for both Linux and Windows (and perhaps Mac as well).

Sounds good. But I cannot find an example of that. Do you know some examples so that we can learn from them?
Or we can simply ignore the suffix, like
```
if NAME in [p.stem for p in Path(BUILD_DIR).glob("*")]:
```

I see that most of the libs do not use the jit load. They move the build to the setup step (i.e., `pip install` step), and load it as a module.  Not sure if this will be good practice for nerfstudio and nerfacc.

machenmusik:
> Sounds good. But I cannot find an example of that. Do you know some examples so that we can learn from them? 

I was thinking simply to do logical OR of exists calls; they should be lightning quick.

> I see that most of the libs do not use the jit load. They move the build to the setup step (i.e., `pip install` step), and load it as a module. Not sure if this will be good practice for nerfstudio and nerfacc.

Right, the install would then take a rather long time even if the user never utilized those functions.

lsongx:
> I was thinking simply to do logical OR of exists calls; they should be lightning quick.

Agree. I think we may wait and see if there are enough people from different platforms willing to help test it.
Besides, NerfAcc @liruilong940607  may also provide a better solution. We can come back to this later.
Do you think currently the directory check solution can get merged? After searching for a while, the stuck problem brought by non-empty build dir looks not common.

machenmusik:
Note: this did NOT work (and would have rebuilt for many minutes if I hadn't commented out the rmtree)
```
if os.path.exists(os.path.join(BUILD_DIR, f"{NAME}.so")) or os.path.exists(os.path.join(BUILD_DIR, f"{NAME}.dll")) or os.path.exists(os.path.join(BUILD_DIR, f"{NAME}.a")):
```
I would merge with the directory check.

machenmusik (approved):
Thanks for contributing and refining!  @tancik any last requests?

tancik:
> I see that most of the libs do not use the jit load. They move the build to the setup step (i.e., pip install step), and load it as a module. Not sure if this will be good practice for nerfstudio and nerfacc.

I think this is really what we should be doing. Since this is the first cuda code in this repo, maybe we should do this now.

lsongx:
For the sometimes unwanted time-consuming build (they do not use it),  what about providing an extra step for `pip install`? 

The default behavior of `pip install` remains the same (no compilation). Then if someone wants to use it, we first throw an error message, telling the user to run another `pip install [compile setup]`.

tancik:
yea, let's start with that for now, it could be `pip install .[cuda]`

machenmusik:
The only concern I have is that the CUDA compile times are egregiously long, and people should be warned that the extra install will be lengthy.  It may also be worth trying to mention that second install step after doing the initial one for those that want to get it out of the way, and allowing a way to ask for all the compiles to be triggered up front.

If nerfacc doesn't have the same behavior (so still stalls) or requires a separate additional install step, that could get tedious.

tancik:
> The default behavior of pip install remains the same (no compilation). Then if someone wants to use it, we first throw an error message, telling the user to run another pip install [compile setup].

I think this solution of yours solves the issue of it taking a while, since people will only run it if necessary

`Nerfacc` is trickier, we can figure out what to do with that later.

machenmusik:
> > The default behavior of pip install remains the same (no compilation). Then if someone wants to use it, we first throw an error message, telling the user to run another pip install [compile setup].
> 
> I think this solution of yours solves the issue of it taking a while, since people will only run it if necessary
> 
> `Nerfacc` is trickier, we can figure out what to do with that later.

Trickier perhaps, but if we want to avoid JIT stalls for more deterministic behavior, there seems little point in doing so for field components but still stalling on nerfacc's CUDA JIT.  That is actually worse IMO since you have to do two separate CUDA compiles, one of which is unexpected.

tancik:
> Trickier perhaps, but if we want to avoid JIT stalls for more deterministic behavior, there seems little point in doing so for field components but still stalling on nerfacc's CUDA JIT. That is actually worse IMO since you have to do two separate CUDA compiles, one of which is unexpected.

I agree. I've got to chat with @liruilong940607 to get his input first.

machenmusik:
Great.  For pytorch cuda, IIRC we download (pre-built) wheels, perhaps we can do similar for nerfacc and nerfstudio...

tancik:
> Great. For pytorch cuda, IIRC we download (pre-built) wheels, perhaps we can do similar for nerfacc and nerfstudio...

That would be cool, it surpasses my knowledge on python deployment. How easy is that to do in practice? Also does pypi support multiple wheels?

machenmusik:
I am not terribly familiar with the build mechanics either - but it does seem like the right approach.  Volunteers welcome...

machenmusik:
https://discuss.python.org/t/what-to-do-about-gpus-and-the-built-distributions-that-support-them/7125

machenmusik:
From some light research, making wheels available for various permutations seems a challenge;
at the moment I would suggest restoring the behavior we had with nerfacc 0.2.x (with the appropriate change to nerfacc)
https://github.com/nerfstudio-project/nerfstudio/pull/1277#issuecomment-1404215735
until there is a better solution that does not force lengthy CUDA recompilation every time on Windows.

machenmusik:
(It seems unfortunate that we're holding this up on the CUDA JIT question, when the nerfplayer functionality seems to work well.)

tancik:
> (It seems unfortunate that we're holding this up on the CUDA JIT question, when the nerfplayer functionality seems to work well.)

True, is the cuda in a good state now (ie the original issue you brought up @machenmusik is resolved). If so I'm fine with merging and addressing larger cuda changes later.

machenmusik:
@tancik yes the Windows CUDA recompilation is fixed by combination of https://github.com/nerfstudio-project/nerfstudio/pull/1277/commits/59cb8d2ca49fa500bee331337da7e2dfcd26d49a in this PR, and also https://github.com/KAIR-BAIR/nerfacc/pull/150 which is merged but may need a release and dependency bump.

tancik:
Ok, @lsongx feel free to merge when ready.

lsongx:
Thanks, merged! Feel free to mention me if there are any bug reports or feature requests or ... related to these PRs. Will be happy to contribute more, including future CUDA changes.

tancik:
Sounds good, thanks for adding this @lsongx!

machenmusik:
Nice work @lsongx !
@tancik do we need to bump nerfacc version separately?

tancik:
@machenmusik the nerfacc version was just bumped. There are also not nerfacc wheels https://github.com/KAIR-BAIR/nerfacc#installation

PR #1282, opened 2023-01-24
Increase max_res of first proposal network

kerrj:
Tweaked first proposal network hashgrid resolution from 64 to 128, helps the disappearing effect on thin objects.


tancik (approved):
LGTM

PR #1284, opened 2023-01-24
Fix total train time metric logging

williamshen-nz:
The total train time metric was not being logged as `writer.write_out_storage()` was not being called.

It would have included the time the viewer was open until the user killed the process which probably isn't what we want as well.

Metric now logged in wandb
image

Thanks for the great project and support on discord by the way!


tancik (approved):
LGTM, thanks for fixing!

PR #1288, opened 2023-01-26
Updated Blender add-on calculating duration

cvachha:
Updated how the Blender add on calculates duration with frame steps and added the case when frame step size is 0.


tancik (approved):
LGTM

PR #1290, opened 2023-01-26
Load path dropdown

THE-COB:
New Load Path button now opens a modal where users can either load an existing path or upload a path.

image


tancik (approved):
LGTM

PR #1294, opened 2023-01-26
Remove bracket from ffmpeg command's crop filter

heymarcell:
Using bracket's in this situation causing syntax errors:
/bin/sh: 1: Syntax error: "(" unexpected


tancik (approved):
LGTM

PR #1300, opened 2023-01-29
Corrected the order of texture arguments

MaroonAmor:
@tancik Before this fix, using `scripts/texture.py` to texture an existing mesh will give an error: `TypeError: unsupported operand type(s) for /`.

This is due to the inconsistent argument order when calling `texture_utils.export_textured_mesh` from `scripts/texture.py`. The arguments `self.px_per_uv_triangle` and `self.output_dir` should be swapped.


tancik (approved):
LGTM, thanks for fixing

PR #1303, opened 2023-01-29
Tancik/protobuf version req update

tancik:
Current setting breaks PEP 508


Zunhammer:
For me it is still violating PEP508, a fix should remove the dot + asterix (.*) behind 3.20.3 like:
```
"protobuf<=3.20.3,!=3.20.0"
```

PR #1319, opened 2023-01-31
adding check if camera_path_dir exists

terrancewang:
currently errors and ends training if datapath/camera_path doesn't exist. should still be able to load a path without erroring with this check.


THE-COB (approved):
Good catch, sorry about that

tancik (approved):
LGTM

Issue #873, opened 2022-11-01
Nan accumulation for nerfacto applied on redwoods2


monniert:
**Describe the bug**

`AssertionError: the min value is -9223372036854775808` in `outputs["accumulation"]` after 490 iterations using nerfacto on nerfstudio/redwoods2. I tried different seeds and still get the error at 490 iterations. This does not happen on the other nerfstudio scenes.

**To Reproduce**

1. Follow the installation guidelines in the readme.md (with CUDA 11.3, Ubuntu 18.04)
2. Download the scene using `ns-download-data --dataset nerfstudio --capture redwoods2`
3. Run `ns-train nerfacto --vis tensorboard --data data/nerfstudio/redwoods2`.

**Expected behavior**

It should train smoothly

**Screenshots**

![Screenshot 2022-11-01 at 7 12 19 PM](https://user-images.githubusercontent.com/25098794/199313878-a4e2f620-1d1e-482c-9546-1d7cd1bd2a4b.png)

**Extra context**

It seems related to #117, which looks resolved by a previous PR. Let me know if you need more details on the machine/package versions

tancik:
I'm able to replicate this error. We would also see this issue when we trained nerfacto for a large number of iterations (>50k). Seems like it appears early for this scene. The issue is likely due to precision / numerical stability, however we have yet to trace it down.

nikmo33:
So Ive been debugging this issue on my side - and have been able to reproduce / remove the following error through the changes here - Though im not really sure why scripting would cause an issue here
https://github.com/nerfstudio-project/nerfstudio/pull/910

snarb:
@tancik I have the same issue with my scene , when training nerfacto model when training for large number of iterations.

snarb:
@tancik Looks like the reason is not in a precision. When training with mixed_precision = False I got Nan much faster and worse quality during training for some reason.

snarb:
@tancik I am using the repository main branch, but not recent. Latest commit Date: 11/2/2022 10:17:45 PM 121771d126a1c90cf39013d8016e16effbfe8efc. 

I have an update. My scenario: I have trained for 1h+ on A100. Then got this error. I have tried 2 times to resume training from the latest checkpoint but got the same error in 10 mins. Then I enabled anomaly detection using
[...]

PR #1331, opened 2023-02-01
Add support for gradient clipping

hturki:
I've found that this makes training more stable in certain cases, especially when using high learning rates

### DIFF
--- a/nerfstudio/engine/optimizers.py
+++ b/nerfstudio/engine/optimizers.py
@@ -18,7 +18,7 @@ Optimizers class.
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any, Dict, List, Type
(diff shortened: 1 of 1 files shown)


tancik (approved):
LGTM, are you also planning to update the nerfacto config?

hturki:
i wasn't planning on it, have been testing on fairly specific usecases so far with a hash-based (but not nerfacto) architecture, and I'm not sure whether this is needed for nerfacto / the specific clipping values I'm using will generalize. For nerfacto I'd maybe start with adding weight decay per https://github.com/nerfstudio-project/nerfstudio/issues/873 and then we could figure out reasonable clipping values if training is still unstable.

tancik:
Gotcha 👍

PR #1334, opened 2023-02-01
add option to log training gradients

hturki:
This might also help with https://github.com/nerfstudio-project/nerfstudio/issues/873


nerfstudio/engine/trainer.py
```diff
@@ -188,7 +190,6 @@ def train(self) -> None:
             step = 0
             for step in range(self._start_step, self._start_step + num_iterations):
                 with TimeWriter(writer, EventName.ITER_TRAIN_TIME, step=step) as train_t:
-
```
hturki, line 191: this is unrelated to my commit, but the build was failing due to formatting and black gave me this.

tancik (approved):
LGTM

PR #1336, opened 2023-02-02
Add lambertian renderer

terrancewang:


nerfstudio/model_components/renderers.py
```diff
@@ -332,3 +332,32 @@ def forward(
...
+
+class LambertianShadingRenderer(nn.Module):
+    """Calculate Lambertian shading."""
+
+    @classmethod
+    def forward(
+        cls,
+        rgb: TensorType["bs":..., "num_samples", 3],
+        normals: TensorType["bs":..., "num_samples", 3],
+        light_direction: TensorType["bs", 3],
```
tancik, line 345: should this be `["bs"..., 3]`

nerfstudio/model_components/renderers.py
```diff
@@ -332,3 +332,32 @@ def forward(
...
+        Args:
+            rgb: Accumulated rgb along a ray.
+            normals: Accumulated normals along a ray.
+            light_direction: Direction of light source.
+            ratio: ambient lighting (1.0) vs. Lambertian shading (0.0) ratio
+
+        Returns:
+            Textureless Lambertian shading, Lambertian shading
+        """
+        lambertian = ratio + (1 - ratio) * (normals.detach() @ light_direction).clamp(min=0)
```
tancik, line 359: Do we always want to detach normals? Maybe we should add an argument that allows the user to choose.

nerfstudio/model_components/renderers.py
```diff
@@ -332,3 +332,32 @@ def forward(
...
+class LambertianShadingRenderer(nn.Module):
+    """Calculate Lambertian shading."""
+
+    @classmethod
+    def forward(
+        cls,
+        rgb: TensorType["bs":..., "num_samples", 3],
+        normals: TensorType["bs":..., "num_samples", 3],
+        light_direction: TensorType["bs", 3],
+        ratio: float = 1.0,
```
tancik, line 346: Seems odd for the default to be non-lambertian. Does something like .5 make more sense?

tancik (commented):
Thanks for adding.

nerfstudio/model_components/renderers.py
tancik: Actually thinking a bit more, I think `ratio` could use a different name like `shading_weight` where when `shading_weight = 1` the render gives you something that is pure lambertian (0 gives no shading). The default can then be 1.0.

terrancewang: yeah that name is more clear

tancik (approved):
LGTM

nerfstudio/model_components/renderers.py
mcallisterdavid: I didn’t see this earlier, but the reason for detaching the normals was that they get supervised by your SDS loss in Dreamfusion if you don’t detach them. The optimization then makes all your normals the same direction which basically turns off lambertian shading. Maybe it’s best to just turn on “detached normals” or something for the Dreamfusion implementation

terrancewang: do you remember if you guys ever just tried using the normals network that stable dreamfusion used? I think they were using the sds loss to supervise normals.

PR #1338, opened 2023-02-02
add neus surface rendering method with mono priors

pablovela5620:
Adding in NeuS surface rendering implementation from [sdfstudio](https://github.com/autonomousvision/sdfstudio). Seems to be working with their provided Replica dataset.
image
I plan on adding the nerfstudio -> sdfstudio parser script as well. This method is too slow to use with the renderer so I'll also be planning on brining in the NeuS-Facto method. I pretty much followed what they implemented very closely, lmk if things make sense or if you'd like me to change implementation details. 

I also need to add in the mesh export to extract .ply file from model

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -92,6 +92,25 @@
       "program": "scripts/train.py",
       "console": "integratedTerminal",
       "args": ["semantic_nerf"]
-    }
+    },
+    {
+      "name": "Python: Neus on Replica",
+      "type": "python",
+      "request": "launch",
+      "program": "scripts/train.py",
+      "console": "integratedTerminal",
+      "justMyCode": false,
+      "env": {"CUDA_VISIBLE_DEVICES":"0"},
+      "args": [
+        "neus",
+        "--pipeline.model.sdf-field.inside-outside=True",
+        "--pipeline.model.mono-depth-loss-mult=0.1",
+        "--pipeline.model.mono-normal-loss-mult=0.05",
+        "--viewer.websocket-port=7007",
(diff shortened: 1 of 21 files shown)


tancik:
I sent you some comments on discord regarding the integration.

gowriaddepalli:
@pablovela5620 @tancik can someone please link discord link here for reference. Thanks in advance.

pablovela5620:
> @pablovela5620 @tancik can someone please link discord link here for reference. Thanks in advance.

it was through a direct message so there's no link that can be shared, id be fine posting it here if that's okay with Tancik as well

tancik:
It was basically about breaking the PR into multiple smaller PRs so that it is easier to review.

pablovela5620:
- [x]  Dataset/Parser - this will include the depth/normal predictions and download from sdfstudio demo dataset. I will need to eventually add the depth/normal generation via omnidata or through the provided Polycam depth (I think the former is better to start since it means it can still work with COLMAP)
- [ ]  Losses - Main additions are the monosdf normal/depth (these don't totally make sense without data above) for indoor scene reconstruction / sparse views
- [ ]  Samplers + Models + Fields. I couldn't really find a good way to break this up since the sampler/fields/models are all based on the SDF (which no other current models provide), Unfortunately I think this will probably be a fairly large PR. I'll remove to start everything that doesn't have to do with Neus (since sdfstudio included all other methods that make more complex like VolSDF and Unisurf)
- [ ]  Marching Cubes/ Mesh Extraction. To extract mesh from SDF and texture

PR #1341, opened 2023-02-03
Fix normals visualization

tancik:
Gradient Normals:
Before:
image

After:
image

Predicted Normals:
Before:
image

After:
Pasted Graphic 8


yaozhuyu-nreal:
Hi, I find that  normalizing normals to range [0, 1] leads to wrong results (not everywhere) when doing poisson reconstruction. Maybe poisson reconstruction (o3d.geometry.TriangleMesh.create_from_point_cloud_poisson) needs to make use of point cloud normals in range [-1, 1].

tancik:
@yaozhuyu-nreal good catch, this needs to be fixed.

PR #1343, opened 2023-02-03
New config parameter --viewer.codec defaults to VP8, but allows H264

### LINKED ISSUE #1325: odd width and/or height prevent viewer from using H264 instead of VP8
https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/viewer/server/viewer_utils.py#L694
is hardcoded to select VP8 as video encoder.
I wanted to explore using H264 which is often hardware accelerated in decoders and encoders.

However, in doing so I noticed: 
- the selected encoder was x264 not NVENC et al. (which will likely be slower)
- errors from width and/or height having odd values (I wonder whether this could explain memory leaks)

machenmusik:
Fixes #1325.

### DIFF
--- a/nerfstudio/configs/base_config.py
+++ b/nerfstudio/configs/base_config.py
@@ -21,7 +21,7 @@ from __future__ import annotations
 import warnings
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Literal, List, Optional, Tuple, Type
+from typing import Any, List, Literal, Optional, Tuple, Type
 
 # model instances
 from nerfstudio.utils import writer
--- a/nerfstudio/configs/base_config.py
+++ b/nerfstudio/configs/base_config.py
@@ -149,3 +149,5 @@ class ViewerConfig(PrintableConfig):
     skip_openrelay: bool = False
     """Avoid using openrelay to communicate with the viewer. Try disabling if you have trouble
     connecting to the viewer"""
+    codec: str = "H264"
+    """Video codec that viewer will use; H264 or VP8."""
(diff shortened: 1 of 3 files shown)


tancik:
Im fine with making H264 the default, but Im curious, are there any specific reasons to prefer it over VP8?

Also with H264 I see alot of spurious printing
image

machenmusik:
H264 has better compression (perhaps better for remote viewers) -- and may facilitate hardware encoding which should be somewhat faster / less burdensome for the server.

Not sure where the log spam from libx264 is from, probably somewhere in aiortc etc.?

nerfstudio/configs/base_config.py
```diff
@@ -149,3 +149,5 @@ class ViewerConfig(PrintableConfig):
     skip_openrelay: bool = False
     """Avoid using openrelay to communicate with the viewer. Try disabling if you have trouble
     connecting to the viewer"""
+    codec: str = "H264"
```
tancik, line 152: Can you make this `codec: Literal["H264", "VP8"]`, Then the options can be removed from the docstring

tancik:
I've dug around and haven't found any obvious ways to disable the logging. Lets keep the default to VP8 for now and then consider changing if we figure it out.

machenmusik:
Done.
(What I've seen about PyAV logging doesn't seem to work on the libx264 spam...)

machenmusik:
Sidebar: I notice that when doing dynamic image size calculation (e.g. changing viewer position/orientation), there are quite a number of different sizes, which seem to cause a good number of encoder changes.  I wonder it would be better to have fewer discrete sub-frame sizes, to thrash a bit less, similar to downsampling... but that's probably more important if hardware encoding.

tancik (approved):
LGTM

PR #1347, opened 2023-02-03
double-quote filenames in process-data commands so that on Windows, paths with spaces don't fail

### LINKED ISSUE #1346: On Windows, ns-process-data fails for paths with spaces
Using `ns-process-data` with `images` or `video` that has spaces in the path fails.

machenmusik:
Fixes #1346 .
Confirmed on Windows 10 for `ns-process-data` `images` and `video` with spaces in filename.

Unable to test all processing options though.


machenmusik:
Could use help testing that it doesn't break non-Windows, and also the more esoteric options for process-data.

(Not sure if related to the big filed against training on Windows)

tancik (approved):
Works on linux!

PR #1357, opened 2023-02-06
Allow methods configs to be discovered from external packages.

jkulhanek:
This PR implements a discovery mechanism for TrainerConfigs. The main objective is to allow other researchers to publish their own methods which use Nerfstudio without them needing to copy the source code. They will publish only the method implementation with TrainerConfig, and it will be automatically discovered and registered by Nerfstudio.

The implementation uses the package metadata to find the TrainerConfigs. In particular, it will look for the "nerfstudio.method_configs" entry point.

I believe this will improve the collaboration as contributors no longer need to clone the entire repository if they only want to try a small change. Also, this way, Nerfstudio's authors can wait if an implementation works well before porting it into the Nerfstudio codebase.

### DIFF
--- a/docs/developer_guides/config.md
+++ b/docs/developer_guides/config.md
@@ -118,3 +118,27 @@ Often times, you just want to play with the parameters of an existing model with
   # NOTE: the dataparser and associated configurations go at the end of the command
   ns-train {METHOD_NAME} --vis viewer {DATA_PARSER} --scale-factor 0.5
   ```
+  
+### Extending Nerfstudio with custom methods
+In order to extend the Nerfstudio and register your own methods, you can package your code as a python package
+and register it with Nerfstudio as a `nerfstudio.method_configs` entrypoint in the `pyproject.toml` file.
+The Nerfstudio will automatically look for all registered methods and will register them to be used
+by method such as `ns-train`.
+
+Here is an example:
+```python
+"""my_package/my_config.py"""
+
+MyConfig = TrainerConfig(
+    method_name="my-method",
+    ...
+)
+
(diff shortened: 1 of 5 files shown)


tancik (commented):
Cool idea. How does this handle versioning? Ie does the external model define a nerfstudio version compatibility? Im just curious what happens if we make breaking changes to the nerfstudio repo (as it is still in pretty early development).

jkulhanek:
Thanks! The idea is for the external model to define the Nerfstudio version as python dependency. Therefore, breaking changes should not break compatibility as external packages will likely target a specific version of Nerfstudio.

Errors should be fixed. Not sure about the naming of the new class "MethodSpecification" which contains the TrainerConfig and the method description. What do you think?

tancik (approved):
Name is good with me. Thanks for adding!

machenmusik:
Nice!  Were you thinking that there is a built-in method that would be a good test case to show this decomposition in action, or was the idea to enable the functionality and the first method to use it would become the de facto example?

jkulhanek:
I thought of it in the context of adding new methods. I believe the Plenoxels could be a good candidate. Adding it directly to the Nerfstudio repo would make Nerfstudio harder to install while most people don’t need it.

vincent-lecrubier-skydio:
I landed here after searching "Plenoxels" in this repo. If anyone works on integrating plenoxels with Nerfstudio I'd be interested!

jkulhanek:
I have it on my todo list. Unfortunately I didn’t have enough time lately. I will prepare something after the ICCV deadline (if nobody does it sooner).

PR #1358, opened 2023-02-06
add viewer.local config property when viewing local instance

machenmusik:
If running local server instance, avoid using relays to communicate with the viewer.

### DIFF
--- a/nerfstudio/configs/base_config.py
+++ b/nerfstudio/configs/base_config.py
@@ -153,3 +153,5 @@ class ViewerConfig(PrintableConfig):
     connecting to the viewer"""
     codec: Literal["H264", "VP8"] = "VP8"
     """Video codec that viewer will use."""
+    local: bool = False
+    """If running local server instance, avoid using relays to communicate with the viewer."""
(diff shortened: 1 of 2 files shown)


tancik (commented):
I think this in practice does the exact same thing as `--viewer.skip-openrelay`
When `RTCPeerConnection` doesn't receive an input, I believe it falls back to [this default](https://github.com/aiortc/aiortc/blob/d30b6f75d4457204b1ab336106a248d535be053a/src/aiortc/rtcicetransport.py#L205-L209) which is what we use.

machenmusik:
It's not quite the same actually, for example if you try in Airplane Mode, you can't connect to the Google server.

tancik:
Hmm, I guess I don't understand their code well enough. I thought it uses `stun:stun.l.google.com:19302` when no argument is provided.
However if you can confirm that `--viewer.local` works in airplane mode and `--viewer.skip-openrelay` doesn't, that's enough for me ;)

machenmusik:
> --viewer.local works in airplane mode 
Yes
> --viewer.skip-openrelay doesn't,
So it's interesting.  This does work for me, presumably because it fails to find/connect to the Google server and then gives up.

But if the point is to avoid even trying to touch the Google server, then you want the behavior of local and not skip-openrelay.

PR #1359, opened 2023-02-06
Handle uncalibrated Metashape sensor

gilureta:
This fix allows loading Metashape XML files which
contain an additional uncalibrated sensor.

Metashape XML format keeps unmatched images but
assignes a dummy sensor to them. The sensor has
no calibration data.


tancik (approved):
LGTM, thanks for fixing this. I imagine there are many metashape edge cases like this and Im glad to see them getting patched.

PR #1371, opened 2023-02-07
Improve ns-process-data images when using existing COLMAP model

jkulhanek:
- Additional flag --skip-images for ns-process-data, which skips copying and downscaling of images. The typical use case is if the user already downscaled the images and wants to run just COLMAP
- Additional parameter --colmap-model-path can be used with --skip-colmap. It allows specifying the path to the COLMAP model as opposed to always using "colmap/sparse/0"
- Fixes a bug with --skip-colmap. When --skip-colmap is used with an existing COLMAP model, the filenames are not renamed (but they are renamed when they are copied and downscaled). This makes the resulting transforms.json contain invalid paths.

NOTE: I accidentally closed PR #1366. This PR contains the same changes.

### DIFF
--- a/scripts/process_data.py
+++ b/scripts/process_data.py
@@ -74,6 +74,10 @@ class ProcessImages:
         will downscale the images by 2x, 4x, and 8x."""
     skip_colmap: bool = False
     """If True, skips COLMAP and generates transforms.json if possible."""
+    skip_images: bool = False
+    """If True, skips copying and downscaling of images and only runs COLMAP if possible and enabled"""
+    colmap_model_path: str = "colmap/sparse/0"
+    """Optionally sets the path of the colmap model. Used only when --skip-colmap is set to True."""
     colmap_cmd: str = "colmap"
     """How to call the COLMAP executable."""
     gpu: bool = True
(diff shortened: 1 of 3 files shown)


scripts/process_data.py
```diff
@@ -74,6 +74,10 @@ class ProcessImages:
         will downscale the images by 2x, 4x, and 8x."""
     skip_colmap: bool = False
     """If True, skips COLMAP and generates transforms.json if possible."""
+    skip_images: bool = False
```
tancik, line 77: Make more descriptive, -> `skip_image_processing`

scripts/process_data.py
```diff
@@ -133,11 +150,16 @@ def main(self) -> None:
                 sys.exit(1)
 
         # Save transforms.json
-        if (colmap_dir / "sparse" / "0" / "cameras.bin").exists():
+        if self.skip_colmap:
+            colmap_model_path = Path(self.colmap_model_path)
+        else:
+            colmap_model_path = colmap_dir / "sparse" / "0"
+
+        if (colmap_model_path / "cameras.bin").exists():
```
tancik, line 158: We should print an error if the file doesn't exist.

tancik (commented):
Thanks for cleaning up this pipeline. We have received multiple requests for supporting externally colmaped data.

scripts/process_data.py
jkulhanek: Ok

scripts/process_data.py
jkulhanek: I added error printing. Is this what you meant?

tancik (approved):
LGTM

PR #1379, opened 2023-02-08
Add ns-viewer command to run viewer only with model in eval mode

jkulhanek:
This PR adds a command to start the viewer with the model in eval mode.
It is intended as an easier alternative to "ns-train --viewer.start-train=False"


scripts/viewer/run_viewer.py
```diff
@@ -0,0 +1,118 @@
...
+        viewer_log_path = base_dir / config.viewer.relative_log_filename
+        viewer_state, banner_messages = None, None
+        viewer_state, banner_messages = viewer_utils.setup_viewer(
+            config.viewer, log_filename=viewer_log_path, datapath=config.pipeline.datamanager.dataparser.data
+        )
+
+        # setup logging
+        writer.setup_local_writer(config.logging, max_iter=config.max_num_iterations, banner_messages=banner_messages)
+        writer.put_config(name="config", config_dict=dataclasses.asdict(config), step=0)
+        profiler.setup_profiler(config.logging)
```
tancik, line 73: Probably don't need a profiler

scripts/viewer/run_viewer.py
```diff
@@ -0,0 +1,118 @@
...
+        base_dir = config.get_base_dir()
+        viewer_log_path = base_dir / config.viewer.relative_log_filename
+        viewer_state, banner_messages = None, None
+        viewer_state, banner_messages = viewer_utils.setup_viewer(
+            config.viewer, log_filename=viewer_log_path, datapath=config.pipeline.datamanager.dataparser.data
+        )
+
+        # setup logging
+        writer.setup_local_writer(config.logging, max_iter=config.max_num_iterations, banner_messages=banner_messages)
+        writer.put_config(name="config", config_dict=dataclasses.asdict(config), step=0)
```
tancik, line 72: Does it run without this?

tancik (commented):
Thanks for adding. We should separate "Resume from checkpoint / visualize existing run" in the readme and "train your first model" docs to loading a checkpoint, and visualizing trained model.

machenmusik:
(`ns-viewer` -- or should it be `ns-view` ?)

scripts/viewer/run_viewer.py
jkulhanek: Sure

scripts/viewer/run_viewer.py
jkulhanek: It does. Is it ok not to print the banner messages (f"Viewer at: {viewer_state.viewer_url}")? The viewer_url is already mentioned in the log, so I guess it is safe to remove.

jkulhanek:
> (`ns-viewer` -- or should it be `ns-view` ?)

Or `ns-start-viewer`?

scripts/viewer/run_viewer.py
jkulhanek: No, it doesn't work without the local_writer

jkulhanek: I still need `writer.setup_local_writer`. Otherwise I get the following error:
image

jkulhanek: I will remove the dependency

scripts/viewer/run_viewer.py
```diff
@@ -0,0 +1,109 @@
...
+        config.viewer = self.viewer.as_viewer_config()
+        config.viewer.num_rays_per_chunk = num_rays_per_chunk
+
+        self._start_viewer(config, pipeline)
+
+    def _start_viewer(self, config: TrainerConfig, pipeline):
+        base_dir = config.get_base_dir()
+        viewer_log_path = base_dir / config.viewer.relative_log_filename
+        viewer_state = None
+        viewer_state, _ = viewer_utils.setup_viewer(
```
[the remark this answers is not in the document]
jkulhanek: I have removed the `writer.setup_local_writer`, but I had to mock `writer.GLOBAL_BUFFER`. Is it ok?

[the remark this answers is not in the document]
tancik: Hmm maybe it is necessary, we use it to store `EventName.VIS_RAYS_PER_SEC` which is used to figure out the resolution of the render.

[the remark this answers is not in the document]
jkulhanek: We can still store these events without setup_local_writer by manually initialising the GLOBAL_BUFFER. What is the reason for avoiding setup_local_writer?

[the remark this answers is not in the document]
tancik: No reason, I just thought it was only necessary for training and therefore not needed here. But since it is also used for other things, we can keep it around.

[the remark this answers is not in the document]
jkulhanek: I added it back.

jkulhanek:
I also updated the docs.

tancik (approved):
LGTM. Thanks for adding this feature, it makes a lot of sense to have.

PR #1385, opened 2023-02-09
Add support for is_fisheye flag in ingp transforms.json

jkulhanek:
This PR adds the support for instant-ngp "is_fisheye" flag in the transforms.json file.


tancik (approved):
LGTM

PR #1386, opened 2023-02-09
make robust and patch_size model params

machenmusik:


machenmusik:
(This should be mergeable into main without breakage.)

PR #1391, opened 2023-02-09
add time

lsongx:
Added time to the `spiral` rendering script

**Known limitations:**
`DyCheck` data set has a different `z` axis. So the video is upside down:

https://user-images.githubusercontent.com/12553253/217924350-bf009510-5aba-462a-80f4-7aaf964d6e58.mp4

A quick fix is to add `-` to the up axis (but this is not included in this PR as it impacts other datasets)

https://user-images.githubusercontent.com/12553253/217924591-b5eca1dd-fb50-4d5f-873b-82020d53cdc7.mp4

Using exported camera path is fine

https://user-images.githubusercontent.com/12553253/217924749-359147fe-397a-4f7b-9507-491f9545169d.mp4

**But** exported camera path is with a fixed timestamp. Perhaps we can improve UI and allow user to specify timestamp for key frames as well?

### DIFF
--- a/nerfstudio/cameras/camera_paths.py
+++ b/nerfstudio/cameras/camera_paths.py
@@ -102,7 +102,7 @@ def get_spiral_path(
 
     times = None
     if camera.times is not None:
-        times = torch.linspace(0, 1, steps)[:,None]
+        times = torch.linspace(0, 1, steps)[:, None]
     return Cameras(
         fx=camera.fx[0],
         fy=camera.fy[0],
--- a/nerfstudio/cameras/camera_paths.py
+++ b/nerfstudio/cameras/camera_paths.py
@@ -100,12 +100,16 @@ def get_spiral_path(
         new_c2ws.append(c2wh[:3, :4])
     new_c2ws = torch.stack(new_c2ws, dim=0)
 
+    times = None
+    if camera.times is not None:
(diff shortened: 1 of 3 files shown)


tancik (approved):
LGTM, thanks for the quick fix.

machenmusik:
> But exported camera path is with a fixed timestamp. Perhaps we can improve UI and allow user to specify timestamp for key frames as well?

Is this still an issue after this PR?  I would think that (relative) time should progress from 0 to 1 linearly by default, which does nothing for static methods

lsongx:
@machenmusik Yes. After exporting camera path, all poses are with the same timestamp. Perhaps we could add a timestamp for each key frame? Timestamps can then be displayed below the camera keyframe selection.
Screenshot 2023-02-10 at 10 22 45 AM

a1600012888:
> nerfstudio-project#1391

Is this still an issue now? Seems that this still is not supported now?  Thanks.

xiazhi1:
> > #1391
> 
> Is this still an issue now? Seems that this still is not supported now? Thanks.
Same question for me,Hope it can be addressed sooner!

xiazhi1:
@lsongx

lsongx:
@a1600012888 @xiazhi1 I think yes. I don't know how to improve the UIs. But currently you may change the json file by your self to specify a time.

PR #1396, opened 2023-02-10
Improve ns-install-cli

jkulhanek:
- All entrypoints are automatically registered for ns-install-cli completions
- ns-install-cli detects conda environment and instead of putting the activation scripts to bashrc or zshrc it uses the activation scripts of the conda environment.

### DIFF
--- a/scripts/completions/install.py
+++ b/scripts/completions/install.py
@@ -252,6 +252,7 @@ def _update_conda_scripts(
     elif mode == "install":
         # Install completions.
         activate_path.parent.mkdir(exist_ok=True, parents=True)
+        deactivate_path.parent.mkdir(exist_ok=True, parents=True)
         with activate_path.open("w+", encoding="utf8") as f:
             f.write(activate_source_lines)
         with deactivate_path.open("w+", encoding="utf8") as f:
--- a/scripts/completions/install.py
+++ b/scripts/completions/install.py
@@ -10,17 +10,24 @@ import shutil
 import stat
 import subprocess
 import sys
-from typing import List, Union
+from typing import List, Optional, Union
 
 import tyro
 from rich.console import Console
 from rich.prompt import Confirm
 from typing_extensions import Literal, assert_never
+from typing_extensions import get_args as typing_get_args
+
+if sys.version_info < (3, 10):
+    import importlib_metadata
+else:
+    from importlib import metadata as importlib_metadata
 
 ConfigureMode = Literal["install", "uninstall"]
 ShellType = Literal["zsh", "bash"]
 
 CONSOLE = Console(width=120)
+HEADER_LINE = "# Source nerfstudio autocompletions."
 
 ENTRYPOINTS = [
     "ns-install-cli",
@@ -31,6 +38,11 @@ ENTRYPOINTS = [
     "ns-render",
     "ns-dev-test",
 ]
+
+
+def _get_all_entry_points() -> List[str]:
+    entry_points = importlib_metadata.distribution("nerfstudio").entry_points
+    return [x.name for x in entry_points]
 
 
 def _check_tyro_cli(script_path: pathlib.Path) -> bool:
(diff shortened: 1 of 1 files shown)


scripts/completions/install.py
```diff
@@ -10,17 +10,24 @@
...
+else:
+    from importlib import metadata as importlib_metadata
 
 ConfigureMode = Literal["install", "uninstall"]
 ShellType = Literal["zsh", "bash"]
 
 CONSOLE = Console(width=120)
+HEADER_LINE = "# Source nerfstudio autocompletions."
 
 ENTRYPOINTS = [
```
brentyi, line 32: this list is obsolete now, right?

brentyi:
I need some more time to fully go through this but just wanted to say this looks really clean @jkulhanek!

Not in scope here, but I'm wondering if there's a clean way to merge some of this logic upstream into `tyro`, especially since the script looks even less nerfstudio-specific now. It'd be really cool if we had a completion setup flow like this for all `tyro.cli()` applications.

scripts/completions/install.py
jkulhanek: Indeed, I removed it in the last commit.

scripts/completions/install.py
```diff
@@ -189,6 +202,101 @@ def _update_rc(
...
+
+    conda_path = pathlib.Path(os.environ["CONDA_PREFIX"])
+    activate_path = conda_path / "etc/conda/activate.d/nerfstudio-activate.sh"
+    if mode == "uninstall":
+        if activate_path.exists():
+            os.remove(activate_path)
+        CONSOLE.log(f":broom: Existing completions uninstalled from {conda_path}.")
+    elif mode == "install":
+        # Install completions.
+        with activate_path.open("w+", encoding="utf8") as f:
```
brentyi, line 227: When testing on my machine the directory didn't exist, I had to add a `activate_path.parent.mkdir(exist_ok=True, parents=True)` above this line

scripts/completions/install.py
```diff
@@ -189,6 +202,101 @@ def _update_rc(
...
+            f.write(source_lines)
+        # TODO: handle also the deactivate case
```
brentyi, line 229: in scope for this PR or no?

scripts/completions/install.py
```diff
@@ -189,6 +202,101 @@ def _update_rc(
...
+
+def _get_conda_path() -> Optional[pathlib.Path]:
+    """
+    Returns the path to the conda environment if
+    the nerfstudio package is installed in one.
+    """
+    conda_path = None
+    if "CONDA_PREFIX" in os.environ:
+        # Conda is active, we will check if
+        # the Nerfstudio is installed in the conda env
```
brentyi, line 246: nit: line break / comment formatting is a little weird

maybe we could also make the comment style consistent within this file, like ending with periods etc?

brentyi (commented):
I went through the install process and only had one minor issue, if that's resolved I think this is good to merge :)

scripts/completions/install.py
jkulhanek: It should be. Unfortunately, I don't know how to do it. Once the completions are sourced it is hard to undo it. I am not sure if it is that big of an issue.

scripts/completions/install.py
jkulhanek: Better?

scripts/completions/install.py
jkulhanek: Thanks for pointing it out. I didn't realise the directory could be missing.

scripts/completions/install.py
brentyi: Yes, thanks :)

scripts/completions/install.py
brentyi: Yeah, I agree it's low priority.

I checked locally, and it looks like this can be done with something like
```
complete -r ns-train
```
in bash[^1], or
```
unset '_comps[ns-train]'
```
in zsh[^2].

Do you see anything wrong with this + do you have time to give implementation a try? If it looks sensible but you don't have time we could also just merge as-is and open another issue. 

[^1]: https://stackoverflow.com/questions/43794270/disable-or-unset-specific-bash-completion
[^2]: https://unix.stackexchange.com/questions/589807/how-to-remove-one-command-completion-from-zsh

jkulhanek: Thanks! I added the implementation. It works for bash, but I haven't tested it with zsh.

scripts/completions/install.py
```diff
@@ -189,6 +224,107 @@ def _update_rc(
...
+    deactivate_path = conda_path / "etc/conda/deactivate.d/nerfstudio_deactivate.sh"
+    if mode == "uninstall":
+        if activate_path.exists():
+            os.remove(activate_path)
+        if deactivate_path.exists():
+            os.remove(deactivate_path)
+        CONSOLE.log(f":broom: Existing completions uninstalled from {conda_path}.")
+    elif mode == "install":
+        # Install completions.
+        activate_path.parent.mkdir(exist_ok=True, parents=True)
```
brentyi, line 254: On my machine, needed to add this for deactivate_path as well

scripts/completions/install.py
brentyi: Works great in zsh for me :)

scripts/completions/install.py
jkulhanek: Sorry I didn't realise that. Added now.

brentyi (approved):

PR #1399, opened 2023-02-11
RealityCapture export ambigous documentation

### LINKED ISSUE #1398: RealityCapture export ambigous documentation
The description in the [CLI documentation for RealityCapture]( https://docs.nerf.studio/en/latest/reference/cli/ns_process_data.html#realitycapture) ambigously mentions export should be done in CSV. But RealityCapture has several csv formats and the correct format isn't even called csv, but is called **internal/external camera parameters** (which is a csv format but isn't called that in RC). This makes the documentation a bit confusing on what to use. 

Screenshot 2023-02-11 024040

DFin:
Fixes #1398

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -120,6 +120,6 @@
     "__atomic": "cpp",
     "__node_handle": "cpp"
   },
-  "python.analysis.typeCheckingMode": "basic",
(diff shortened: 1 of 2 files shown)


DFin:
accidently commited more than the itended file. Closing.

PR #1400, opened 2023-02-11
Switch to Generic model interface

jkulhanek:
This PR changes the Model's class interface to accept the type of the config. This should improve the type hints for the models (as you no longer need to override the __init__ method).
But I am not sure if this is needed or if using Generic could break compatibility for older python versions.
Please feel free to reject this PR if you are not sure of the benefits.

### DIFF
--- a/docs/developer_guides/config.md
+++ b/docs/developer_guides/config.md
@@ -31,14 +31,8 @@ class NerfactoModelConfig(ModelConfig):
     _target: Type = field(default_factory=lambda: NerfactoModel)
     ...
 
-class NerfactoModel(Model):
-     """Nerfacto model
-
-    Args:
-        config: Nerfacto configuration to instantiate model
-    """
-
-    config: NerfactoModelConfig
+class NerfactoModel(Model[NerfactoModelConfig]):
+     """Nerfacto model"""
     ...
 ```
 
--- a/docs/developer_guides/pipelines/models.md
+++ b/docs/developer_guides/pipelines/models.md
@@ -111,14 +111,8 @@ There are a lot of options! Thankfully, our config system makes this easy to han
 Furthermore, you have Python autocomplete and static checking working in your favor. At the top of every Model, we specify the config and then can easily pull of values throughout the implementation. Let's take a look at the beginning of the NerfactoModel implementation.
 
 ```python
-class NerfactoModel(Model):
-    """Nerfacto model
-
-    Args:
-        config: Nerfacto configuration to instantiate model
-    """
-
-    config: NerfactoModelConfig
+class NerfactoModel(Model[NerfactoModelConfig]):
+    """Nerfacto model"""
 
     def populate_modules(self):
         """Set the fields and modules."""
--- a/nerfstudio/data/datamanagers/base_datamanager.py
+++ b/nerfstudio/data/datamanagers/base_datamanager.py
@@ -258,14 +258,12 @@ class DataManager(nn.Module):
 
 
 @dataclass
-class VanillaDataManagerConfig(InstantiateConfig):
+class VanillaDataManagerConfig(InstantiateConfig["VanillaDataManager"]):
     """Configuration for data manager instantiation; DataManager is in charge of keeping the train/eval dataparsers;
     After instantiation, data manager holds both train/eval datasets and is in charge of returning unpacked
     train/eval data at each iteration
     """
 
-    _target: Type = field(default_factory=lambda: VanillaDataManager)
-    """Target class to instantiate."""
     dataparser: AnnotatedDataParserUnion = BlenderDataParserConfig()
     """Specifies the dataparser used to unpack the data."""
     train_num_rays_per_batch: int = 1024
--- a/nerfstudio/data/datamanagers/base_datamanager.py
+++ b/nerfstudio/data/datamanagers/base_datamanager.py
@@ -61,7 +61,7 @@ from nerfstudio.data.utils.dataloaders import (
     FixedIndicesEvalDataloader,
     RandIndicesEvalDataloader,
 )
-from nerfstudio.data.utils.nerfstudio_collate import nerfstudio_collate
+from nerfstudio.data.utils.nerfstudio_collate import CollateFunction, nerfstudio_collate
 from nerfstudio.engine.callbacks import TrainingCallback, TrainingCallbackAttributes
 from nerfstudio.model_components.ray_generators import RayGenerator
 from nerfstudio.utils.misc import IterableWrapper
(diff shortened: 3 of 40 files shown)


docs/developer_guides/config.md
```diff
@@ -31,14 +31,12 @@ class NerfactoModelConfig(ModelConfig):
     _target: Type = field(default_factory=lambda: NerfactoModel)
     ...
 
-class NerfactoModel(Model):
+class NerfactoModel(Model[NerfactoModelConfig]):
      """Nerfacto model
 
     Args:
         config: Nerfacto configuration to instantiate model
     """
```
tancik, line 39: Remove

docs/developer_guides/pipelines/models.md
```diff
@@ -111,15 +111,13 @@ There are a lot of options! Thankfully, our config system makes this easy to han
 Furthermore, you have Python autocomplete and static checking working in your favor. At the top of every Model, we specify the config and then can easily pull of values throughout the implementation. Let's take a look at the beginning of the NerfactoModel implementation.
 
 ```python
-class NerfactoModel(Model):
+class NerfactoModel(Model[NerfactoModelConfig]):
     """Nerfacto model
 
     Args:
         config: Nerfacto configuration to instantiate model
     """
```
tancik, line 119: Remove here and in all other models since it is no longer an argument.

tancik (commented):
This is a great idea! So much cleaner!

jkulhanek:
Now I am not sure if it is a good idea. The api should be consistent, so we would want to apply this also to DataManager and Pipeline classes. The problem is for example in DynamicBatchPipeline. This class inherits from VanillaPipeline with a fixed config type, but it changes the config type on top. I am afraid that implementing this would make the API less consistent.

jkulhanek:
I suggest to close this for now. What do you think?

tancik:
I suppose. I do really like how this makes the code cleaner. I wonder if there are ways we can refactor the `DynamicBatchPipeline` so that it can fit into this framework.

jkulhanek:
This will be elegant in Python 3.12 (https://peps.python.org/pep-0696/). I will try to think of something for the currently supported Python.

jkulhanek:
It may be part of typing_extensions ...

jkulhanek:
What do you think now?

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -352,9 +346,6 @@ def _get_pixel_sampler(  # pylint: disable=no-self-use
         self, dataset: InputDataset, *args: Any, **kwargs: Any
     ) -> PixelSampler:
         """Infer pixel sampler to use."""
-        if self.config.patch_size > 1:
-            return PatchPixelSampler(*args, **kwargs, patch_size=self.config.patch_size)
-
```
tancik, line 357: Merge issue?

jkulhanek: Yes! Sorry for that.

tancik:
It looks good to me. I'd like to get @brentyi opinion.

brentyi (commented):
+1 for looks good, big fan of PRs with more lines deleted than added. There's always more to do but I'd support merging. :)

Only important thing is that this blurb in `docs/developer_guides/config.md` probably needs a tweak:
```
You can then enable type/auto complete on the config passed into the `NerfactoModel` by specifying the config type below the class definition.
```

A pair of thoughts related to this PR, that I don't feel strongly about:
- It would be nice if `.instantiate()` in `InstantiateConfig` was typed correctly. I think this is doable with another generic + some special handling for forward references (since configs for a model are defined before the model). This is a lot of generics but we could get rid of the `_target` fields that are scattered everywhere...
```python
@dataclass
class SomeConfig(InstantiateConfig[Model]):
    ...

class SomeModel(Model[SomeConfig]):
    ...
```
- Alternatively, if we generalize this generic structure to a `Configurable` class that inherits from `Generic[TConfig]`, we can assign `_target` values to config dataclasses automatically/dynamically by overriding `__class_getitem__` in `Configurable`.

nerfstudio/utils/misc.py
```diff
@@ -20,9 +20,12 @@
 from typing import Any, Callable, Dict, List, Optional, Union
 
 import torch
+from typing_extensions import TypeVar
 
+T = TypeVar("T")
```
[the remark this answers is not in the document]
jkulhanek: I had this first, but it failed because, in some cases, T is also `int`.

jkulhanek:
> * Alternatively, if we generalize this generic structure to a `Configurable` class that inherits from `Generic[TConfig]`, we can assign `_target` values to config dataclasses automatically/dynamically by overriding `__class_getitem__` in `Configurable`.

Adding this would introduce one-to-one correspondence between configs and models. This is not the case now, as multiple models share same config.

nerfstudio/utils/misc.py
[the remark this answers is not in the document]
brentyi: Oh interesting, yeah I see, when this function says `stuff` it really means it huh

brentyi:
> Adding this would introduce one-to-one correspondence between configs and models. This is not the case now, as multiple models share same config.

Is this true beyond mipnerf using the vanilla nerf config structure? This is the only one I'm aware of, and it'd be quick to add a dummy class. The base `InstantiateConfig` class could also still have a `_target` field that can be overridden as is currently done, it just wouldn't need to be repeated in every config class definition.

jkulhanek:
Sure, I agree, I will do it.
________________________________

jkulhanek:
Oh, another problem:
```
tyro._instantiators.UnsupportedTypeAnnotationError: Default values for generic subparsers are not supported.
```

brentyi:
Shoot! Are you able to share a fork / branch that causes this error? I can check whether this is supportable or not.

Alternatively, I think it would make sense to get this PR merged before conflicts start creeping in and we can revisit next steps in a separate PR. (if you have time)

jkulhanek:
I have added a proof-of-concept implementation with forward refs (correct types for instantiated objects). Currently, it fails with the error I mentioned. But what do you think of the overall concept?

jkulhanek:
I don’t think that there is a way to make correct type annotations for both the config and the model without forward references. Personally, I would be ok with only having correct annotations for the model and not for the “setup” method of config (avoiding forward refs and using _target). I am also ok with the code in the current PR with forward refs-I believe I can get around the tyro bug. Which one would you prefer?

I also won’t have time to work on it till 9th of March
________________________________

brentyi:
> Personally, I would be ok with only having correct annotations for the model and not for the “setup” method of config (avoiding forward refs and using _target).

Yeah, I also don't feel strongly, but have heard frustration expressed about verbose/confusing typing syntax so let's just go with this.

Maybe we could think about a bigger overhaul at some point, the way PyTorch modules are instantiated + structured makes it hard but the [Implicitron config](https://pytorch3d.org/tutorials/implicitron_config_system) system, [hydra-zen](https://github.com/mit-ll-responsible-ai/hydra-zen), what nerfstudio has etc all have subsets of nice usability things (conciseness, type safety, ease-of-use with inheritance + documentation, ...). It would be nice to get all of them, with less syntax and less tradeoffs...

tancik:
Is this worth reviving, or should we close it for now?

jkulhanek:
Let's just close it.

PR #1410, opened 2023-02-13
add monosdf normal loss and midas scale invariant depth loss

pablovela5620:
Continue integration of Neus (w/ monocular cues) from SDFStudio. More info on plan [here](https://github.com/nerfstudio-project/nerfstudio/pull/1338#issuecomment-1427178041)

Included normal loss specified in monosdf paper, and scale invariant depth loss from MiDas paper (based on omnidata repo implementation)


nerfstudio/model_components/losses.py
```diff
@@ -308,3 +308,234 @@ def depth_loss(
         return urban_radiance_field_depth_loss(weights, termination_depth, predicted_depth, steps, sigma)
 
     raise NotImplementedError("Provided depth loss type not implemented.")
+
+
+def monosdf_normal_loss(normal_pred: torch.Tensor, normal_gt: torch.Tensor) -> torch.Tensor:
+    """normal consistency loss as monosdf https://niujinshuchong.github.io/monosdf/
+    Args:
+        normal_pred (torch.Tensor): volume rendered normal
```
tancik, line 316: No need to include type since it is in arguments.

nerfstudio/model_components/losses.py
```diff
@@ -308,3 +308,234 @@ def depth_loss(
...
+    """
+    # avoid division by 0 (if M = sum(mask) = 0: image_loss = 0)
+    valid = M.nonzero()
+
+    image_loss[valid] = image_loss[valid] / M[valid]
+
+    return torch.mean(image_loss)
+
+
+def mse_loss(prediction: torch.Tensor, target: torch.Tensor, mask: torch.Tensor, reduction=reduction_batch_based):
```
tancik, line 393: The reduction type can be a literal, ie `reduction: Literal["batch", "image"] = "batch"`

nerfstudio/model_components/losses.py
```diff
@@ -308,3 +308,234 @@ def depth_loss(
...
+        M: number of valid pixels of each image
+    Returns:
+        mean of average of valid pixels of an image
+    """
+    # avoid division by 0 (if M = sum(mask) = 0: image_loss = 0)
+    valid = M.nonzero()
+
+    image_loss[valid] = image_loss[valid] / M[valid]
+
+    return torch.mean(image_loss)
```
tancik, line 390: I feel like the logic here can be build directly into the functions, it seems pretty minimal in practice.

nerfstudio/model_components/losses.py
```diff
@@ -308,3 +308,234 @@ def depth_loss(
...
+    M = torch.sum(mask, (1, 2))
+    res = prediction - target
+    image_loss = torch.sum(mask * res * res, (1, 2))
+
+    return reduction(image_loss, 2 * M)
+
+
+def gradient_loss(prediction: torch.Tensor, target: torch.Tensor, mask: torch.Tensor, reduction=reduction_batch_based):
+    """
+    gradient loss from omnidata repo
```
tancik, line 413: Do you have a link? Can you also provide more specific information about what this loss does.

nerfstudio/model_components/losses.py
```diff
@@ -308,3 +308,234 @@ def depth_loss(
...
+    def forward(self, prediction: torch.Tensor, target: torch.Tensor, mask: torch.Tensor):
+        """
+        Args:
+            prediction: predicted depth map
+            target: ground truth depth map
+            mask: mask of valid pixels
+        Returns:
+            mse loss based on reduction function
+        """
+        return mse_loss(prediction, target, mask, reduction=self.__reduction)
```
tancik, line 462: This class doesn't provide any new functionality, I think it can be removed.

nerfstudio/model_components/losses.py
```diff
@@ -308,3 +308,234 @@ def depth_loss(
...
+
+
+class GradientLoss(nn.Module):
+    """
+    regularizer term from MiDaS paper
```
tancik, line 467: Can you provide more information about what this loss does and also documentation for `scales`. Any reason not to add the scales logic to the function defined at L411 and then remove this class? It's unclear to me if there is any reason to have both a class based loss and a functional one.

nerfstudio/model_components/losses.py
```diff
@@ -308,3 +308,234 @@ def depth_loss(
...
+            )
+
+        return total
+
+
+class ScaleAndShiftInvariantLoss(nn.Module):
+    """
+    Scale and shift invariant loss as described in
+    "Towards Robust Monocular Depth Estimation: Mixing Datasets for Zero-shot Cross-dataset Transfer"
+    https://arxiv.org/pdf/1907.01341.pdf
```
tancik, line 508: Add documentation for arguments and properties (`prediction_ssi`).

tancik (commented):
Thanks for working on adding these losses! I left an initial set of comments that mostly have to do with documentation and API. I realize that most of this code was taken from other repos, these comments are mainly given with the goal is making the code follow the style of nerfstudio.

nerfstudio/model_components/losses.py
pablovela5620: Decided to move everything to this Midas loss instead and just consolidate things here. Let me know if that works

pablovela5620:
Okay I think that should have taken care of everything, I won't fully know if these things are broken until I implement the rest of Neus (so I can verify that we are correctly using monocular cues with the learned sdf function) but I can verify when I start working on brining in the `base_surface_model` and `neus`

pablovela5620:
I haven't used tensortyping before but it seems like it doesn't allow ... to the right of a fixed value. For some reason SDFStudio reshapes from ['num_samples', 1] to [1, 32, -1] as you can see here https://github.com/autonomousvision/sdfstudio/blob/1a4c961528be33b2affd9836614f45a487392061/nerfstudio/models/base_surface_model.py#L386

With a sample size of 1024 this ends up being [1,32,32]. To get it working right now I just set it at a fixed 1,32,32 but I'll fix that once I figure out why exactly they changed things like that (probably due to the way they need to compute scale/shift). It will be easier once I start working on the other portions of the port so that I can actually train the model with the Neus method

pablovela5620:
Appreciate the patience. Should be all resolved now.

nerfstudio/utils/math.py
```diff
@@ -285,3 +286,26 @@ def safe_normalize(
...
     """
     return vectors / (torch.norm(vectors, dim=-1, keepdim=True) + eps)
+
+
+def masked_reduction(loss: TensorType[0], M: TensorType[0], reduction_type: Literal["image", "batch"]):
+    """
+    Whether to consolidate the loss across the batch or across the image
+    Args:
+        loss: loss tensor
+        M: mask tensor
```
tancik, line 296: Rename `loss` -> `input_tensor`, and `M` -> `mask` and update docs to not be loss specific.

nerfstudio/model_components/losses.py
```diff
@@ -308,3 +310,218 @@ def depth_loss(
...
+    ) -> TensorType[0]:
+        """
+        Args:
+            prediction: predicted depth map
+            target: ground truth depth map
+            mask: mask of valid pixels
+        Returns:
+            mse loss based on reduction function
+        """
+        M = torch.sum(mask, (1, 2))
```
tancik, line 356: Avoid single letter variables.

tancik (changes requested):
Couple of small remaining comments. Thanks for bearing with me. Should be good after resolved.

pablovela5620:
okay should be good now!

nerfstudio/utils/math.py
```diff
@@ -288,24 +288,24 @@ def safe_normalize(
     return vectors / (torch.norm(vectors, dim=-1, keepdim=True) + eps)
 
 
-def masked_reduction(loss: TensorType[0], M: TensorType[0], reduction_type: Literal["image", "batch"]):
+def masked_reduction(input_tensor: TensorType[0], mask: TensorType[0], reduction_type: Literal["image", "batch"]):
     """
     Whether to consolidate the loss across the batch or across the image
```
tancik, line 293: nit: `loss` -> `input_tensor`

nerfstudio/utils/math.py
```diff
@@ -288,24 +288,24 @@ def safe_normalize(
     return vectors / (torch.norm(vectors, dim=-1, keepdim=True) + eps)
 
 
-def masked_reduction(loss: TensorType[0], M: TensorType[0], reduction_type: Literal["image", "batch"]):
+def masked_reduction(input_tensor: TensorType[0], mask: TensorType[0], reduction_type: Literal["image", "batch"]):
```
tancik, line 291: I missed this earlier, I think the size is incorrect here, `TensorType[0]` is a single number.

tancik (approved):
LGTM

f-dy:
@pablovela5620 isn't the "scale invariant depth loss from MiDas paper" supposed to be on disparity (inverse depth) rather than depth?

Issue #1424, opened 2023-02-14
Wrong data range for LPIPS calculation?


iamNCJ:
In nerfstudio, the LPIPS metrics in all models are calculated with torchmetrics `LearnedPerceptualImagePatchSimilarity` module, like [here](https://github.com/nerfstudio-project/nerfstudio/blob/9808ebbcb8cc0be355c688b5fab8e5e45955b5ad/nerfstudio/models/instant_ngp.py#L140), but actually, as they noted in their [document](https://torchmetrics.readthedocs.io/en/v0.11.1/image/learned_perceptual_image_patch_similarity.html#:~:text=normalize%20(bool)%20%E2%80%93%20by%20default%20this%20is%20False%20meaning%20that%20the%20input%20is%20expected%20to%20be%20in%20the%20%5B%2D1%2C1%5D%20range.%20If%20set%20to%20True%20will%20instead%20expect%20input%20to%20be%20in%20the%20%5B0%2C1%5D%20range.), this module expects data range to be `[-1, 1]`, but it seems that currently rgb values with data range `[0, 1]` are sent in and thus will cause the LPIPS metrics to be lower & better than they should be.

This problem is easy to be resolved by setting `normalize` to `True` when instantiating the `LearnedPerceptualImagePatchSimilarity` module.

```python
self.lpips = LearnedPerceptualImagePatchSimilarity(normalize=True)
```

tancik:
Fixed in #1425

PR #1425, opened 2023-02-14
Fix LPIPS data range

iamNCJ:
Change the expected data range of LPIPS module (`LearnedPerceptualImagePatchSimilarity`) from [-1, 1] to [0, 1] as mentioned in #1424


tancik (approved):
Thanks for adding this fix!

Issue #806, opened 2022-10-19
Poor performance on NeRF Blender Synthetic Data & potential bugs


dragonlong:
**Describe the bug**
Hi, when I was trying to test nerfstudio-0.1.5 on the original NeRF dataset by using `./scripts/benchmarking/launch_train_blender.sh -m nerfacto -g 0`, it seems the training loss for some datasets will suddenly jump into NaN for some cases, or the losses will not converge well after 2k steps in my case. Another concern is that the testing performance doesn't seem to match with expected performance, the psnr should reach > 30dB after some training. Is there anything special that I need to take care of when doing this testing? I have attached the screenshots below to help diagnosis. Thanks!

**To Reproduce**
Use `nerfstudio-0.1.5`, and run `./scripts/benchmarking/launch_train_blender.sh -m nerfacto -g 0` on NeRF blender dataset

**Expected behavior**
The performance should at least match with the performance reported with previous NeRF papers, like >23dB for drums, and >30dB for others

**Screenshots**
<img width="1686" alt="image" src="https://user-images.githubusercontent.com/8425887/196750189-d6b516ea-3170-453a-90f4-56b796647f11.png">
<img width="1685" alt="image" src="https://user-images.githubusercontent.com/8425887/196751594-9129111e-9473-496a-8561-d048cd27f8f2.png">

tancik:
It looks like you are running the `nerfacto` model. This model is designed for real world unbounded scenes, as such the configuration is not ideal for unbounded synthetic scenes like the blender dataset. For example the following flags should be set: `--pipeline.model.near-plane 2. --pipeline.model.far-plane 6. --pipeline.datamanager.camera-optimizer.mode off --pipeline.model.use-average-appearance-embedding False`.
These alone aren't sufficient, other methods hardcode into the model that the background is white which is not done in `nerfacto` and other method uniformly sample the space which is also not done in `nerfacto`.

dragonlong:
Hi Matthew, 

Thanks for the feedback! I have launched again the experiments with both instant-ngp and nerfacto, now the training profile looks normal to me at least from the training side. Will close this ticket afterwards
![image](https://user-images.githubusercontent.com/8425887/197074712-15ccdfd2-5f63-4ce5-b17a-da809bdd419d.png)

dubrovin-sudo:
[...]

PR #1441, opened 2023-02-16
Allow use of black or white background in Nerfacto models

iSach:
For synthetic data like the Blender dataset, the background is white and thus it is more appropriate to use a white value for untrained areas, as done in Instant-NGP. 

Although already partly implemented, the option was simply not allowed.

- I cleaned the use of the color for the background to be consistent between methods.
- I allowed white and black in Nerfacto and NerfPlayer-Nerfacto
- I updated the blender benchmark script to use the white background in addition to the other methods mentioned in issue #806.

On the Blender benchmark, losses are obviously lower, and seem to converge faster.
image
image


iSach:
It reduces artifacts on the object for blender results but for some reason adds more artifacts around the object, which it is exactly what it is supposed to fix.

However, it greatly helps quantitavely. For the Hot Dog scene, the PSNR goes from 19 to 27.

image
image

iSach:
Also, surprisingly, it does not seem sufficient whereas for Instant-NGP, this single option suffices to fix the training. Nerfacto for some reason requires additional tuning (especially modifying the far/near plane parameters) that Instant-NGP does not require.

Does anybody have an idea why?

iSach:
A last note: maybe it could be interesting or useful to some data to allow the use of any color (or at least all colors in the "COLORS_DICT"). It would require no change than allowing these color strings in the method parameters.

tancik (approved):
LGTM, thanks for adding

tancik:
For better blender results you probably want to disable scene contraction, and use a uniform sampling for the first step of the proposal sampler (instead of piecewise sampling).

tancik:
To test for lint errors locally, run `ns-dev-test`

iSach:
> For better blender results you probably want to disable scene contraction, and use a uniform sampling for the first step of the proposal sampler (instead of piecewise sampling).

If I'm not wrong (sorry for that, still new and navigating the whole code), it's not possible with current parameters of the model right? Just to know, if so it could be a good idea to add them and I might look at it, in order to better support synthetic data.

tancik:
Feel free to add the sampler parameter to the model.

sam598:
> It reduces artifacts on the object for blender results but for some reason adds more artifacts around the object, which it is exactly what it is supposed to fix.
> 
> However, it greatly helps quantitavely. For the Hot Dog scene, the PSNR goes from 19 to 27.
> 
> image image

The artifacts are happening because you can't just train a solid color for a transparent object. You need to use alpha/transparency to control density.

iSach:
considering the way the background color gets added to the total color with (1-accumulated weights) * background color, shouldn’t it train it to have zero density to be close to the background color? I don’t see alpha being used right?Le 23 mars 2023 à 22:52, sam598 ***@***.***> a écrit :﻿

It reduces artifacts on the object for blender results but for some reason adds more artifacts around the object, which it is exactly what it is supposed to fix.
However, it greatly helps quantitavely. For the Hot Dog scene, the PSNR goes from 19 to 27.
 

The artifacts are happening because you can't just train a solid color for a transparent object. You need to use alpha/transparency to control density.

sam598:
alpha is an explicit value. background color is ambiguous.

PR #1446, opened 2023-02-17
Equirectangular input

THE-COB:
- Added 2 presets for sampleing equirectangular images (one with 8 images per frame and one with 14)
- Added equirectangular camera type to ns-process-data images and video


scripts/process_data.py
```diff
@@ -264,6 +278,12 @@ def main(self) -> None:
             self.data, image_dir=image_dir, num_frames_target=self.num_frames_target, verbose=self.verbose
         )
 
+        # Generate planar projections if equirectangular
+        if self.camera_type == "equirectangular":
+            image_dir = process_data_utils.generate_planar_projections_from_equirectangular(
+                image_dir, (1200, 1200), self.images_per_equirect
```
machenmusik, line 284: Is hardcoding 1200x1200 resolution the right way to go here? Seems that it should be dependent on the input resolution...

THE-COB: Matt suggested I do that since it gets down sampled for training anyway. What would you recommend though?

machenmusik: - Does power of two matter/help, if downsampling?
- Sounds like Matt's suggesting no higher than that, which makes sense currently; I'm wondering whether people will try with 360 videos with less than 4K resolution, for which 1200x1200 might actually be larger than necessary.

machenmusik: Maybe it begs the question overall of: what is the highest image input resolution that nerfstudio will effectively use?  Not just an equirectangular-input question.

THE-COB: do you think it should be set to like a proportion?

scripts/process_data.py
```diff
@@ -264,6 +278,12 @@ def main(self) -> None:
             self.data, image_dir=image_dir, num_frames_target=self.num_frames_target, verbose=self.verbose
         )
 
+        # Generate planar projections if equirectangular
+        if self.camera_type == "equirectangular":
+            image_dir = process_data_utils.generate_planar_projections_from_equirectangular(
```
THE-COB, line 283: There's an issue with where the images are being put at the end of the video processing

nerfstudio/process_data/process_data_utils.py
```diff
@@ -469,3 +479,72 @@ def save_mask(
...
+        BarColumn(),
+        TaskProgressColumn(show_speed=True),
+        ItersPerSecColumn(suffix="equirect frames/s"),
+        TimeRemainingColumn(elapsed_when_finished=True, compact=True),
+    )
+
+    curr_im = 0
+    with progress:
+        for i in progress.track(os.listdir(frame_dir), description="", total=num_ims):
+            if i.endswith(".jpg") or i.endswith(".png") or i.endswith(".jpeg") or i.endswith(".JPG"):
```
tancik, line 536: `i.lower().endswith(('.jpg', '.jpeg', '.png'))`

scripts/process_data.py
tancik: I'm not sure of a great way to choose the resolution. I agree that 1200 might be too much in many cases. One possible heuristic would be
`num_image * res**2 = orig_height*orig_width`

nerfstudio/process_data/process_data_utils.py
cvachha: Also add ".JPG" since those images aren't recognized as ".jpg"

nerfstudio/process_data/process_data_utils.py
```diff
@@ -469,3 +479,72 @@ def save_mask(
         cv2.imwrite(str(mask_path_i), mask_i)
     CONSOLE.log(":tada: Generated and saved masks.")
     return mask_path / "mask.png"
+
+
+def generate_planar_projections_from_equirectangular(
```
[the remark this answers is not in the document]
THE-COB: yeah sure but it's only gonna have the 1 function in it right?

nerfstudio/process_data/process_data_utils.py
THE-COB: I think that's why matt did i.lower() so we wouldn't have to also do .JPG

nerfstudio/process_data/process_data_utils.py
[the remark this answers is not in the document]
tancik: I think thats fine. I can imagine other equirectangular related functions in the future.

scripts/process_data.py
THE-COB: I just tested out this heuristic, it works very well (n=1)

nerfstudio/process_data/equirect_utils.py
```diff
@@ -0,0 +1,120 @@
...
+    returns:
+        The target resolution of the perspective projections.
+    """
+
+    for i in os.listdir(image_dir):
+        if i.lower().endswith((".jpg", ".png", ".jpeg")):
+            im = np.array(cv2.imread(os.path.join(image_dir, i)))
+            res_squared = (im.shape[0] * im.shape[1]) / num_images
+            return (int(np.sqrt(res_squared)), int(np.sqrt(res_squared)))
+    return (1200, 1200)
```
tancik, line 120: You should raise an error if no images are found.

tancik (approved):
LGTM

antoinebio:
Hi,
thanks for the great work effort you put into nerfstudio.

would you please manage TIF equirectangular photo input as well when specifying the source for 360° images ?
I tried to modify those 2 parts of https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/process_data/equirect_utils.py

image


image


I reported that here https://github.com/yzslab/nerfstudio/issues/1#issuecomment-1812098807

but I mention the suggestion also in that post.

THE-COB:
> Hi, thanks for the great work effort you put into nerfstudio.
> 
> would you please manage TIF equirectangular photo input as well when specifying the source for 360° images ? I tried to modify those 2 parts of https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/process_data/equirect_utils.py
> 
> image
> 
> image
> 
> I reported that here [yzslab#1 (comment)](https://github.com/yzslab/nerfstudio/issues/1#issuecomment-1812098807)
> 
> but I mention the suggestion also in that post.

https://github.com/nerfstudio-project/nerfstudio/pull/2613

Issue #1412, opened 2023-02-13
"hloc" is hard to setup in docker image dromni/nerfstudio:0.1.17


edwardchenomni:
**Describe the bug**
While the instruction to manually install hloc for using --sfm-tool hloc in ns-process-data command is clearly shown.
It hard to achieve the manual installation process when using "docker image dromni/nerfstudio:0.1.17".

**To Reproduce**
1. run the container `docker run --gpus all --rm -it dromni/nerfstudio:0.1.17`
2. follow the installation instruction on [Hierarchical-Localization](https://github.com/cvg/Hierarchical-Localization)
`git clone --recursive https://github.com/cvg/Hierarchical-Localization/`
`cd Hierarchical-Localization/`
3. install hloc `python3 -m pip install -e .`

**Expected behavior**
Instead of smoothly installed, an error popped up complaining about:
`ERROR: Could not find a version that satisfies the requirement pycolmap>=0.3.0 (from hloc) (from versions: none)`
`ERROR: No matching distribution found for pycolmap>=0.3.0`

**Screenshots**
![image](https://user-images.githubusercontent.com/125268002/218446033-ee5e5f70-86bf-4341-9184-b248124452f4.png)

It seems like there no pycolmap candidate for python 3.10?
Is there any easy fix on this?

Best,
Edward Chen

pablovela5620:
Looks like they just recently added support for 3.10 [here](https://github.com/colmap/pycolmap/pull/109). I would say that it may be worth building from source in the docker file since it looks like for GPU support that's needed. 

<img width="947" alt="image" src="https://user-images.githubusercontent.com/25287427/219448525-e24122f6-1d13-4c76-827f-d203593ee402.png">

Zunhammer:
Regarding the webpage, it is not yet available for Python 3.10 and I would avoid to change the python version in the whole docker image. As pablovela5620 mentioned, probably the best way would be to build from source into the image. If you need help doing this feel free to contact me directly (NicZer0 on nerfstudio discord) or if there is multiple people who need this I could also add it to the dromni/nerfstudio image.

Zunhammer:
@edwardchenomni I addd hloc to the 0.1.18 image (dromni/nerfstudio:0.1.18). For me it seems to work fine, I would appreciate if you could test and verify that it is working and close the issue if so.
Best,
Nico

dignakov:
Not the OP, but I was trying to get hloc to work as well when I came across this issue.

[...]

PR #1448, opened 2023-02-18
Bump Dockerfile versions, add user to sudoers list.

Zunhammer:
- Bump CUDA from 11.7 to 11.8 (should improve computation times for new GPUs a little)
- Bump Colmap from 3.7 to 3.8
- Added the user ('user') to sudoers list withput password prompting. Now you can use elevated rights, e.g. by typing ```sudo apt update```
- Added curl (required for some download tasks)
- Some small improvements
- Updated docs


pablovela5620:
May be out of scope but would also be awesome to add in bloc support to docker file https://github.com/nerfstudio-project/nerfstudio/issues/1412

tancik (approved):
LGTM, thanks for keeping this up-to-date! Note that there is now a dev container - https://github.com/nerfstudio-project/nerfstudio/blob/8e4ae6bebfc3a93f351d63456ddd3f3b46c515dc/.devcontainer/docker-compose.yml#L6
I think it is pointing to the most up to date version.

Zunhammer:
Hey Tancik,
thanks for pointing this out, I actually missed the docker-compose file. It is pointing to the latest version of the image and is only pre-configuring all the parameters you'd usually have to give, so everything fine. I also was thinking of doing this, good to know that someone else alrady did :)

I can also have a look into the mentioned issue, however this is independant of this PR.

Best,
Nico

PR #1454, opened 2023-02-19
Fix colmap bug when not all images are matched

tancik:
An introduced bug would cause incorrect `transforms.json` files if colmap did not match all of the images as the `filename_map` would be incorrect.

This fix reverts to existing behavior. However this fix will break the functionality introduced in #1371

FYI - @jkulhanek

### DIFF
--- a/nerfstudio/process_data/colmap_utils.py
+++ b/nerfstudio/process_data/colmap_utils.py
@@ -604,14 +604,6 @@ def colmap_to_json(
     cameras = read_cameras_binary(cameras_path)
     images = read_images_binary(images_path)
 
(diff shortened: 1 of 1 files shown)


jkulhanek:
How was the filename_map incorrect?

tancik:
Here is an example of the map when some images weren't matched.

image

jkulhanek:
Why was e.g. frame_00005 missing from the images folder when the images were renamed here: https://github.com/nerfstudio-project/nerfstudio/blob/e2dadc20ab5be3ba41ed1fe104c37272f3631018/nerfstudio/process_data/process_data_utils.py#L151 ? Did you skip the processing of images?
________________________________

PR #1459, opened 2023-02-20
ns-process-data: fix image renaming for the --skip-colmap case

jkulhanek:
This PR (hopefuly) fixes problems introduced in #1366 and partially fixed in ( #1454 ). I am sorry for the problems introduced by the first PR.

@tancik , can you please take a look? This should fix the problems in the original PR while adding back the functionality of (--skip-colmap).

### DIFF
--- a/nerfstudio/process_data/process_data_utils.py
+++ b/nerfstudio/process_data/process_data_utils.py
@@ -249,7 +249,7 @@ def copy_images(data: Path, image_dir: Path, verbose) -> OrderedDict[Path, Path]
         image_dir: Path to the output directory.
         verbose: If True, print extra logging.
     Returns:
-        The number of images copied.
+        The mapping from the original filenames to the new ones.
     """
     with status(msg="[bold yellow]Copying images...", spinner="bouncingBall", verbose=verbose):
         image_paths = list_images(data)
--- a/nerfstudio/process_data/process_data_utils.py
+++ b/nerfstudio/process_data/process_data_utils.py
@@ -24,7 +24,7 @@ from typing import List, Optional, Tuple
 import cv2
 import numpy as np
 from rich.console import Console
-from typing_extensions import Literal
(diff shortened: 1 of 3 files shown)


nerfstudio/process_data/process_data_utils.py
```diff
@@ -241,7 +241,7 @@ def copy_and_upscale_polycam_depth_maps_list(
     return copied_depth_map_paths
 
 
-def copy_images(data: Path, image_dir: Path, verbose) -> int:
+def copy_images(data: Path, image_dir: Path, verbose) -> OrderedDict[Path, Path]:
     """Copy images from a directory to a new directory.
 
     Args:
```
tancik, line 247: Update `Returns` doc

tancik (commented):
Tested, seems to be working. Thanks for the quick fix.

scripts/process_data.py
```diff
@@ -129,7 +131,11 @@ def main(self) -> None:
         # Copy and downscale images
         if not self.skip_image_processing:
             # Copy images to output directory
-            num_frames = process_data_utils.copy_images(self.data, image_dir=image_dir, verbose=self.verbose)
+            image_rename_map_paths = process_data_utils.copy_images(
+                self.data, image_dir=image_dir, verbose=self.verbose
+            )
+            image_rename_map = OrderedDict((a.name, b.name) for a, b in image_rename_map_paths.items())
```
tancik, line 137: Isn't `image_rename_map_paths` already an `OrderedDict`?

jkulhanek: It is. Here the rename_map takes the filenames as strings.

brentyi: just wondering: why not use a regular dict?

jkulhanek: No reason, switching to dict.

tancik (approved):
Lgtm! Thanks for fixing.

PR #1464, opened 2023-02-21
Uniform Sampler in Nerfacto

iSach:
- Added a parameter in the model config of Nerfacto to replace the piecewise initial sampler by a uniform one, preferred for bounded scenes like the Blender dataset.
- Added this parameter when benchmarking Blender on Nerfacto.
- Removed small debug (I guess?) code in `ray_samplers.py` which printed the scene's box each time (for NGP models).


tancik (approved):
LGTM

PR #1466, opened 2023-02-21
Refactor schedulers

tancik:
The schedulers code now uses the same structure that other classes use.

### DIFF
--- a/nerfstudio/engine/schedulers.py
+++ b/nerfstudio/engine/schedulers.py
@@ -32,7 +32,7 @@ class SchedulerConfig(InstantiateConfig):
     """_target: target class to instantiate"""
 
 
-class Scheduler:
+class Scheduler:  # pylint: disable=too-few-public-methods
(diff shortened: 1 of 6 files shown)


nerfstudio/engine/schedulers.py
```diff
@@ -25,92 +26,71 @@
...
-    """Basic scheduler config with self-defined exponential decay schedule"""
+    """Basic scheduler config"""
 
-    _target: Type = field(default_factory=lambda: ExponentialDecaySchedule)
+    _target: Type = field(default_factory=lambda: Scheduler)
+    """_target: target class to instantiate"""
+
+
+class Scheduler:  # pylint: disable=too-few-public-methods
+    """Exponential learning rate decay function."""
```
brentyi, line 36: @tancik looks like you missed a docstring?

tancik: Thanks for catching

Issue #1465, opened 2023-02-21
Specifying "mask_path" in dataset causes runtime error


emd4600:
**Describe the bug**
It is impossible to train datasets with mask on Nerfacto due to a runtime error.

**To Reproduce**
1. Make a dataset that uses `mask_path`
2. Attempt to train it (I tried with Nerfacto, I don't know if it fails on other methods)

**Expected behavior**
After the dataset loads, the following error appears:
```
....
  File "/home/eric/nerfstudio/nerfstudio/engine/trainer.py", line 203, in train
    loss, loss_dict, metrics_dict = self.train_iteration(step)
  File "/home/eric/nerfstudio/nerfstudio/utils/profiler.py", line 43, in wrapper
    ret = func(*args, **kwargs)
  File "/home/eric/nerfstudio/nerfstudio/engine/trainer.py", line 371, in train_iteration
    _, loss_dict, metrics_dict = self.pipeline.get_train_loss_dict(step=step)
  File "/home/eric/nerfstudio/nerfstudio/utils/profiler.py", line 43, in wrapper
    ret = func(*args, **kwargs)
  File "/home/eric/nerfstudio/nerfstudio/pipelines/base_pipeline.py", line 255, in get_train_loss_dict
    ray_bundle, batch = self.datamanager.next_train(step)
  File "/home/eric/nerfstudio/nerfstudio/data/datamanagers/base_datamanager.py", line 418, in next_train
    batch = self.train_pixel_sampler.sample(image_batch)
  File "/home/eric/nerfstudio/nerfstudio/data/pixel_samplers.py", line 197, in sample
    pixel_batch = self.collate_image_dataset_batch(
  File "/home/eric/nerfstudio/nerfstudio/data/pixel_samplers.py", line 99, in collate_image_dataset_batch
    collated_batch = {
  File "/home/eric/nerfstudio/nerfstudio/data/pixel_samplers.py", line 100, in <dictcomp>
    key: value[c, y, x] for key, value in batch.items() if key != "image_idx" and value is not None
RuntimeError: indices should be either on cpu or on the same device as the indexed tensor (cpu)
```

**Potential solution**
[...]

PR #1467, opened 2023-02-21
semantic_nerfw_configurability 

nikmo33:
Fixes an issue currently with the semantic nerf vis, where currently all dataparsers pass in colours as a list of tuples, whereas the typing was enforced to be a tensor. Changes the type of colors in Semantics class to match that of classes, and convert to tensor within semantic nerf before indexing.

Currently the nerfacto model  does not pass any gradients from the seg head into any weights used for rgb rendering, from local experimentation I found that passing gradients through to these parameters improves the segmentation performance significantly, while only causing a very minor regression in the rgb metrics ( from 27.2 to 27.1 PSNR). Hence, adding this is a configrable option during training. Due to this - scaling segmentation loss also becomes more important and hence exposing this as an additional hyper-parameter as well

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -3,7 +3,7 @@ version: "3.8"
 services:
   nerfstudio:
     shm_size: '12gb'
-    image: dromni/nerfstudio:0.1.17
+    image: dromni/nerfstudio:0.1.18
     # build:
     #   context: ..
     #   dockerfile: ./Dockerfile
--- a/Dockerfile
+++ b/Dockerfile
@@ -105,12 +105,24 @@ ENV PATH="${PATH}:/home/user/.local/bin"
 SHELL ["/bin/bash", "-c"]
 
 # Upgrade pip and install packages.
-RUN python3.10 -m pip install --upgrade pip setuptools pathtools promise
+RUN python3.10 -m pip install --upgrade pip setuptools pathtools promise pybind11
 # Install pytorch and submodules (Currently, we still use cu116 which is the latest version for toch 1.12.1 and is compatible with CUDA 11.8).
 RUN python3.10 -m pip install torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116
 # Install tynyCUDNN (we need to set the target architectures as environment variable first).
 ENV TCNN_CUDA_ARCHITECTURES=${CUDA_ARCHITECTURES}
 RUN python3.10 -m pip install git+https://github.com/NVlabs/tiny-cuda-nn.git#subdirectory=bindings/torch
+
+# Install pycolmap 0.3.0, required by hloc.
+RUN git clone --branch v0.3.0 --recursive https://github.com/colmap/pycolmap.git && \
+    cd pycolmap && \
+    python3.10 -m pip install . && \
+    cd ..
+
+# Install hloc master (last release (1.3) is too old) as alternative feature detector and matcher option for nerfstudio.
+RUN git clone --branch master --recursive https://github.com/cvg/Hierarchical-Localization && \
+    cd Hierarchical-Localization && \
+    python3.10 -m pip install -e . && \
+    cd ..
 
 # Copy nerfstudio folder and give ownership to user.
 ADD . /home/user/nerfstudio
--- a/docs/quickstart/blender_addon.md
+++ b/docs/quickstart/blender_addon.md
@@ -11,7 +11,7 @@ This Blender add-on allows for compositing with a Nerfstudio render as a backgro
 
 ## Prerequisites
 
-[Blender](https://www.blender.org/) on Mac or Windows. The python file add-on can be downloaded from the main repo in the `scripts/blender` folder. Before starting, export the mesh or point cloud representation of the NeRF from Nerfstudio, which will be used as reference for the actual NeRF in the Blender scene. Mesh export at a good quality is preferred, however, if the export is not clear or the NeRF is very large, a detailed point cloud export will also work.
+First install [Blender](https://www.blender.org/). The python file add-on can be downloaded from the main repo in the `scripts/blender` folder. Before starting, export the mesh or point cloud representation of the NeRF from Nerfstudio, which will be used as reference for the actual NeRF in the Blender scene. Mesh export at a good quality is preferred, however, if the export is not clear or the NeRF is very large, a detailed point cloud export will also work.
 
 ## Add-on Setup Instructions
 
--- a/docs/quickstart/installation.md
+++ b/docs/quickstart/installation.md
@@ -1,8 +1,35 @@
 # Installation
 
 ## Prerequisites
+::::::{tab-set}
+:::::{tab-item} Linux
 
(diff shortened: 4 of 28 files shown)


tancik (approved):
LGTM, thanks for updating!

machenmusik:
trying this, updated from main, I see this runtime error, which may be from main and not this changeset:

```
│ D:\code\nerfstudio\nerfstudio\data\pixel_samplers.py:100 in <dictcomp>                           │
│                                                                                                  │
│    97 │   │                                                                                      │
│    98 │   │   c, y, x = (i.flatten() for i in torch.split(indices, 1, dim=-1))                   │
│    99 │   │   collated_batch = {                                                                 │
│ ❱ 100 │   │   │   key: value[c, y, x] for key, value in batch.items() if key != "image_idx" an   │
│   101 │   │   }                                                                                  │
│   102 │   │                                                                                      │
│   103 │   │   assert collated_batch["image"].shape == (num_rays_per_batch, 3), collated_batch[   │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
RuntimeError: indices should be either on cpu or on the same device as the indexed tensor (cpu)
```

machenmusik:
Applying the recommended fix for #1465 and commenting out semantic_nerfw.py:223 allows the model to train.

nikmo33:
Hmm @machenmusik i don't see either of those issues while training locally. 
Since colors are a List I'm not sure the given fix will work for my case. 
And I do not see the mask issue raised when training locally. Do you have a branch / command I can use to reproduce locally?

nikmo33:
Also the mask changes cause my training times to go down around 10x, and runs successfully when I revert all the changes made

machenmusik:
To be clear, you are saying that in your configuration, you can revert https://github.com/nerfstudio-project/nerfstudio/pull/1467/commits/d3c9c829ba950dd179764cf510ac9d3094f934ba and it doesn't throw errors while training semantic-nerfw (as well as being 10x faster), correct?

nikmo33:
> d3c9c82

Yes reverting this commit does not lead to training errors for me. I think the semantic issues you are facing are due to us using different dataparsers. I am using the minimal_dataparser. I will make the required changes to fix this. 

> To be clear, you are saying that in your configuration, you can revert [d3c9c82](https://github.com/nerfstudio-project/nerfstudio/commit/d3c9c829ba950dd179764cf510ac9d3094f934ba) and it doesn't throw errors while training semantic-nerfw (as well as being 10x faster), correct? Do you also need to revert the constructor change to avoid the log message? I am not sure the other changes should be reverted, particularly the semantics output option which kills the viewer for me when selected.

Yes reverting the given commit causes a 10x speedup and still runs fine. 
I think the issue was the mismatch between the friends-dataparser and the minimal-dataparser. To be consistent with before Ive just changed the minimal dataparser to match the correct type

machenmusik:
Thanks for making changes; unfortunately, I am getting this error at runtime
```
│    97 │   │                                                                                      │
│    98 │   │   c, y, x = (i.flatten() for i in torch.split(indices, 1, dim=-1))                   │
│    99 │   │   collated_batch = {                                                                 │
│ ❱ 100 │   │   │   key: value[c, y, x] for key, value in batch.items() if key != "image_idx" an   │
│   101 │   │   }                                                                                  │
│   102 │   │                                                                                      │
│   103 │   │   assert collated_batch["image"].shape == (num_rays_per_batch, 3), collated_batch[   │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
RuntimeError: indices should be either on cpu or on the same device as the indexed tensor (cpu)
```
`

nikmo33:
> raise ValueError("mask")
Command - `ns-train semantic-nerfw --data data/friends/ELR-apartment/ --vis tensorboard friends-data`
I cannot reproduce -
```
with your suggestion on line 92 -
    ret = func(*args, **kwargs)
  File "/home/nikhil/nerfstudio-main/nerfstudio/pipelines/base_pipeline.py", line 255, in get_train_loss_dict
    ray_bundle, batch = self.datamanager.next_train(step)
  File "/home/nikhil/nerfstudio-main/nerfstudio/data/datamanagers/base_datamanager.py", line 427, in next_train
    batch = self.train_pixel_sampler.sample(image_batch)
  File "/home/nikhil/nerfstudio-main/nerfstudio/data/pixel_samplers.py", line 198, in sample
    pixel_batch = self.collate_image_dataset_batch(
  File "/home/nikhil/nerfstudio-main/nerfstudio/data/pixel_samplers.py", line 92, in collate_image_dataset_batch
    raise ValueError("mask")
ValueError: mask

```
```
without - 
Step (% Done)       Train Iter (time)    ETA (time)           Train Rays / Sec      
----------------------------------------------------------------------------------- 
110 (0.37%)         29.405 ms            14 m, 38 s           139.34 K              
120 (0.40%)         29.227 ms            14 m, 33 s           140.18 K              
130 (0.43%)         29.089 ms            14 m, 28 s           140.84 K              
140 (0.47%)         29.422 ms            14 m, 38 s           139.26 K              
150 (0.50%)         29.720 ms            14 m, 47 s           137.91 K              
160 (0.53%)         29.744 ms            14 m, 47 s           137.82 K              
170 (0.57%)         29.364 ms            14 m, 35 s           139.57 K              
180 (0.60%)         29.364 ms            14 m, 35 s           139.59 K              
190 (0.63%)         29.662 ms            14 m, 44 s           138.21 K              
200 (0.67%)         29.975 ms            14 m, 53 s           136.75 K  
```

nikmo33:
@machenmusik are you able to run this command on the main branch? If not, then I say we merge this PR, and open an issue and debug separately

machenmusik:
Agreed, although I am a little concerned that the reference dataset for semantic-nerfw at the moment doesn't seem to work with it anymore... was trying to ensure that semantic-nerfw has not inadvertenttly changed here to support only your datasets (which I am not sure are available to others).

If someone else can quickly confirm that friends dataset also fails for them with simple ns-train against main / 0.1.18, as it does for me, yes let's track that as separate issue and merge this.

nikmo33:
> Agreed, although I am a little concerned that the reference dataset for semantic-nerfw at the moment doesn't seem to work with it anymore... was trying to ensure that semantic-nerfw has not inadvertently changed here to support only your datasets (which I am not sure are available to others).
> 
> 
> 
> If someone else can quickly confirm that friends dataset also fails for them with simple ns-train against main / 0.1.18, as it does for me, yes let's track that as separate issue and merge this.  

I'm a bit lost, does semantic nerf-w fail for you both on this branch and main? 
Friends data trains successfully for me as shown above.

machenmusik:
Yes, fails on both.  I notice that I am not providing friends-data on command line, and that I was doing Monica apartment... will try ELR-apartment later to confirm.  Since it works on your end, I think we should merge here and file separate issue.

machenmusik:
Thanks @nikmo33 for working through issues!

nikmo33:
> Thanks @nikmo33 for working through issues!

I do not think we should be making separate logic changes as part of this same PR after the it has already gotten an approval. These defeats the purpose of CI, as these changes have not been reviewed by any of the approvers. Please open a separate PR for this kind of issue on the future, especially since it totally unrelated to the PR already approved. Thanks.

machenmusik:
Understood.  IMO, one challenge is that most approvers are on Linux, and so Windows coverage lags.

PR #1476, opened 2023-02-22
make alpha threshold configurable

hturki:
As a follow-up to https://github.com/nerfstudio-project/nerfstudio/issues/1099 (I've been trying to debug differences between nerfacc's iNGP performance and that of this repo as the former gives better PSNR results), I noticed that alpha_thre is currently hardcoded to 1e-2, which causes all samples to get rejected initially when training against the blender scenes. Setting it to 0 like in the nerfacc repo seems to fix the issue

### DIFF
--- a/nerfstudio/model_components/ray_samplers.py
+++ b/nerfstudio/model_components/ray_samplers.py
@@ -430,6 +430,7 @@ class VolumetricSampler(Sampler):
         near_plane: float = 0.0,
         far_plane: Optional[float] = None,
         cone_angle: float = 0.0,
+        alpha_thre: float = 1e-2,
     ) -> Tuple[RaySamples, TensorType["total_samples",]]:
         """Generate ray samples in a bounding box.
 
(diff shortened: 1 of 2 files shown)


nerfstudio/model_components/ray_samplers.py
```diff
@@ -469,14 +470,13 @@ def forward(
             t_max=t_max,
             scene_aabb=self.scene_aabb,
             grid=self.occupancy_grid,
-            # this is a workaround - using density causes crash and damage quality. should be fixed
```
tancik, line 472: Did you find this not to be an issue for real scenes?

hturki: if anything the existing iNGP code seems to do poorly as it stands - this is what I get after 30k iterations on the bww_entrance dataset:

image

(fwiw it seems to actually degrade over time, at 3k iterations it looks like this:)

image

(and then this at 6k iterations):

image


And this is what I get after this commit:

image

Lmk if there's a more rigorous test suite that I should be running

PR #1497, opened 2023-02-24
Rename train_split_percentage with train_split_fraction

### LINKED ISSUE #1411: Rename train_split_percentage to train_split_fraction?
Here:
https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/dataparsers/nerfstudio_dataparser.py#L63

The `train_split_percentage`parameter for the nerfstudio dataparser is actually a fraction (from 0 to 1), not a percentage (from 0 to 100), so the parameter should be named `train_split_fraction` (or simply `train_split`).

Help should read "The fraction (not percent) of images to use for training".

Or it could be done the other way round (percentage instead of fraction, the default being 90).

97littleleaf11:
Closes #1411


tancik (approved):
LGTM

PR #1499, opened 2023-02-25
Call load_state_dict on model classes

jkulhanek:
This PR enables models to handle load_state_dict calls. Therefore, model classes can easily remap keys and do some shape changes on the checkpoint tensors. Also, strict checkpoint validation is used for pipeline.model.

### DIFF
--- a/nerfstudio/pipelines/base_pipeline.py
+++ b/nerfstudio/pipelines/base_pipeline.py
@@ -21,7 +21,7 @@ import typing
 from abc import abstractmethod
 from dataclasses import dataclass, field
 from time import time
-from typing import Any, Dict, List, Optional, Type, Union, cast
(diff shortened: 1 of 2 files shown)


jkulhanek:
I know this is a tiny change, but this will make it easier to reimplement methods with dynamically-sized parameters (e.g. plenoxels or pointnerf). I am currently working on pointnerf reimplementation integrated into nerfstudio, but I think it would be better to keep it in a separate repo, as it requires some cuda extensions to be build. What do you think? I will also prepare an example repo demonstrating how to build a separate "extension" method implementation.

tancik (approved):
LGTM. I think you plan also sounds good. Implementing new models externally seems like a more scalable approach going forward, having an example will be helpful.

Issue #1438, opened 2023-02-16
Mip-Nerf-360 Data Converting


j93hahn:
Hi, has anyone been able to convert the Mip-NeRF-360 datasets used here https://jonbarron.info/mipnerf360/ into the specified NeRFStudio format, via the ns-process-data command? I seem unable to get the COLMAP operations to work and am wondering if anyone has had luck with that.

tancik:
Thats odd, it worked for us. Is it all of the datasets, or just one of them?

97littleleaf11:
It seems that the camera type of official datasets (generated by Colmap) hasn't been supported by nerfstudio. Using `ns-download-data` to get the supported data from nerfstudio team might work. Check this branch https://github.com/nerfstudio-project/nerfstudio/pull/1139 for more details.

jkulhanek:
However, if you use the original images from the dataset, they are downscaled in a different way. After downscaling the width and height are rounded up, which is different from nerfstudio, where it is assumed they are rounded down. You can use ns-process-data to downsample the images or manually modify `cameras/cameras.py:799` to round up instead.

j93hahn:
> It seems that the camera type of official datasets (generated by Colmap) hasn't been supported by nerfstudio. Using `ns-download-data` to get the supported data from nerfstudio team might work. Check this branch #1139 for more details.

Hi thank you for your response; which dataset is it listed under? When I run `ns-download-data -h`, I do not see the MipNerf360 dataset as an option.

j93hahn:
> Thats odd, it worked for us. Is it all of the datasets, or just one of them?

Yeah I am unable to get any of them to work -- I am running into many errors with the CUDA backend/binary file inconsistencies and cannot even get COLMAP to properly work. Would it be possible to share the datasets? I would like to experiment with all of them but am particularly interested in `bonsai`, `garden`, and `stump`.

j93hahn:
Okay thanks @tancik @97littleleaf11 @jkulhanek. I am using the datasets taken directly off of MipNeRF360 and am just going to convert the `poses_bounds.npy` files into the `transforms.json` files. Is there a script you ran to run this? All the datasets have already been processed by COLMAP so I don't think I need that

jkulhanek:
Can you try the `--skip-colmap` option in `ns-process-data` images?

jkulhanek:
I guess you can also add `--skip-process-images`

yyzzyy78:
[...]

Issue #1276, opened 2023-01-23
Not an issue, question about fisheye distortion.


KamiCalcium:
Hi @ethanweber , 

What was the formula you used to implement the fisheye ray direction model here? I've only seen theta = atan(r), where r = sqrt(x^2+y^2) model for fisheye before and cannot understand your code here. I will appreciate it if you explain it a bit or share pointers for the derivation. Thank you very much!
https://github.com/nerfstudio-project/nerfstudio/blob/3b8e2ce9ec4750b21b6b673bf62bedb61e1da2cf/nerfstudio/cameras/cameras.py#L658-L669

KamiCalcium:
Related to https://github.com/nerfstudio-project/nerfstudio/issues/868#issuecomment-1396326938 as well.

Hi @jake-austin, I make a new issue here.

qhdqhd:
same question...
Did you solve it?

KamiCalcium:
> same question... Did you solve it?

My two cents:

The terminology distort and undistort are somewhat confusing here. In opencv's [documentation](https://docs.opencv.org/3.4/db/d58/group__calib3d__fisheye.html): There are two steps for fisheye distortion: i) radial and tangential distortion, ii) applying fisheye effect. Let us not consider radial and tangential distortion for now, i.e., only consider fisheye effect. That said, all distortion coefficients k are 0, thus x'=(θ/r)a, and y'=(θ/r)b, where θ=atan(r), means we use the atan model for fisheye effect. Note that here the mapping from projected 2D coordinates (a, b) to distorted 2D projection coordinates (x', y') is called **distort**.  Now, we do a simply transform: tan(θ)=r, we then have a=x'tan(θ)/θ and b=y'tan(θ)/θ. This is called **undistort**: mapping (x', y') -> (a, b).

Now, in nerf training process, given a fisheye image (already distorted), we want to find the ray coordinates correspondence so that we need to do **undistort**. Check here: https://github.com/nerfstudio-project/nerfstudio/blob/3b8e2ce9ec4750b21b6b673bf62bedb61e1da2cf/nerfstudio/cameras/cameras.py#L667-L669, if everything divided by cos(theta), it becomes: 
```
 directions_stack[..., 0][mask] = torch.masked_select(coord_stack[..., 0] * tan_theta / theta, mask).float() 
 directions_stack[..., 1][mask] = torch.masked_select(coord_stack[..., 1] * tan_theta / theta, mask).float() 
[...]

PR #1500, opened 2023-02-25
Add support for more colmap camera models in ns-process-data

jkulhanek:
This PR adds support for more COLMAP camera models.
It addresses this issue: #1438

I was also thinking... perhaps I am missing something, however, isn't the FISHEYE distortion implementation wrong?
The code here: https://github.com/google/nerfies/blob/main/nerfies/camera.py doesn't match the COLMAP's implementation (https://github.com/colmap/colmap/blob/dev/src/base/camera_models.h). The implementation was taken from here: https://github.com/google/nerfies/blob/main/nerfies/camera.py, however, did the original code consider FISHEYE models?


tancik:
The multinerf code was used as reference. They extend the nerfies code to work with fisheye.https://github.com/google-research/multinerf/blob/main/internal/camera_utils.py
Perhaps it is also incorrect?

jkulhanek:
I was wrong. The math in multinerf is correct.

tancik (approved):
LGTM. Thanks for adding. Other colmap camera models have been requested a number of times, I'm glad to see them added!

jkulhanek:
This is related: #1276

tancik:
This appears to break processing when depth data isn't included. There is a PR to fix here - https://github.com/nerfstudio-project/nerfstudio/pull/1528

jkulhanek:
I don’t think so. It was definitely broken by PR #1531 (I mentioned it in the PR comments). In this PR I believe I fixed it for the ns-process-data images, but the same bug was also in other places apparently.

atonderski:
> I was wrong. The math in multinerf is correct.

@jkulhanek Would you mind quickly elaborating? I followed down the same rabithole and do not see how the implementations are equivalent. Thanks in advance!

jkulhanek:
The undistort solves the original distortion in normal space. Then try substituting https://github.com/google-research/multinerf/blob/5d4c82831a9b94a87efada2eee6a993d530c4226/internal/camera_utils.py#L585 to the original equations. It should lead to the radial equations if I remember correctly.

jkulhanek:
Consider undistort as a “black box” that solves the original problem. Then work backwards from the substitution

atonderski:
Thanks, appreciate it!

PR #1501, opened 2023-02-25
Fixes bug introduced in #1499 preventing load_from_checkpoint to work correctly 

jkulhanek:
I am very sorry to have introduced the bug in #1499 :( 
Please consider accepting this PR soon as it prevents people from restarting training from checkpoints.
This PR should fix the introduced bug and update the tests to cover it.

### DIFF
--- a/nerfstudio/pipelines/base_pipeline.py
+++ b/nerfstudio/pipelines/base_pipeline.py
@@ -21,7 +21,7 @@ import typing
 from abc import abstractmethod
 from dataclasses import dataclass, field
 from time import time
-from typing import Any, Dict, List, Optional, Type, Union, cast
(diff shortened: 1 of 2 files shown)


tancik (approved):
Lgtm

nerfstudio/pipelines/base_pipeline.py
```diff
@@ -372,10 +370,7 @@ def load_pipeline(self, loaded_state: Dict[str, Any], step: int) -> None:
         """
         state = {key.replace("module.", ""): value for key, value in loaded_state.items()}
```
jkulhanek, line 371: Can you please fix it here as well?

PR #1508, opened 2023-02-25
Extending ns-render to support loop rendering using interpolated cameras

ponimatkin:
This PR extends possibility of ns-render utility to render the data using already available camera poses. It is helpful when working with large datasets (ScanNet, ARKitScenes) as well as HPC clusters where computational nodes are not connected to the internet.

### DIFF
--- a/scripts/render.py
+++ b/scripts/render.py
@@ -252,13 +252,18 @@ def get_crop_from_json(camera_json: Dict[str, Any]) -> Optional[CropData]:
 
 @dataclass
 class RenderTrajectory:
-    """Load a checkpoint, render a trajectory, and save to a video file."""
+    """Load a checkpoint, render a trajectory, and save to a video file.
+    The following trajectory options are available,
+    filename: Load from trajectory created using viewer or blender vfx plugin.
+    interpolate: Create trajectory by interpolating between eval dataset images.
+    spiral: Create a spiral trajectory (can be hit or miss).
+    """
 
     load_config: Path
     """Path to config YAML file."""
     rendered_output_names: List[str] = field(default_factory=lambda: ["rgb"])
     """Name of the renderer outputs to use. rgb, depth, etc. concatenates them along y axis"""
-    traj: Literal["spiral", "filename", "loop"] = "spiral"
+    traj: Literal["spiral", "filename", "interpolate"] = "spiral"
     """Trajectory type to render. Select between spiral-shaped trajectory, trajectory loaded from
     a viewer-generated file and camera paths from the dataset interpolated and rendered in a loop."""
     downscale_factor: int = 1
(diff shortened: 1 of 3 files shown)


scripts/render.py
```diff
@@ -255,7 +259,7 @@ class RenderTrajectory:
     """Path to config YAML file."""
     rendered_output_names: List[str] = field(default_factory=lambda: ["rgb"])
     """Name of the renderer outputs to use. rgb, depth, etc. concatenates them along y axis"""
-    traj: Literal["spiral", "filename"] = "spiral"
```
tancik, line 258: Now that we are adding more options, can you add a description of each option in the class docstring.

tancik (commented):
Thanks for adding! As you probably guessed, we had this at some point a long time ago, but it never made it into the render.py script.

scripts/render.py
ponimatkin: I tried expanding docstring a bit in https://github.com/nerfstudio-project/nerfstudio/pull/1508/commits/cf01d605a1b67f6ea435d8cac37ff8a07f9376c9, do you think it is good or a bit verbose?

tancik: Sorry, I should have been more specific. I was thinking the main docstring could be updated to something like,

```
"""Load a checkpoint, render a trajectory, and save to a video file.
The following trajectory options are available,
filename: Load from trajectory created using viewer or blender vfx plugin.
interpolate: Create trajectory by interpolating between eval dataset images.
spiral: Create a spiral trajectory (can be hit or miss).
"""
```

scripts/render.py
```diff
@@ -255,8 +258,9 @@ class RenderTrajectory:
...
-    """Trajectory to render."""
+    traj: Literal["spiral", "filename", "loop"] = "spiral"
```
tancik, line 261: I was thinking a little more about the naming. Perhaps another term like `interpolate` would be more descriptive than `loop`

ponimatkin: Ah I see! Sorry for the confusion. :)

In this commit I've added this here https://github.com/ponimatkin/nerfstudio/commit/b8f7eb8a17ea35abec1c422db109d1eb3f088e2e. Then here https://github.com/ponimatkin/nerfstudio/commit/1df0ea87ab9fee0bd5570174e01ecd1a6b0e4c61 I've actually went further on and removed any mentions of the loops.

scripts/render.py
ponimatkin: Ah I see! Sorry for the confusion. :)

I've added your suggestions here b8f7eb8a17ea35abec1c422db109d1eb3f088e2e.

tancik (approved):
LGTM, thanks for adding this functionality.

PR #1509, opened 2023-02-25
feat(engine): improve type hints

SauravMaheshkar:
This PR aims to improve and add to the currently existing type hints in the `engine` submodule.

PR #1515, opened 2023-02-26
feat: improve type hints

SauravMaheshkar:
Following up with work done in #1509, this PR aims to further improve and add to the currently existing type hints.


tancik (approved):
LGTM. Thanks for cleaning!

PR #1525, opened 2023-02-27
Add weight-decay option to RAdam optimizer

jkulhanek:
This PR only adds the weight_decay option to RAdam optimizer.


tancik (approved):
LGTM

Issue #1530, opened 2023-02-28
RuntimeError: expected scalar type Float but found Double when render interpolate on latest version


cv-lab-x:
run error when render interpolate on latest version, but it works well when render spiral

`🎥 Rendering 🎥 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% ? --:--
Traceback (most recent call last):
  File "/miniconda3/envs/nerfstudio/bin/ns-render", line 8, in <module>
    sys.exit(entrypoint())
  File "/nerfstudio/scripts/render.py", line 341, in entrypoint
    tyro.cli(RenderTrajectory).main()
  File "/nerfstudio/scripts/render.py", line 325, in main
    _render_trajectory_video(
  File "///nerfstudio/scripts/render.py", line 115, in _render_trajectory_video
    outputs = pipeline.model.get_outputs_for_camera_ray_bundle(camera_ray_bundle)
  File "/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 27, in decorate_context
    return func(*args, **kwargs)
  File "//nerfstudio/nerfstudio/models/base_model.py", line 177, in get_outputs_for_camera_ray_bundle
    outputs = self.forward(ray_bundle=ray_bundle)
  File "//nerfstudio/nerfstudio/models/base_model.py", line 139, in forward
    return self.get_outputs(ray_bundle)
  File "//nerfstudio/nerfstudio/models/nerfacto.py", line 267, in get_outputs
    field_outputs = self.field(ray_samples, compute_normals=self.config.predict_normals)
  File "///miniconda3/envs/nerfstudio/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File //nerfstudio/nerfstudio/fields/base_field.py", line 111, in forward
    field_outputs = self.get_outputs(ray_samples, density_embedding=density_embedding)
  File "/nerfstudio/nerfstudio/fields/nerfacto_field.py", line 319, in get_outputs
    outputs[FieldHeadNames.PRED_NORMALS] = self.field_head_pred_normals(x)
  File "/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
[...]

PR #1532, opened 2023-02-28
Adding conversion to torch.float32 in get_interpolated_poses_many.

ponimatkin:
The issue #1530 could be potentially solved by this small patch.


tancik (approved):
LGTM

PR #1537, opened 2023-03-01
Fix crash when having no eval dataset and not using the viewer

iSach:
When using WandB or TensorBoard, Nerfstudio considers that evaluation is enabled and thus crashes if the evaluation dataset is empty.

I just added a check before performing an evaluation iteration to avoid that issue.


tancik (approved):
LGTM

PR #1541, opened 2023-03-01
MinimalDataParser add options, allow Nerfstudio data format to specify test indices

jkulhanek:
PR allows more options for the MinimalDataparser and allows NerfstudioDataparser to specify train/test indices in the json.


jkulhanek:
It also tracks the changes to the coord system done in the ns-process-data.

jkulhanek:
I am unsure if we should track the "applied_transform" done by ns-process data. This would allow us to get back the original COLMAP poses if needed.

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -182,22 +182,30 @@ def _generate_dataparser_outputs(self, split="train"):
...
-            indices = i_train
-        elif split in ["val", "test"]:
-            indices = i_eval
+        if f"{split}_indices" in meta:
+            # Datasets sets the train/test indices
+            indices = meta[f"{split}_indices"]
+            CONSOLE.log(f"[yellow] Dataset is overriding {split}_indices to {indices}")
+            indices = np.array(indices, dtype=np.int32)
+            if indices.max() >= len(poses):
+                raise RuntimeError(f"Invalid dataset indices: index {indices.max()} > {len(poses)} dataset images.")
```
tancik, line 191: Instead of storing indicies, can we store filenames?

jkulhanek:
I have also added some simple tests for nerfstudio_parser. It can be extended in future.

tancik (approved):
LGTM, thanks for making the changes to use filenames.

PR #1542, opened 2023-03-01
Added video cropping for equirectangular and made it independent of mask in normal videos

THE-COB:


nerfstudio/process_data/equirect_utils.py
```diff
@@ -33,17 +33,92 @@
 from nerfstudio.utils.rich_utils import ItersPerSecColumn
 
 
+def crop_bottom(bound_arr: list, fov: int, percent_crop: float) -> list:
```
tancik, line 36: Add `_` to the start of these crop functions, I don't think we should expect others to use them.

tancik, line 36: Use `List[float]` for the type.

nerfstudio/process_data/equirect_utils.py
```diff
@@ -33,17 +33,92 @@
...
+        fov (int): Field of view of the camera.
+        percent_crop (float): Percent of the image to crop from the top.
+
+    Returns:
+        list: A new list of bounds with the top cropped.
+    """
+    degrees_chopped = 180 * percent_crop
+    new_top_start = -90 + degrees_chopped + fov / 2
+    for i, el in enumerate(bound_arr):
+        print(i, el)
```
tancik, line 76: Remove print statement

tancik (commented):
Nice feature, thanks for adding!

tancik (approved):
LGTM!

PR #1549, opened 2023-03-02
Instant ngp args

ponimatkin:
This commit adds ability to parametrize Instant-NGP model with `log2-hashmap-size` and `max-res` and also computes `per_level_scale` variable https://github.com/nerfstudio-project/nerfstudio/blob/f3a2ea0c925c8956effdbe36ff70149d58f08025/nerfstudio/fields/instant_ngp_field.py#L100
dynamically as in `nerfacto` model.


tancik (approved):
LGTM

PR #1551, opened 2023-03-02
Avoid loading entire dataset on GPU

tancik:
If the dataset is large, or the GPU doesn't have a lot of memory, loading all of the data onto the GPU could cause OOM errors.


machenmusik:
May need to reopen mask / semantics issues; see https://github.com/nerfstudio-project/nerfstudio/commit/a91b92f89d401115b42e6ad295a8b850e02ee86f#r102847947

PR #1554, opened 2023-03-05
Add option to use Pixel Perfect SFM

pablovela5620:
Added the option to use https://github.com/cvg/pixel-perfect-sfm, when using hloc. Also added deps to docker since its a bit of a pain to install otherwise. I modified the docker-compose file as `dromni/nerfstudio:0.1.18` would need to be updated and I'm guessing that won't be till 0.1.19, I can change back to dromni image when that happens. I'm not sure if it makes a huge difference (with camera optimization enabled) but thought I may be worth a PR.

Let me know your thoughts


f-dy:
Thanks! that definitely looks like an interesting option!
What did you observe when looking at the results?
Camera optimization may not be able to fully recover from a noisy initialization and may even degrade the results (eg on the synthetic datasets), so I would also compare pixel-perfect without camera optimization to COLMAP with camera optimization.

tancik (approved):
LGTM, thanks for adding.

pablovela5620:
> Thanks! that definitely looks like an interesting option! What did you observe when looking at the results? Camera optimization may not be able to fully recover from a noisy initialization and may even degrade the results (eg on the synthetic datasets), so I would also compare pixel-perfect without camera optimization to COLMAP with camera optimization.

TBH I didn't find a significant difference, but maybe that's due to the dataset I was using. Would be good to see some more results from others

f-dy:
Did you try *without* camera pose optimization?

nkyriazis:
I can't figure out why the refine_pixsfm option is not exposed through cli (ns-process-data).

tancik:
Are you using the pip package, or the github version?

nkyriazis:
I'm running from the latest image on the docker hub.

pablovela5620:
it should be exposed? does it not show when using the `images` or `videos` argument? Here's what mine shows when I do `ns-process-data images --help`
Screenshot from 2023-03-19 13-56-34

pablovela5620:
> Did you try _without_ camera pose optimization?

I did not, this would be good to try, just have not had the chance as I'm working on bringing in SDF based methods

tancik:
Maybe the confusion is that it isn't exposed for `ns-process-data video`

devernay:
`ns-process-data video` should only extract images from video, is there a chance we could allow both `video` and `images` on the same line, with separate options? Or should `video` automatically inherit from all `images`options?

tancik:
I think `video` should essentially just inherit all of the image options.

nkyriazis:
> it should be exposed? does it not show when using the `images` or `videos` argument? Here's what mine shows when I do `ns-process-data images --help` Screenshot from 2023-03-19 13-56-34

image
image

tancik:
The latest pip package (used in docker) does not have this yet. You need to use the latest github version.

nkyriazis:
Could you identify for me what the missing link is? I see that several things are in place and I'm trying to figure out why my expectation that the parameter should be there is not met.

PR #1556, opened 2023-03-05
Fixes + Feature

iSach:
* Minor renaming
* Fixed the issue of gray-only results due to white background. This is most likely due to hitting a local minimum. 
* Added like in NerfPlayer an option to disable viewing dependency when using a neural decoder. Not sure about how to apply that for the linear decoder case.


iSach:
Fail, sorry (:

PR #1557, opened 2023-03-05
Correct instructions for Docker build

JRombouts:
This fixes the instruction for the cuda-architecture specific Docker build.


JRombouts:
@Zunhammer - since you were the last to touch this line, could you review?

Zunhammer:
You are right, either --tag or -t is the correct command. Thanks for fixing. The "-f Dockerfile" is optional though.

JRombouts:
You're welcome :)

I left the '-f Dockerfile' to be consistent with the rest of the examples, and I like its explicitness.

PR #1561, opened 2023-03-06
Added documentation for equirectangular input

THE-COB:


docs/quickstart/custom_dataset.md
```diff
@@ -290,3 +290,35 @@ ns-process-data metashape --data {data directory} --xml {xml file} --output-dir
...
+For videos we reccomend taking a video with the camera held on top of your head. This will result in any unwanted campurer to just be in the bottom of each frame image and therefore can be cropped out.
+
+For a video, process the data using the following command:
+
+```bash
+ns-process-data video --camera-type equirectangular --images-per-equirect {8, or 14} --num-frames-target {num equirectangular frames to sample from} --percent-crop {top bottom left right} --data {data directory} --output-dir {output directory}
+```
+See the equirectangular images section above for a description of the `--images-per-equirect` argument.
+
+The `num-frames-target` argument is optional but it is recomended to set it to 3*(seconds of video) frames. For example, if you have a 30 second video, you would use `--num-frames-target 90` (3*30=90).
```
akristoffersen, line 317: Is there a reason for 3 * seconds of video? from what i understand our usual target is 100-300 images.

THE-COB: It's kinda different for equirectangular for a few reasons. First, each equirectangular frame gets turned into 8 images all at the same location but at different angles so to capture the same amount of space as a normal video it would result in 8 times the number of frames. Second, this was just what I found out when working empirically. Do you think I should word this differently?

THE-COB: I added a line about how it's not exact

docs/quickstart/custom_dataset.md
```diff
@@ -290,3 +290,35 @@ ns-process-data metashape --data {data directory} --xml {xml file} --output-dir
...
+For a set of equirectangular images, process the data using the following command:
+
+```bash
+ns-process-data images --camera-type equirectangular --images-per-equirect {8, or 14} --data {data directory} --output-dir {output directory}
+```
+
+The images-per-equirect argument is the number of images that will be sampled from each equirectangular image. We have found that 8 images per equirectangular image is sufficient for most use cases so it defaults to that. However, if you find that there isn't enough detail in the nerf or that colmap is having trouble aligning the images, you can try increasing the number of images per equirectangular image to 14.
+
+### Videos
+For videos we reccomend taking a video with the camera held on top of your head. This will result in any unwanted campurer to just be in the bottom of each frame image and therefore can be cropped out.
```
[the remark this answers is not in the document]
THE-COB: fixed

akristoffersen (approved):

f-dy:
@THE-COB what about adding something like that to the doc https://docs.nerf.studio/en/latest/quickstart/custom_dataset.html#data

Equirectangular images coming from most 360 cameras will have issues, because it's physically impossible to build a 360 snapshot camera using sensors and lens with a single optical center. The stitched 360 images from a GoPro MAX or insta 360 or Ricoh Theta may look nice, but they will have parallax issues, especially if your scene contains objects that are close. So don't expect much quality from the resulting NeRF.

Probably a better way to use these images is to use the raw fisheye images rather than the stitched equirectangular images.

True panoramic devices use either a catadioptric system (no classical lens), or a rotating optical system (eg the classical DSLR-on-a-gimbal), so that all optical rays pass through the same optical center.

More info: https://cave.cs.columbia.edu/projects/categories/project?cid=Computational+Imaging&pid=Catadioptric+Cameras+for+360+Degree+Imaging

PR #1563, opened 2023-03-07
Added light Colab demo, omits COLMAP but notebook starts up faster.

chris838:
In case others find it useful - this change adds my streamlined colab notebook. I've excluded COLMAP and aggregated pip dependencies into a wheelhouse folder.

It's about 3-5 times faster to set up vs the current demo notebook, but it can't use arbitrary photos/videos without camera transforms, i.e. anything that needs to run COLMAP.

The zip dependency file should probably be moved to a more official GDrive account.


tancik:
Closing since the latest colab now uses the same trick.

PR #1567, opened 2023-03-08
bug fix for model loading from checkpoint

terrancewang:
fix for bug introduced in https://github.com/nerfstudio-project/nerfstudio/pull/1501

model state dict will load incorrectly if any parameters contain "_model" in their name. 

fix so that only the first instance of  "_model" is replaced


nerfstudio/pipelines/base_pipeline.py
```diff
@@ -108,7 +108,7 @@ def device(self):
 
     def load_state_dict(self, state_dict: Mapping[str, Any], strict: bool = True):
         model_state = {
-            key.replace("_model.", ""): value for key, value in state_dict.items() if key.startswith("_model.")
+            key.replace("_model.", "", 1): value for key, value in state_dict.items() if key.startswith("_model.")
```
jkulhanek, line 111: Wouldn’t it be better to just strip the prefix instead of running replace?

jkulhanek: Like `key[len(“_model.”):]`

jkulhanek (commented):
Can you please also fix it on line 373?

terrancewang:
> Can you please also fix it on line 373?

yeah, looks like the "module." should always be a prefix too i think

nerfstudio/pipelines/base_pipeline.py
```diff
@@ -370,7 +370,7 @@ def load_pipeline(self, loaded_state: Dict[str, Any], step: int) -> None:
             loaded_state: pre-trained model state dict
             step: training step of the loaded checkpoint
         """
-        state = {key.replace("module.", ""): value for key, value in loaded_state.items()}
+        state = {key[len("module."):]: value for key, value in loaded_state.items()}
```
jkulhanek, line 373: I don't think this will work. You can have keys without the "module." prefix, right?

terrancewang: you're right, just pushed a commit to fix this

jkulhanek:
Can you run `ns-dev-test`?

terrancewang:
> Can you run `ns-dev-test`?

yeah it passes the test vanilla pipeline test

jkulhanek:
But the linter fails, right? Just looking at the CI test results…

jkulhanek (approved):
Thanks for the bug fix.

PR #1568, opened 2023-03-08
Replace default colormap option with turbo for single-channel output

DreekFire:


tancik (commented):
Works great! Thanks for adding.

tancik (approved):

Issue #1571, opened 2023-03-08
PCL cannot load pointclouds written by nerfstudio


f-dy:
We get this when loading nerfstudio's ply with pcl (https://pointclouds.org/):
```
Failed to find match for field 'x'.
Failed to find match for field 'y'.
Failed to find match for field 'z'.
Failed to find match for field 'normal_x'.
Failed to find match for field 'normal_y'.
Failed to find match for field 'normal_z'.
Failed to find match for field 'curvature'.
```
The issue is also described here: https://github.com/isl-org/Open3D/issues/1633

[it looks like](https://github.com/isl-org/Open3D/issues/1633#issuecomment-1014225858) it's just a matter of using the new Open3D Tensor IO `o3d.t.io.write_point_cloud` instead of the legacy `o3d.io.write_point_cloud` [here](https://github.com/nerfstudio-project/nerfstudio/blob/a5eea5f14dabba87e80135b73a8cbf41b988106f/scripts/exporter.py#L274)

I will test this, but putting this as a reminder.

f-dy:
PR https://github.com/nerfstudio-project/nerfstudio/pull/1572 resolves this issue

PR #1572, opened 2023-03-09
scripts/exporter.py: export PLY compatible with PCL

f-dy:
Use the newer Open3D Tensor API to write pointclouds as float32, which are compatible with PCL (see https://github.com/isl-org/Open3D/issues/1633).

this also reduces the file size (almost by half)

fixes https://github.com/nerfstudio-project/nerfstudio/issues/1571

### DIFF
--- a/scripts/exporter.py
+++ b/scripts/exporter.py
@@ -2,6 +2,8 @@
 Script for exporting NeRF into other formats.
 """
 
+# pylint: disable=no-member
+
 from __future__ import annotations
 
 import json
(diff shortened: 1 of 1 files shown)


f-dy:
https://github.com/nerfstudio-project/nerfstudio/actions/runs/4377569585/jobs/7661246358#step:10:13

@tancik any idea how to fix this?

open3d.core.Dtype has been around [since open3d 0.11.0](http://www.open3d.org/docs/0.11.0/python_api/open3d.core.Dtype.html)

tancik (approved):
LGTM. Thanks for improving the compatibility!

PR #1575, opened 2023-03-09
Changed percent-crop to crop-factor so it's less confusing

THE-COB:
- changed argument
- updated docs
- Added value error if crop values are less than 0 or more than 1

### DIFF
--- a/scripts/process_data.py
+++ b/scripts/process_data.py
@@ -293,7 +293,7 @@ class ProcessVideo:
     percent_radius_crop: float = 1.0
     """Create circle crop mask. The radius is the percent of the image diagonal."""
     crop_factor: Tuple[float, float, float, float] = (0.0, 0.0, 0.0, 0.0)
-    """Portion of the image to crop. (top, bottom, left, right)"""
+    """Portion of the image to crop. All values should be in [0,1]. (top, bottom, left, right)"""
     use_sfm_depth: bool = False
     """If True, export and use depth maps induced from SfM points."""
     include_depth_debug: bool = False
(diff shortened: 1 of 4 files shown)


scripts/process_data.py
```diff
@@ -292,8 +292,8 @@ class ProcessVideo:
     """
     percent_radius_crop: float = 1.0
     """Create circle crop mask. The radius is the percent of the image diagonal."""
-    percent_crop: Tuple[float, float, float, float] = (0.0, 0.0, 0.0, 0.0)
-    """Percent of the image to crop. (top, bottom, left, right)"""
+    crop_factor: Tuple[float, float, float, float] = (0.0, 0.0, 0.0, 0.0)
+    """Portion of the image to crop. (top, bottom, left, right)"""
```
tancik, line 296: Add, `Value should be in [0,1]`

scripts/process_data.py
```diff
@@ -312,6 +312,10 @@ def main(self) -> None:  # pylint: disable=R0915
         image_dir = self.output_dir / "images"
         image_dir.mkdir(parents=True, exist_ok=True)
 
+        for i in self.crop_factor:
+            if i < 0 or i > 1:
+                raise ValueError("crop-factor values must be between 0 and 1")
+
```
tancik, line 318: Add this check inside the functions that actually do the crop. Also for consistency, instead of using `raise`, do something like, `CONSOLE.print("[red] Invalid crop...` followed by a `sys.exit(1)`

tancik (approved):
LGTM

PR #1578, opened 2023-03-09
Patch Sampling bug fix

akristoffersen:
Issue with dynamic batch size. Fixed with override to PixelSampler method.


tancik (approved):
LGTM

PR #1584, opened 2023-03-10
k-planes implementation

hturki:
This should be a pretty faithful implementation of the official K-Planes implementation at https://github.com/sarafridov/K-Planes. 

It can load models trained with the official implementation and gives comparable PSNR metrics when trained from scratch (31.69 vs 31.86 db in PSNR)

### DIFF
--- a/nerfstudio/fields/kplanes_field.py
+++ b/nerfstudio/fields/kplanes_field.py
@@ -42,8 +42,16 @@ except ImportError:
 CONSOLE = Console(width=120)
 
 
-def init_grid_param(out_dim: int, reso: List[int], a: float = 0.1, b: float = 0.5):
-    """Initializes the grid parameters."""
+def init_grid_param(out_dim: int, reso: List[int], a: float = 0.1, b: float = 0.5) -> nn.ParameterList:
+    """Initializes the grid parameters.
+    Args:
+        out_dim: Dimension of feature vectors stored in grid
+        reso: Grid resolution
+        a: the lower bound of the uniform distribution used for initialization of spatial planes
+        b: the upper bound of the uniform distribution used for initialization of spatial planes
+    Returns:
+        Grid parameters
+    """
 
     in_dim = len(reso)
     has_time_planes = in_dim == 4
@@ -61,8 +69,15 @@ def init_grid_param(out_dim: int, reso: List[int], a: float = 0.1, b: float = 0.
 
 
 @torch.jit.script
-def grid_sample_wrapper(grid: torch.Tensor, coords: torch.Tensor, align_corners: bool = True) -> torch.Tensor:
-    """Generates feature vectors by sampling and interpolating between grid values."""
+def grid_sample_wrapper(grid: torch.Tensor, coords: torch.Tensor) -> torch.Tensor:
+    """Generates feature vectors by sampling and interpolating between grid values.
+    Args:
+        grid: Input grid
+        coords: Coordinates to sample at
+
+    Returns:
+        Feature vectors
+    """
 
     grid_dim = coords.shape[-1]
 
@@ -84,7 +99,7 @@ def grid_sample_wrapper(grid: torch.Tensor, coords: torch.Tensor, align_corners:
     interp = F.grid_sample(
         grid,  # [B, feature_dim, reso, ...]
         coords,  # [B, 1, ..., n, grid_dim]
-        align_corners=align_corners,
+        align_corners=True,
         mode="bilinear",
         padding_mode="border",
     )
@@ -100,7 +115,14 @@ def interpolate_ms_features(
     ms_grids: Union[nn.ModuleList, List[nn.ParameterList]],
     concat_features: bool,
 ) -> torch.Tensor:
-    """Combines/interpolates features across multiple dimensions and scales."""
+    """Combines/interpolates features across multiple dimensions and scales.
+    Args:
+        pts: Coordinates to sample at
+        ms_grids: Grids to sample
+        concat_features: Whether to concatenate features at different scales
+    Returns:
+        Feature vectors
+    """
 
     coo_combs = list(itertools.combinations(range(pts.shape[-1]), 2))
 
@@ -128,16 +150,32 @@ def interpolate_ms_features(
 
 
 class KPlanesField(Field):
-    """K-Planes field."""
+    """K-Planes field.
+
+    Args:
+        aabb: Parameters of scene aabb bounds
+        num_images: How many images exist in the dataset
+        geo_feat_dim: Output geo feat dimensions
+        grid_base_resolution: Base grid resolution
+        grid_feature_dim: Dimension of feature vectors stored in grid
+        concat_features_across_scales: Whether to concatenate features at different scales
+        multiscale_res: Multiscale grid resolutions
+        spatial_distortion: Spatial distortion to apply to the scene
+        appearance_embedding_dim: Dimension of appearance embedding. Set to 0 to disable
+        use_average_appearance_embedding: Whether to use average appearance embedding or zeros for inference
+        linear_decoder: Whether to use a linear decoder instead of an MLP
+        linear_decoder_layers: Number of layers in linear decoder
+    """
 
     def __init__(
         self,
         aabb: TensorType,
         num_images: int,
         geo_feat_dim: int = 15,
-        grid_config: List[Dict] = field(default_factory=lambda: []),
         concat_features_across_scales: bool = True,
-        multiscale_res: List[int] = field(default_factory=lambda: [1, 2, 4, 8]),
+        grid_base_resolution: List[int] = field(default_factory=lambda: [128, 128, 128]),
+        grid_feature_dim: int = 32,
+        multiscale_res: List[int] = field(default_factory=lambda: [1, 2, 4]),
         spatial_distortion: Optional[SpatialDistortion] = None,
         appearance_embedding_dim: int = 0,
         use_average_appearance_embedding: bool = True,
@@ -150,7 +188,7 @@ class KPlanesField(Field):
         self.register_buffer("aabb", aabb)
         self.num_images = num_images
         self.geo_feat_dim = geo_feat_dim
-        self.grid_config = grid_config
+        self.grid_base_resolution = grid_base_resolution
         self.concat_features_across_scales = concat_features_across_scales
         self.spatial_distortion = spatial_distortion
 
@@ -159,10 +197,9 @@ class KPlanesField(Field):
         self.feature_dim = 0
         for res in multiscale_res:
             # initialize coordinate grid
-            config = grid_config[0].copy()
             # Resolution fix: multi-res only on spatial planes
-            config["resolution"] = [r * res for r in config["resolution"][:3]] + config["resolution"][3:]
-            gp = init_grid_param(out_dim=config["output_coordinate_dim"], reso=config["resolution"])
+            resolution = [r * res for r in grid_base_resolution[:3]] + grid_base_resolution[3:]
+            gp = init_grid_param(out_dim=grid_feature_dim, reso=resolution)
 
             # shape[1] is out-dim - Concatenate over feature len for each scale
             if self.concat_features_across_scales:
@@ -256,7 +293,7 @@ class KPlanesField(Field):
             # Input should be in [-1, 1]
             grid_input = (grid_input - self.aabb[0]) * (2.0 / (self.aabb[1] - self.aabb[0])) - 1.0
 
-        if len(self.grid_config[0]["resolution"]) == 4:
+        if len(self.grid_base_resolution) == 4:
             grid_input = torch.cat([grid_input, (2 * ray_samples.times - 1).reshape(-1, 1)], -1)
 
         features = interpolate_ms_features(
@@ -330,7 +367,14 @@ class KPlanesField(Field):
 
 
 class KPlanesDensityField(Field):
-    """A lightweight density field module."""
+    """A lightweight density field module.
+    Args:
+        aabb: Parameters of scene aabb bounds
+        resolution: Grid resolution
+        num_output_coords: dimension of grid feature vectors
+        spatial_distortion: Spatial distortion to apply to the scene
+        linear_decoder: Whether to use a linear decoder instead of an MLP
+    """
 
     def __init__(
         self,
--- a/nerfstudio/fields/kplanes_field.py
+++ b/nerfstudio/fields/kplanes_field.py
@@ -1,0 +1,388 @@
+# Copyright 2022 The Nerfstudio Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Fields for K-Planes (https://sarafridov.github.io/K-Planes/).
+"""
+
+import itertools
+from dataclasses import field
+from typing import Dict, List, Optional, Tuple, Union
+
+import torch
+import torch.nn.functional as F
+from rich.console import Console
+from torch import nn
+from torchtyping import TensorType
+
+from nerfstudio.cameras.rays import RaySamples
+from nerfstudio.field_components.activations import trunc_exp
+from nerfstudio.field_components.embedding import Embedding
+from nerfstudio.field_components.field_heads import FieldHeadNames
+from nerfstudio.field_components.spatial_distortions import SpatialDistortion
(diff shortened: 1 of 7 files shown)


hturki:
Update - the original difference in PSNR was due to a dataloader bug in the original implementation (https://github.com/sarafridov/K-Planes/issues/11). Both implementations now give comparable PSNR values and I've updated the PR description accordingly.

nerfstudio/data/dataparsers/blender_dataparser.py
```diff
@@ -88,7 +88,10 @@ def _generate_dataparser_outputs(self, split="train"):
 
         # in x,y,z order
         camera_to_world[..., 3] *= self.scale_factor
-        scene_box = SceneBox(aabb=torch.tensor([[-1.5, -1.5, -1.5], [1.5, 1.5, 1.5]], dtype=torch.float32))
+        radius = 1.3 if "ship" not in str(self.data) else 1.5
```
tancik, line 91: Remove special logic per scene.

tancik (commented):
Thanks for reimplementing k-planes.

I left some initial comments. One broader one is that I would like to avoid adding the alpha compositing in this PR. It's mostly tangential to k-planes and touches many base files.

nerfstudio/fields/kplanes_field.py
```diff
@@ -0,0 +1,409 @@
...
+try:
+    import tinycudann as tcnn
+except ImportError:
+    # tinycudann module doesn't exist
+    pass
+
+CONSOLE = Console(width=120)
+
+
+def init_grid_param(grid_nd: int, in_dim: int, out_dim: int, reso: List[int], a: float = 0.1, b: float = 0.5):
```
Giodiro, line 46: Here `grid_nd` is always 2, it would be cleaner to remove the configuration option and other references.

Giodiro: Also `in_dim` can be easily inferred from `reso` (`in_dim = len(reso)`), best to remove it from the arguments.

nerfstudio/fields/kplanes_field.py
```diff
@@ -0,0 +1,409 @@
...
+    interp = interp.view(B, feature_dim, n).transpose(-1, -2)  # [B, n, feature_dim]
+    interp = interp.squeeze()  # [B?, n, feature_dim?]
+
+    return interp
+
+
+def interpolate_ms_features(
+    pts: torch.Tensor,
+    ms_grids: Union[nn.ModuleList, List[nn.ParameterList]],
+    grid_dimensions: int,
```
Giodiro, line 103: Similarly `grid_dimensions` is always 2

nerfstudio/fields/kplanes_field.py
```diff
@@ -0,0 +1,409 @@
...
+        grid_input = ray_samples.frustums.get_positions().view(-1, 3)
+
+        if self.spatial_distortion is not None:
+            grid_input = self.spatial_distortion(grid_input)
+            grid_input = grid_input / 2  # from [-2, 2] to [-1, 1]
+        else:
+            # Input should be in [-1, 1]
+            grid_input = (grid_input - self.aabb[0]) * (2.0 / (self.aabb[1] - self.aabb[0])) - 1.0
+
+        if self.grid_config[0]["input_coordinate_dim"] == 4:
```
Giodiro, line 271: Time needs to be normalized: in nerfstudio datasets it's in [0, 1] but for interpolation it must be in [-1, 1] (do `timestamps = (timestamps * 2) - 1`)

nerfstudio/fields/kplanes_field.py
```diff
@@ -0,0 +1,409 @@
...
+        grid_input = ray_samples.frustums.get_positions().view(-1, 3)
+
+        if self.spatial_distortion is not None:
+            grid_input = self.spatial_distortion(grid_input)
+            grid_input = grid_input / 2
+        else:
+            grid_input = (grid_input - self.aabb[0]) * (2.0 / (self.aabb[1] - self.aabb[0])) - 1.0
+
+        if ray_samples.times is not None and self.hexplane:
+            grid_input = torch.cat([grid_input, ray_samples.times.view(-1, 1)], -1)
```
Giodiro, line 398: As above, time needs to be normalized

nerfstudio/models/kplanes.py
```diff
@@ -0,0 +1,396 @@
...
+    _target: Type = field(default_factory=lambda: KPlanesModel)
+
+    near_plane: float = 2.0
+    """How far along the ray to start sampling."""
+    far_plane: float = 6.0
+    """How far along the ray to stop sampling."""
+
+    grid_config: List[Dict] = field(
+        default_factory=lambda: [
+            {"grid_dimensions": 2, "input_coordinate_dim": 3, "output_coordinate_dim": 32, "resolution": [64, 64, 64]}
```
Giodiro, line 73: here grid_dimensions is always 2, I would delete it.

nerfstudio/models/kplanes.py
```diff
@@ -0,0 +1,396 @@
...
+        ]
+    )
+
+    is_ndc: bool = False
```
Giodiro, line 77: Never used

nerfstudio/models/kplanes.py
```diff
@@ -0,0 +1,396 @@
...
+    is_contracted: bool = False
+    multiscale_res: List[int] = field(default_factory=lambda: [1, 2, 4, 8])
+    concat_features_across_scales: bool = True
+    linear_decoder: bool = False
+    linear_decoder_layers: Optional[int] = 4
```
Giodiro, line 82: default to 1 is better

nerfstudio/models/kplanes.py
```diff
@@ -0,0 +1,396 @@
...
+            self.density_fns.extend([network.density_fn for network in self.proposal_networks])
+
+        # Samplers
+        update_schedule = lambda step: np.clip(
+            np.interp(step, [0, self.config.proposal_warmup], [0, self.config.proposal_update_every]),
+            1,
+            self.config.proposal_update_every,
+        )
+
+        if self.config.is_contracted or self.config.is_ndc:
```
Giodiro, line 180: Here there's no point in using the `is_ndc` config as nerfstudio doesn't have ndc datasets

nerfstudio/models/kplanes.py
```diff
@@ -0,0 +1,396 @@
...
+
+        return outputs
+
+    def get_metrics_dict(self, outputs, batch):
+        image = self._get_image(batch, outputs)
+
+        metrics_dict = {}
+        metrics_dict["psnr"] = self.psnr(outputs["rgb"], image)
+        if self.training:
+            metrics_dict["plane_tv"] = compute_grid_tv(self.field.grids)
```
Giodiro, line 296: Maybe these should be computed in the `get_loss_dict` function directly (they're not really metrics like PSNR is).

Giodiro (commented):
Hi @hturki , Thanks for the PR! I left a few minor comments on places which can be cleaned up; as well as the normalization of time which is important.
I'll try to add a commit on top for the missing regularizers!

Giodiro:
To add to @iSach 's comment, you can have a look [here](https://github.com/akristoffersen/nerfstudio_kplanes/pull/5) for his implementation of the regularizers!

iSach:
> Is there a reason you left behind the temporal losses? (sparse transients, temporal smoothness)
> 
> Also, I'm not sure I understand the code for the plane TV loss. In the paper, they apply it only on space-only planes and in 1D on space-time planes. However, the implementation here computes in 2D two times over space-only planes, and one time over space-time planes?
> 
> ```python
>     total = 0.0
>     for grids in multi_res_grids:
>         if len(grids) == 3:
>             spatial_grids = [0, 1, 2]
>         else:
>             spatial_grids = [0, 1, 3]  # These are the spatial grids; the others are spatiotemporal
>         for grid_id in spatial_grids:
>             total += compute_plane_tv(grids[grid_id])
>         for grid in grids:
>             total += compute_plane_tv(grid)
> ```

Actually, I don't even understand exactly what is going on in the "compute_plane_tv" function.

```
    batch_size, c, h, w = t.shape
    count_h = batch_size * c * (h - 1) * w
    count_w = batch_size * c * h * (w - 1)
    h_tv = torch.square(t[..., 1:, :] - t[..., : h - 1, :]).sum()
    w_tv = torch.square(t[..., :, 1:] - t[..., :, : w - 1]).sum()
    return 2 * (h_tv / count_h + w_tv / count_w)  # This is summing over batch and c instead of avg
```

* batch_size is always 1. Why is there this dimension? It's not the batch size, it's number of models.
* the sum() then dividing by count_h (or w) is exactly the same here as doing .mean() on h_tv (or w_tv). What is meant here by summing over batch and c instead of avg?
* Why is the result multiplied by 2?

If anybody understands this better, I'd like to know about it.

nerfstudio/configs/base_config.py
```diff
@@ -28,6 +28,7 @@
 
 warnings.filterwarnings("ignore", module="torchvision")
 
+
```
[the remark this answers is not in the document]
hturki: Will do, but FYI I get these sort of extraneous diffs after running "black docs/ nerfstudio/ scripts/ tests/." I'll run it selectively only on the files that I touch but it would be nice if one could just run that one command and not have to check file-by-file in the future

nerfstudio/models/kplanes.py
hturki: I personally like being able to look at these values before they get scaled with loss coefficients, but I don't have very strong opinions if others disagree

nerfstudio/configs/base_config.py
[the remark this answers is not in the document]
tancik: Running `ns-dev-test` should run black, isort, lint, tests

hturki:
Higher-level comments:
- Will split out the alpha compositing into its own PR
- Re: some of the TV-related code - I mainly copied code from the official repo at https://github.com/sarafridov/K-Planes to make sure that the port was faithful / that the numbers line up with the paper. I'd refer any questions to the official implementation, but happy to simplify things / cut out parts that aren't relevant to nerfstudio
- Re: not adding time-related losses - I was hoping to start with that's needed for static scenes in this PR (which I should have mentioned in the PR description), as the paper mentions doing things like importance sampling for the DyNeRF scenes that would likely involve making more changes than what I wanted to do with this initial PR. I'm happy to add the time losses in this PR, but I (or someone else) would ideally do any other required things (like the importance sampling) in a follow-up to this

Giodiro:
> Higher-level comments:
> 
>     * Will split out the alpha compositing into its own PR
> 
>     * Re: some of the TV-related code - I mainly copied code from the official repo at https://github.com/sarafridov/K-Planes to make sure that the port was faithful / that the numbers line up with the paper. I'd refer any questions to the official implementation, but happy to simplify things / cut out parts that aren't relevant to nerfstudio
> 
>     * Re: not adding time-related losses - I was hoping to start with that's needed for static scenes in this PR (which I should have mentioned in the PR description), as the paper mentions doing things like importance sampling for the DyNeRF scenes that would likely involve making more changes than what I wanted to do with this initial PR. I'm happy to add the time losses in this PR, but I (or someone else) would ideally do any other required things (like the importance sampling) in a follow-up to this

Regarding time: the PR already seems to support time-datasets, you can test that with dycheck or dnerf which are in nerfstudio. importance sampling is definitely out of scope though!

iSach:
I indeed think importance sampling is out of scope, but it can be a very important convergence factor.

I have a large dataset (10500 images) with a large scale scene with small dynamic objects, and it made qualitative results incredibly better by implementing (some sort of) importance sampling.

However, my implementation changes more and bigger parts of Nerfstudio (and is probably not the cleanest), so I'll probably do it separately.

aaryapatel007 (approved):

nerfstudio/configs/base_config.py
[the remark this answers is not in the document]
hturki: my point is that running ns-dev-test causes changes in files that I'm not touching. For example, I see diffs such as this when running it on a fresh checkout of main:

image

Furthermore, running pylint (on a fresh checkout) seems to fail:

image

Is there possibly something I'm doing wrong here?

nerfstudio/engine/schedulers.py
```diff
@@ -107,3 +108,37 @@ def func(step):
 
         scheduler = lr_scheduler.LambdaLR(optimizer, lr_lambda=func)
         return scheduler
+
+
+@dataclass
+class CosineWithWarmupSchedulerConfig(SchedulerConfig):
```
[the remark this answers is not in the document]
tancik: Have you run `pip install -e .[dev]`?

nerfstudio/configs/base_config.py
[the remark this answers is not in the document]
tancik: Have you run `pip install -e .[dev]`?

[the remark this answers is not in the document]
hturki: hmm, I must have had the wrong versions of the libraries installed or something. things work now after reinstalling - sorry for the noise

aaryapatel007 (approved):

hturki:
@tancik I think I've addressed most of the feedback in this PR, but I'm now getting a test failure (https://github.com/nerfstudio-project/nerfstudio/actions/runs/4423210082/jobs/7755716203) - do you know what might be going on?

aaryapatel007 (approved):

nerfstudio/fields/kplanes_field.py
```diff
@@ -0,0 +1,388 @@
...
+    import tinycudann as tcnn
+except ImportError:
+    # tinycudann module doesn't exist
+    pass
+
+CONSOLE = Console(width=120)
+
+
+def init_grid_param(out_dim: int, reso: List[int], a: float = 0.1, b: float = 0.5):
+    """Initializes the grid parameters."""
```
tancik, line 46: For all docstrings can you follow the form,

```
""" Desc

Args:
      a: stuff
      
Returns:
      desc
"""
```

nerfstudio/models/kplanes.py
```diff
@@ -0,0 +1,420 @@
...
+    """K-Planes Model Config"""
+
+    _target: Type = field(default_factory=lambda: KPlanesModel)
+
+    near_plane: float = 2.0
+    """How far along the ray to start sampling."""
+    far_plane: float = 6.0
+    """How far along the ray to stop sampling."""
+
+    grid_config: List[Dict] = field(
```
tancik, line 70: Add argstrings to all arguments

tancik (commented):
Thoughts on moving the k-planes encoding out of the field and into `encodings.py` similar to what we did with tensorf. This will make it easier for others to use these encodings in their projects.

hturki:
@tancik I think we could move the k-planes encoding to ```encodings.py```, but there seems to be a ```TriplaneEncoding``` class there already for TensorRF that behaves differently from the K-Planes one and I'm not sure it would be straightforward to cleanly merge in the functionality needed to support both into that single encoding. Should I create a separate ```KPlanesEncoding``` or do you have a better suggestion?

tancik:
> @tancik I think we could move the k-planes encoding to `encodings.py`, but there seems to be a `TriplaneEncoding` class there already for TensorRF that behaves differently from the K-Planes one and I'm not sure it would be straightforward to cleanly merge in the functionality needed to support both methods into that single encoding. Should I create a separate `KPlanesEncoding` or do you have a better suggestion?

Creating a `KPlanesEncoding` sounds good to me.

aaryapatel007 (approved):

akristoffersen:
Is there something that's blocking this PR? Happy to get the branch updated / get through the final cleanup if no one else is working on it.

Giodiro:
If we can merge the encoding first (https://github.com/nerfstudio-project/nerfstudio/pull/1658) and use it from this PR it should be good to go?

hturki:
hey all, I've been dealing with a paper-related deadline but getting this over the line should be straightforward and I'll take care of it today

hturki:
hey @tancik - are there any todos that you'd still like for me to address in this PR?

tancik:
Sorry for the delay, Im traveling at the moment, won't have a chance to do a thorough look for a few days.

hturki:
no rush on my side!

tancik:
Sorry for taking so long to get back to this. Since my last comment we have started rethinking how we organize new models. Would you be willing to host the k-planes repo on your own github? We have a guide for setting it up [here](https://docs.nerf.studio/en/latest/developer_guides/new_methods.html), given this PR, it should slot in nicely. We can then add a page to methods so that others can easily find it. For example, we just did this for [instruct-nerf2nerf](https://docs.nerf.studio/en/latest/nerfology/methods/in2n.html)

hturki:
sgtm, this will realistically happen near the end of may at the earliest but will let you know once I give it a go

Giodiro:
Hi! K-planes integration is live at https://github.com/Giodiro/kplanes_nerfstudio
I assumed the k-planes embedding will still live in nerfstudio, but if you want to change that let me know!

I'll write a PR for the docs soon, and I think we can close this one!

tancik:
> Hi! K-planes integration is live at https://github.com/Giodiro/kplanes_nerfstudio I assumed the k-planes embedding will still live in nerfstudio, but if you want to change that let me know!
> 
> I'll write a PR for the docs soon, and I think we can close this one!

This is great, thanks for your effort! We should keep the embedding in the main repo. Looking forward to the docs PR. Will close this PR.

Issue #1581, opened 2023-03-10
IndexError: index 3 is out of bounds for axis 0 with size 3


hugosbnarciso:
**Describe the bug**
```
[13:26:20] 🎉 Done refining intrinsics.                                                              colmap_utils.py:182
Traceback (most recent call last):
  File "D:\Anaconda3\envs\nerfstudio\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "D:\Anaconda3\envs\nerfstudio\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "D:\Anaconda3\envs\nerfstudio\Scripts\ns-process-data.exe\__main__.py", line 7, in <module>
  File "D:\nerfstudio\scripts\process_data.py", line 970, in entrypoint
    tyro.cli(Commands).main()
  File "D:\nerfstudio\scripts\process_data.py", line 434, in main
    num_matched_frames = colmap_utils.colmap_to_json(
  File "D:\nerfstudio\nerfstudio\process_data\colmap_utils.py", line 459, in colmap_to_json
    applied_transform = applied_transform[np.array([1, 0, 2, 3]), :]
IndexError: index 3 is out of bounds for axis 0 with size 3
```

**To Reproduce**
`ns-process-data video --data D:\nerfstudio\data\video\in\video.mp4 --output-dir D:\nerfstudio\data\video\out --matching-method sequential --verbose`

**Expected behavior**
It should complete the COLMAP script and create the transforms.json , it does all the resizing though

**Additional context**
Yesterday it worked, after i did a git pull / pip install e . this error appears, not sure it's related though

lsycool:
+1

hugosbnarciso:
i believe this commit is the culprit: https://github.com/nerfstudio-project/nerfstudio/commit/c61d3e9387425e568ee7317a14d517c4e5a8bdf0 , as it worked when i reverted back

PR #1586, opened 2023-03-11
Fixed: MinimalDataParser add options, allow Nerfstudio data format to specify test indices

jkulhanek:
This PR introduces #1541 back.
Fixes: #1581
Revert was here: #1583


nerfstudio/process_data/colmap_utils.py
```diff
@@ -455,6 +455,11 @@ def colmap_to_json(
     out = parse_colmap_camera_params(cam_id_to_camera[1])
     out["frames"] = frames
 
+    applied_transform = np.eye(4)[:3, :]
+    applied_transform = applied_transform[np.array([1, 0, 2]), :]
```
tancik, line 459: Just confirming, this is the only change from the last PR?

jkulhanek: I also changed the tests because the interface of NerfstudioDataParserConfig changed. But otherwise, yes.

tancik (approved):
LGTM

PR #1587, opened 2023-03-11
Updated Colab to Python3.9 and added colmap install speedup.

akristoffersen:
Thanks to azua_23423333 on Discord for sharing the colmap installation speed-up!


tancik (approved):
Lgtm

PR #1594, opened 2023-03-12
Adds util function to convert transformed poses back + tests

jkulhanek:
This PR adds a util function to convert poses back from transformed coordinates. It also improves tests for ns-process-data images.

Please check if the naming and overall conventions are ok and if placing the function to base_dataparser is fine.


tancik (approved):
LGTM. This is a useful function to have. Function names look good.

nickponline:
@jkulhanek How do you get  the pointcloud back into the same coordinate system as the cameras though?

jkulhanek:
Just multiply by the dataparser_transform and then by dataparser_scale.

nickponline:
@jkulhanek Hmmm I did try and and the inverse but I think there is something else. For example here are the camera poses I'm putting through nerf and the pointcloud (generated using MVS). 

image

Then I've tried:

Cameras poses multiplied by dataparser_transform and dataparser_scale
image

Pointcloud multiplied by inverse dataparser_transform and divided by dataparser_scale
image

Neither seem to have points in same coordinates as cameras. 

I notice that the input poses are additionally transformed like this:
```
def change_axis(pose):
    ret = copy.deepcopy(pose)
    ret[0:3,2] *= -1 
    ret[0:3,1] *= -1 
    ret = ret[[1,0,2,3],:] 
    ret[2,:] *= -1 
    return ret
```

Is this something that needs to be undone on the pointcloud too?

jkulhanek:
Which version of nerfstudio are you using? Which dataparser and how did you generate the transforms.json?

When transforming your externally generated pc, I think you want to multiply by scale and transforms, not divide. 

Can you try to apply the function implemented in this PR to get the original poses from the transformed ones?

The snippet you sent does the following:
First two lines rotate the world coord system. This I believe is useless, but it is used in the ns-process-data.
Last two lines switch between opencv and opengl camera space conventions. This depends on how you use the resulting proj matrices. The function implemented in this PR should allow you to specify the camera convention, but it also requires you to generate the data with the newest ns-process-data, such that the world coord rotation is included in the dataparser-transform.

nickponline:
Thank you! I got it working, needed to take the pointcloud and divide by `dataparser_scale` then multiply by `np.linalg.inv(dataparser_transform`. Which I think it what you suggested. There was an additional offset but that was because I had centered the camera poses when I generated `transforms.json`

jkulhanek:
Good!

Yaodizhao:
> 谢谢你！我让它工作了，需要获取点云并除以`dataparser_scale`然后乘以`np.linalg.inv(dataparser_transform`。我认为这就是你的建议。有一个额外的偏移，但那是因为我在生成时已将相机姿势居中`transforms.json`
May I ask how to convert the generated point cloud back to the original data? I calculated it but there was a significant difference. May I ask if you can give me some help

PR #1619, opened 2023-03-16
Use the run timestamp as run name on WandB

iSach:
When running various experiments, using the "experiment name" is quite useful for tracking with Weights and Biases, but it quickly becomes a mess with the random names.

This small PR sets the name of each run to the timestamp of the run to more easily identify them from the website.

This way, one can easily distinguish between runs within an experiment by using custom timestamps.


tancik (approved):
LGTM

PR #1620, opened 2023-03-16
add sdf mesh export and texturing

pablovela5620:
Added export for sdf based methods (Currently just NeuS). Tested on a trained Replica scene, this is what the current output looks like. 

https://user-images.githubusercontent.com/25287427/225754201-a33f93d2-a1ea-4031-8edb-dbb254e939df.mov

the command I used to export was
```
ns-export sdf-marching-cubes --load-config outputs/unnamed/neus/2023-03-13_170828/config.yml --output-dir test-mesh/
```
Let me know if class names make sense (It uses marching cubes but this would be different for density/color-based models like nerfacto v.s. sdf models like neus). Once this is merged in I'm planning on porting over the `neus-facto` model as its much faster to train thanks to the proposal mlp.

I also haven't had a chance to test with other scenes since it takes a few hours to train the model and I've just been focusing on the one specific replica scene ( would be good to check with DTU)


ethanweber:
Hey @pablovela5620, cool work. Can you share the commands that you used to process data and train the model as well? I'm looking through the code now. I see that you use the `ns-train neus` command. Does this user depth and normal supervision? It looks like `include_mono_prior` is False, so we could probably just use the NerfstudioDataParser directly and then we can use neus with any dataset that we've processed with our `ns-process-data` command?

Also I see you set max_num_iterations to 100000 (100K). How long does it take for your scene to converge? Just curious how long I should wait until I expect to see something reasonable in the viewer.

Thanks! Excited to help get sdf method integrated.

pablovela5620:
Sure,
First to download the relevant data I used
```
ns-download-data sdfstudio --dataset-name sdfstudio-demo-data
```
the command I used to train the above scene was 
```
ns-train neus --pipeline.model.sdf-field.inside-outside True --pipeline.model.mono-normal-loss-mult 0.05  --pipeline.model.mono-depth-loss-mult 0.1--data data/sdfstduio-demo-data/replica-room0 --include-mono-prior True
```
monocular priors (from the omnidata repo) are very helpful when reconstructing indoor scenes but they aren't required (I did use for the above seen. Currently I haven't included a way to generate the mono priors in nerfstudio and they are just included in the sdfdataset download. I plan on bringing this functionality over after I had neus and neus-facto working (with mesh extraction). You could also try an example without the monocular priors with the following command 

```
ns-train neus --pipeline.model.sdf-field.inside-outside False --data data/sdfstduio-demo-data/dtu-scan65
```

here is what this looks like after 100k iterations (the untextured output from the sdf and then the textured output)


https://user-images.githubusercontent.com/25287427/226199138-05d0389a-46a8-4b0e-9209-c03efcab25ab.mp4


as for the number of iterations, for the dtu example you should have a semi reasonable convergence after ~20k iterations, at 100k iterations using NeuS it takes ~4 hours (an an A6000) so around an hour for the 20k. NeuS-facto uses the proposal network sampler and makes things much faster (20k iterations ~15min), my plan will be to bring that in next.

Lastly for using `ns-process-data` I'll add that in after NeuS-facto as currently its still much to slow using the original NeuS method. That way like you said we can take advantage of the Nerfstudio parser and all the capture methods like colmap/polycam/etc...

Hope that clears everything up!

pablovela5620:
Here's an example of a room I scanned using just a video taken from my IPhone, (This doesn't include the texturing portion, only the marching cubes). I trained this using NeuS-Facto for 50k iterations ~(1 hour?) of training which I have a PR for #1654. I'll do some more digging into how to get it working like this for objects v.s. rooms

https://user-images.githubusercontent.com/25287427/228642260-f28f8458-a47f-47ed-ac78-3609373795a4.mov

ethanweber:
> Here's an example of a room I scanned using just a video taken from my IPhone, (This doesn't include the texturing portion, only the marching cubes). I trained this using NeuS-Facto for 50k iterations ~(1 hour?) of training which I have a PR for #1654. I'll do some more digging into how to get it working like this for objects v.s. rooms

This looks really good. Does it use monocular priors?

nerfstudio/exporter/marching_cubes.py
```diff
@@ -0,0 +1,249 @@
...
+        level: The isovalue at which to approximate the isosurface.
+        coarse_mask: A binary mask tensor of size ("height", "width", "depth") that indicates the regions
+            of the bounding box where the SDF is expected to have a zero-crossing. If
+            provided, the algorithm first evaluates the SDF at the coarse voxels where
+            the mask is True, and then refines the evaluation within these voxels using
+            a multi-scale approach. If None, evaluates the SDF at all points in the
+            bounding box.
+    """
+    # Check if resolution is divisible by 512
+    assert resolution % 512 == 0
```
ethanweber, line 169: Can you throw an error that is descriptive for why "divisible by 512" is important? I see that it's hard-coded below this.

scripts/exporter.py
```diff
@@ -323,6 +324,62 @@ def main(self) -> None:
...
+    def main(self) -> None:
+        """Main function."""
+        if not self.output_dir.exists():
+            self.output_dir.mkdir(parents=True)
+
+        _, pipeline, _ = eval_setup(self.load_config)
+
+        CONSOLE.print("Extract mesh with marching cubes and may take a while")
+
+        assert self.resolution % 512 == 0
```
ethanweber, line 357: Similar comment to before, can you raise some descriptive error?

scripts/exporter.py
```diff
@@ -323,6 +324,62 @@ def main(self) -> None:
...
+    """Target number of faces for the mesh to texture."""
+
+    def main(self) -> None:
+        """Main function."""
+        if not self.output_dir.exists():
+            self.output_dir.mkdir(parents=True)
+
+        _, pipeline, _ = eval_setup(self.load_config)
+
+        CONSOLE.print("Extract mesh with marching cubes and may take a while")
```
ethanweber, line 355: rewrite? "Extracting mesh with marching cubes, which may take a while"

scripts/exporter.py
```diff
@@ -323,6 +324,62 @@ def main(self) -> None:
...
+
+        assert self.resolution % 512 == 0
+        # for sdf we can multi-scale extraction.
+        get_surface_sliding(
```
ethanweber, line 359: Can we rename this function? "get" implies you are returning something, but instead you are writing to the `output_path`. Maybe `write_mesh_with_marching_cubes`? You can decide, but something with writing instead of "get"ing would be nicer IMO.

scripts/exporter.py
```diff
@@ -323,6 +324,62 @@ def main(self) -> None:
...
+            bounding_box_min=self.bounding_box_min,
+            bounding_box_max=self.bounding_box_max,
+            coarse_mask=None,
+        )
+
+        # load the mesh from the marching cubes export
+        mesh = get_mesh_from_filename(
+            str(self.output_dir / "sdf_marching_cubes_mesh.ply"), target_num_faces=self.target_num_faces
+        )
+        CONSOLE.print("Texturing mesh with NeRF")
```
ethanweber, line 372: Nice, this one is written with the active "ing" tense

ethanweber (commented):
A few comments and then it looks good to me! Thanks for being patient on this PR.

scripts/exporter.py
```diff
@@ -323,6 +324,62 @@ def main(self) -> None:
         raise NotImplementedError("Marching cubes not implemented yet.")
 
 
+@dataclass
+class ExportSDFMarchingCubesMesh(Exporter):
```
ethanweber, line 328: Can we make this generic and not specific to an SDF? I think it would be better for nerfstudio. We can specify the isosurface by a threshold that is set as a variable in this function as opposed to assuming it is zero. This way we can do marching cubes on NeRF densities too.

scripts/exporter.py
```diff
@@ -323,6 +324,62 @@ def main(self) -> None:
...
+        assert self.resolution % 512 == 0
+        # for sdf we can multi-scale extraction.
+        get_surface_sliding(
+            sdf=lambda x: pipeline.model.field.forward_geonetwork(x)[:, 0].contiguous(),
+            output_path=self.output_dir,
+            resolution=self.resolution,
+            bounding_box_min=self.bounding_box_min,
+            bounding_box_max=self.bounding_box_max,
+            coarse_mask=None,
+        )
```
ethanweber, line 366: This is where you'd pass in the isosurface threshold. You could set it as zero which would be the default for the SDF case.

nerfstudio/exporter/marching_cubes.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+@torch.no_grad()
+def get_surface_sliding(
+    sdf: Callable,
+    output_path: Path,
+    resolution: int = 512,
+    bounding_box_min: Tuple[float, float, float] = (-1.0, -1.0, -1.0),
+    bounding_box_max: Tuple[float, float, float] = (1.0, 1.0, 1.0),
+    level: float = 0,
```
ethanweber, line 140: Here is the "isosurface threshold" I'm referring to in the other comments from exporter.py. We can use this with NeRFs as well if we pass in a value other than zero.

nerfstudio/exporter/marching_cubes.py
```diff
@@ -0,0 +1,249 @@
...
+            pts_sdf = upsample(pts_sdf)
+            pts_sdf = pts_sdf.reshape(-1)
+
+        threshold /= 2.0
+    return pts_sdf
+
+
+@torch.no_grad()
+def get_surface_sliding(
+    sdf: Callable,
```
ethanweber, line 135: Regarding my comments on "isosurface threshold", `sdf` can be renamed to some name which is a callable function. It'll be sdf for an sdf-based method but densities for a nerf-based method.

scripts/exporter.py
```diff
@@ -323,6 +324,62 @@ def main(self) -> None:
...
+            self.output_dir.mkdir(parents=True)
+
+        _, pipeline, _ = eval_setup(self.load_config)
+
+        CONSOLE.print("Extract mesh with marching cubes and may take a while")
+
+        assert self.resolution % 512 == 0
+        # for sdf we can multi-scale extraction.
+        get_surface_sliding(
+            sdf=lambda x: pipeline.model.field.forward_geonetwork(x)[:, 0].contiguous(),
```
ethanweber, line 360: This is great for the SDF case, but you could imagine changing the lambda function if we want to do marching cubes on NeRF densities.

pablovela5620:
> > Here's an example of a room I scanned using just a video taken from my IPhone, (This doesn't include the texturing portion, only the marching cubes). I trained this using NeuS-Facto for 50k iterations ~(1 hour?) of training which I have a PR for #1654. I'll do some more digging into how to get it working like this for objects v.s. rooms
> 
> This looks really good. Does it use monocular priors? In the PR #1654, what command did you run for this? And you don't use any ground truth depth, correct?
> 
> Also happy to help more now getting this integrated. What's your preference, get this PR merged and then the neus-facto one back-to-back?

I do use monocular priors (no ground truth depth, all priors are generated via the omnidata repo as in the sdfstudio). I think this PR then neus-facto back-to-back makes sense. I'll work on resolving this week

nerfstudio/exporter/marching_cubes.py
pablovela5620: So I'm not totally sure why exactly the value 512 was chosen. Its just a fixed value from the original SDFStudio repo for the minimum grid size of (512, 512, 512), I believe it has to do with the way the multi-resolution grid is evaluated so I'll put a comment about that there but I'm not totally sure

scripts/exporter.py
pablovela5620: Agreed, changed it to more closely follow other examples like the `ExportPoissonMesh`

scripts/exporter.py
pablovela5620: I think this would require a more significant rewrite to make things work with density. I think it's probably worth doing this in a separate PR as I try to figure out the right way of doing it. I want to do some rewriting of the `base_surface_model` and `sdf_field` components as it's currently a bit unwieldy. Let me know your thoughts. I did change all of the above that you mentioned such as adding in the isosurface_threshold parameter.

ethanweber (approved):
Hey @pablovela5620, thanks for resolving the comments! It looks good to me.

PR #1621, opened 2023-03-16
Removed deprecated insta360

THE-COB:
Removed from ns-process-data and docs
also updated documentation to include cropping on equirectangular images


tancik (approved):
LGTM

PR #1622, opened 2023-03-17
Easier nerfstudio usage for research for standard pip install

jake-austin:
Bump tyro version to support (brent suggested that it be unofficial support) custom dataparsers without it crashing without needing to modify nerfstudio source code. Add documentation for a small main.py script that will allow custom configs without modifying the nerfstudio source code

TLDR: Quality of life changes for researchers adding new methods / dataparsers that just pip install and don't want to modify nerfstudio source code


tancik (approved):
LGTM

jkulhanek:
Perhaps we could extend the plugin functionality to not only look for other configs in installed packages but also in some environment variable. That way the code could be tested without having to install the package and without having to write custom train.py.

tancik:
I'm going to close this since #1869 seems like a satisfactory alternative.

PR #1623, opened 2023-03-17
Metadata for cameras

jake-austin:
Adding metadata field to cameras explicitly, will be broadcast to rays generated

I forgot to add this when I made the original cameras to tensordataclass PR... just want a metadata field that gets broadcasted to our rays. This is some code that I'm using on my own fork of this repo where my cameras have stuff in the metadata fields that works for me


tancik (approved):
Lgtm

PR #1628, opened 2023-03-17
Add option to auto select a websocket port

ethanweber:
`--viewer.websocket-port None` will auto select a free port to use


tancik (approved):
LGTM

PR #1631, opened 2023-03-18
Typo in nerfosr validation set

JADGardner:
Fixing a typo for the nerf-osr validation set, 'split' was changed to 'validation' to match nerf-osr file structure but was checking for 'val' in the if statement.


nerfstudio/data/dataparsers/nerfosr_dataparser.py
```diff
@@ -83,13 +83,12 @@ def get_camera_params(scene_dir: str, split: str) -> Tuple[torch.Tensor, torch.T
 
     Args"
       scene_dir : The directory containing the scene data.
-      split : The split for which to load the camera parameters. Either "train", "val", or "test".
+      split : The split for which to load the camera parameters. Either "train", "validation", or "test".
```
tancik, line 86: I missed this earlier, but in the definition change it to,
`def get_camera_params(scene_dir: str, split: Literal["train", "validation", "test"]) -> ...`
And then `Either "train", "validation", or "test".` can be removed from the docstring.

tancik (approved):
LGTM

PR #1579, opened 2023-03-10
Remove webrtc

tancik:
WebRTC is causing more issues than it is worth. Given that the images being sent over the network are relatively small, using a socket seems to work reasonably well and is much more robust.

This will also fix the longstanding issue of the socket connecting but the webRTC not connecting.
Fixes #765 

Shoutout to @brentyi for the idea.

PR #1639, opened 2023-03-20
add quality setting; allow png for lossless compression

machenmusik:
#1579 removed WebRTC from the viewer in favor of JPEG image compression.
However, JPEG compression is lossy, and JPEG quality was hardcoded (75, down from the default of 95).
This can cause noticeable viewer artifacts.

This PR adds: 
- by default, viewer JPEG image format with 90 quality
- configurable viewer PNG compression and JPEG quality settings

### DIFF
--- a/nerfstudio/configs/base_config.py
+++ b/nerfstudio/configs/base_config.py
@@ -30,6 +30,7 @@ from nerfstudio.utils import writer
 
 warnings.filterwarnings("ignore", module="torchvision")
 
+
 # Pretty printing class
 class PrintableConfig:  # pylint: disable=too-few-public-methods
     """Printable Config defining str function"""
--- a/nerfstudio/configs/base_config.py
+++ b/nerfstudio/configs/base_config.py
@@ -23,6 +23,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, List, Optional, Tuple, Type
 
+from typing_extensions import Literal
+
 # model instances
 from nerfstudio.utils import writer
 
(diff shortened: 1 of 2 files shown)


LownyCGI:
Hello, @machenmusik basically we can add support for openEXR aswell, right? it's more color precise.

machenmusik:
> Hello, @machenmusik basically we can add support for openEXR aswell, right? it's more color precise.

In theory, sure - although I don't think Pillow supports OpenEXR directly, and neither do web browsers.  

The viewer tends to change requested image sizes drastically during interaction, as well.  It might be better to integrate the ability to render OpenEXR, for higher quality.

I would suggest working on OpenEXR support as separate PRs, as the complexity could be significantly higher than this simple one.

nerfstudio/configs/base_config.py
```diff
@@ -148,3 +151,9 @@ class ViewerConfig(PrintableConfig):
     actually used in training/evaluation. If -1, display all."""
     quit_on_train_completion: bool = False
     """Whether to kill the training job when it has completed. Note this will stop rendering in the viewer."""
+    image_format: Literal["jpeg", "png"] = "png"
```
tancik, line 154: I think the default should stay jpeg, but we can bump the quality to 85 or 90

machenmusik: For remote usage, I can see using JPEG by default, but would be good to increase quality.

For local usage where not bandwidth constrained, PNG looks much nicer IMO.

I wonder whether the default should be "auto" or something.

tancik (approved):
LGTM

Issue #1644, opened 2023-03-21
ns-process-data ffmpeg command is broken due to [#1598]


jackson-waschura:
**Describe the bug**
If no crop options are provided and all frames are to be extracted, then the ffmpeg command constructed is invalid. It produces this:

```
ffmpeg -i "path/to/video.mp4" -pix_fmt bgr8 -vf path/to/output/images/frame_%05d.png
```

But fails with this:

```
At least one output file must be specified
```

Because the output path is being interpreted as a video filter, I believe.

**To Reproduce**
1. Locate a video (for instance at `path/to/video.mp4`)
2. Run ns-process-data using the following options (ensure number of target frames is greater than number of true frames in the video) `ns-process-data video --data path/to/video.mp4 --output-dir path/to/output --num-frames-target 9999`

**Expected behavior**
Not an error.

**Additional context**
I am running in the official Docker container so I highly doubt this is due to any issues with my ffmpeg install.

**My assessment**
I believe this can be fixed by changing this line ([source](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/process_data/process_data_utils.py#L162))
```
ffmpeg_cmd += f" -pix_fmt bgr8 -vf {crop_cmd[1:]}"
```
to
```
ffmpeg_cmd += " -pix_fmt bgr8"
if crop_cmd != "":
    ffmpeg_cmd += f" -vf {crop_cmd[1:]}"
```

PR #1645, opened 2023-03-21
Fixed incorrect ffmpeg command construction

jackson-waschura:
This PR is meant to fix Issue #1644.

Previously, the `-vf` arg was always added to the command regardless of whether there was a video filter to add or not. This PR changes this so that the `-vf` arg is only added if the `crop_cmd` is not empty (`""`).

Before making this change locally, I hit the error described in issue #1644. After making these changes, I have been able to run `ns-process-data video ...` as expected.


tancik (approved):
LGTM

Issue #1648, opened 2023-03-22
`ns-process-data realitycapture` fails when periods in filename, not only for file extension


machenmusik:
After using RealityCapture to do alignment, and then exporting as internal/external parameters, `ns-process-data realitycapture` is now failing with the following error.  (Appropriate values for --data, --csv, and --output-dir are supplied as well.)

```
  File "C:\code\nerfstudio\nerfstudio\process_data\realitycapture_utils.py", line 68, in realitycapture_to_json
    height, width, _ = img.shape
UnboundLocalError: local variable 'img' referenced before assignment
```

machenmusik:
Splitting into two issues
- failure when periods in filename not only for file extension e.g. `image.problem.jpg` which is fixed by #1649
- `RuntimeError: stack expects each tensor to be equal size, but got [718, 1086, 3] at entry 0 and [1086, 722, 3] at entry 1` which I suspect may have something to do with image rotation metadata

machenmusik:
Fixed via #1649

PR #1649, opened 2023-03-23
realitycapture: allow periods within filename, not only for file extension

machenmusik:
See #1648 which is caused in part by periods within filename, not only for file extension e.g. `image.problem.jpg`


tancik (approved):
LGTM

PR #1656, opened 2023-03-24
Imporve training stability for nerfacto & update pixsfm param

HrsPythonix:
I found that when training nerfacto for a long time (eg. 100000 iter+), it can possibly run into nan, after looking for similar issues, I found https://github.com/nerfstudio-project/nerfstudio/issues/873, and I tried gradient clipping but it did not work, disabling camera optimization helped a little, the training can go to 200000 iter+, but ran nan sooner.

After several experiments, I found that add sigmoid to geo embedding help to prevent from nan, but will reduce the psnr a little.  Another better way is to add lr decay through the scheduler, and I do not observe psnr loss. Both ways can lead to 400000 iters training without nan.

Also, I tried logging the variables to determine who appeared nan in the first place, I found it is the mlp_head who output RGB, the vars before it were normal as usual.  I haven't looked detailed into this tcnn.Network, maybe it will help to determine potential bugs

```
rgb = self.mlp_head(h).view(*outputs_shape, -1).to(directions)
```

Also I update the pixsfm parameter, previous setting can easily run into oom.

### DIFF
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -92,7 +92,8 @@ method_configs["nerfacto"] = TrainerConfig(
             train_num_rays_per_batch=4096,
             eval_num_rays_per_batch=4096,
             camera_optimizer=CameraOptimizerConfig(
-                mode="SO3xR3", optimizer=AdamOptimizerConfig(lr=6e-4, eps=1e-8, weight_decay=1e-2)
+                mode="SO3xR3", optimizer=AdamOptimizerConfig(lr=6e-4, eps=1e-8, weight_decay=1e-2),
+                scheduler = ExponentialDecaySchedulerConfig(lr_final=6e-6, max_steps=200000)
             ),
         ),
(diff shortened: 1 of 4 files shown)


kerrj:
Might be good to break the PixSFM change into a separate PR. I'm not familiar with PixSFM but the scheduler changes look good, another thing that I've found can help with long training is using the RAdam optimizer instead of Adam. If the sigmoid embedding decreases PSNR and scheduling fixes the nan issue on its own I'd advocate for leaving the sigmoid embeddings out, unless theres another reason to use them?

HrsPythonix:
> Might be good to break the PixSFM change into a separate PR. I'm not familiar with PixSFM but the scheduler changes look good, another thing that I've found can help with long training is using the RAdam optimizer instead of Adam. If the sigmoid embedding decreases PSNR and scheduling fixes the nan issue on its own I'd advocate for leaving the sigmoid embeddings out, unless theres another reason to use them?

ok I will open individual PR

PR #1658, opened 2023-03-24
Add KPlanesEncoding encoding

Giodiro:
Connected to the full KPlanes pull request (https://github.com/nerfstudio-project/nerfstudio/pull/1584) this adds the KPlanesEncoding class which can be used from a KPlanes field. 
It wraps plane creation and interpolation for 3D and 4D input coordinates.

Currently multi-scale planes are not inside the encoding, so they must be added in the field class. lmk what you think about this.

### DIFF
--- a/nerfstudio/field_components/encodings.py
+++ b/nerfstudio/field_components/encodings.py
@@ -16,8 +16,9 @@
 Encoding functions
 """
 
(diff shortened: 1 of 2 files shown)


tancik (approved):
LGTM

PR #1667, opened 2023-03-27
Fix bug: sh render is not correct

WANGSSSSSSS:
revise the SH renderer and drop the unnecessary color prior(0.5)


tancik (approved):
Lgtm, thanks for fixing.

PR #1673, opened 2023-03-28
add SO_REUSEADDR option to reuse the port in TIME-WAIT status

yzslab:
Hi,
The default websocket port in TIME-WAIT status will cause the port change to a random one. This pull request will fix it.

### DIFF
--- a/nerfstudio/viewer/server/subprocess.py
+++ b/nerfstudio/viewer/server/subprocess.py
@@ -44,6 +44,7 @@ def is_port_open(port: int):
     """
     try:
         sock = socket.socket()
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         _ = sock.bind(("", port))
         sock.close()
         return True


tancik:
I'm not super familiar with how sockets work so maybe someone else can chime in. For my sake, do you mind explaining some of the implications of this change. Does this allow multiple processes to operate on the same port? If so, is it more likely for the port to get clogged up if a lot of messages are send by multiple processes?

yzslab:
> I'm not super familiar with how sockets work so maybe someone else can chime in. For my sake, do you mind explaining some of the implications of this change. Does this allow multiple processes to operate on the same port? If so, is it more likely for the port to get clogged up if a lot of messages are send by multiple processes?

Hi, sorry for not providing a detailed explanation.
It won't allow multiple processes to bind the TCP socket on the same `IP:PORT`. TCP has a state named [TIME-WAIT (Waiting for enough time to pass to be sure that all remaining packets on the connection have expired)](https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Protocol_operation).
For example, if you start the viewer server on port 7007, and then stop the server without closing the browser that connected to the server first. The TCP port 7007 will enter TIME-WAIT state (You can see that by command `ss -ant '( sport = :7007 )'`). Without SO_REUSEADDR flag set, you will not be able to bind port 7007 again until the end of TIME-WAIT (60 seconds).
This flag is wildely used in various TCP socket servers. The tornado used by nerfstudio websocker server also set this flag automatically: [tornado/netutil.py#L135](https://github.com/tornadoweb/tornado/blob/3e96c7cb7223a05602beb9f823a6c63156fa92da/tornado/netutil.py#L135 )

tancik:
Thanks for the explanation.

tancik (approved):
LGTM

PR #1690, opened 2023-04-03
Refactor of image and video dataset conversion

MalcolmMielle:
The classes to convert images and video to a NerfStudio dataset are moved to the main repo and refactor:
* Code duplications are removed
* Import of the class is now done from nerfstudio package instead of scripts

### DIFF
--- a/nerfstudio/process_data/base_converter_to_nerfstudio_dataset.py
+++ b/nerfstudio/process_data/base_converter_to_nerfstudio_dataset.py
@@ -1,0 +1,46 @@
+# Copyright 2022 The Nerfstudio Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Base class to process images or video into a nerfstudio dataset
+"""
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from pathlib import Path
+
+
+@dataclass
+class BaseConverterToNerfstudioDataset(ABC):
+    """Base class to process images or video into a nerfstudio dataset"""
+
+    data: Path
+    """Path the data, either a video file or a directory of images."""
+    output_dir: Path
+    """Path to the output directory."""
+    verbose: bool = False
+    """If True, print extra logging."""
+
+    def __post_init__(self) -> None:
+        self.output_dir.mkdir(parents=True, exist_ok=True)
+        self.image_dir.mkdir(parents=True, exist_ok=True)
+
+    @property
+    def image_dir(self) -> Path:  # pylint: disable=missing-function-docstring
+        return self.output_dir / "images"
+
+    @abstractmethod
+    def main(self) -> None:
+        """This method implements the conversion logic for each type of data"""
+        raise NotImplementedError("the main method for conversion needs to be implemented")
--- a/nerfstudio/process_data/images_to_nerstudio_dataset.py
+++ b/nerfstudio/process_data/images_to_nerstudio_dataset.py
@@ -1,0 +1,106 @@
+# Copyright 2022 The Nerfstudio Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Processes an image sequence to a nerfstudio compatible dataset."""
+
+from dataclasses import dataclass
+from typing import Optional
+
+from rich.console import Console
+
+from nerfstudio.process_data import equirect_utils, process_data_utils
+from nerfstudio.process_data.colmap_converter_to_nerfstudio_dataset import (
+    ColmapConverterToNerfstudioDataset,
+)
+
+CONSOLE = Console(width=120)
+
+
+@dataclass
+class ImagesToNerfstudioDataset(ColmapConverterToNerfstudioDataset):
+    """Process images into a nerfstudio dataset.
+
+    1. Scales images to a specified size.
+    2. Calculates the camera poses for each image using `COLMAP <https://colmap.github.io/>`_.
+    """
+
+    def main(self) -> None:  # pylint: disable=R0915
+        """Process images into a nerfstudio dataset."""
+
+        # pylint: disable=too-many-statements
+        require_cameras_exist = False
+        if self.colmap_model_path != ColmapConverterToNerfstudioDataset.default_colmap_path():
+            if not self.skip_colmap:
+                raise RuntimeError("The --colmap-model-path can only be used when --skip-colmap is not set.")
+            if not (self.output_dir / self.colmap_model_path).exists():
+                raise RuntimeError(f"The colmap-model-path {self.output_dir / self.colmap_model_path} does not exist.")
+            require_cameras_exist = True
+
+        image_rename_map: Optional[dict[str, str]] = None
+
+        # Generate planar projections if equirectangular
+        if self.camera_type == "equirectangular":
+            pers_size = equirect_utils.compute_resolution_from_equirect(self.data, self.images_per_equirect)
+            CONSOLE.log(f"Generating {self.images_per_equirect} {pers_size} sized images per equirectangular image")
+            self.data = equirect_utils.generate_planar_projections_from_equirectangular(
+                self.data, pers_size, self.images_per_equirect, crop_factor=self.crop_factor
+            )
+
+        summary_log = []
+
+        # Copy and downscale images
+        if not self.skip_image_processing:
+            # Copy images to output directory
+            image_rename_map_paths = process_data_utils.copy_images(
+                self.data, image_dir=self.image_dir, crop_factor=self.crop_factor, verbose=self.verbose
+            )
+            image_rename_map = dict((a.name, b.name) for a, b in image_rename_map_paths.items())
+            num_frames = len(image_rename_map)
+            summary_log.append(f"Starting with {num_frames} images")
+
+            # Downscale images
+            summary_log.append(
+                process_data_utils.downscale_images(self.image_dir, self.num_downscales, verbose=self.verbose)
+            )
+        else:
+            num_frames = len(process_data_utils.list_images(self.data))
+            if num_frames == 0:
+                raise RuntimeError("No usable images in the data folder.")
+            summary_log.append(f"Starting with {num_frames} images")
+
+        # Run COLMAP
+        if not self.skip_colmap:
+            require_cameras_exist = True
+            self._run_colmap()
+            # Colmap uses renamed images
+            image_rename_map = None
+
+        # Export depth maps
+        image_id_to_depth_path, log_tmp = self._export_depth()
+        summary_log += log_tmp
+
+        if require_cameras_exist and not (self.absolute_colmap_model_path / "cameras.bin").exists():
+            raise RuntimeError(f"Could not find existing COLMAP results ({self.colmap_model_path / 'cameras.bin'}).")
+
+        summary_log += self._save_transforms(
+            num_frames,
+            image_id_to_depth_path,
+            None,
+            image_rename_map,
+        )
+
+        CONSOLE.log("[bold green]:tada: :tada: :tada: All DONE :tada: :tada: :tada:")
+
+        for summary in summary_log:
+            CONSOLE.log(summary)
--- a/scripts/process_data.py
+++ b/scripts/process_data.py
@@ -2,479 +2,37 @@
 """Processes a video or image sequence to a nerfstudio compatible dataset."""
 
 
-import shutil
 import sys
 import zipfile
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict, Optional, Tuple, Union
+from typing import Union
 
 import numpy as np
 import tyro
 from rich.console import Console
-from typing_extensions import Annotated, Literal
+from typing_extensions import Annotated
 
 from nerfstudio.process_data import (
-    colmap_utils,
-    equirect_utils,
-    hloc_utils,
     metashape_utils,
     polycam_utils,
     process_data_utils,
     realitycapture_utils,
     record3d_utils,
 )
-from nerfstudio.process_data.process_data_utils import CAMERA_MODELS
-from nerfstudio.utils import install_checks
+from nerfstudio.process_data.colmap_converter_to_nerfstudio_dataset import (
+    BaseConverterToNerfstudioDataset,
+)
+from nerfstudio.process_data.images_to_nerstudio_dataset import (
+    ImagesToNerfstudioDataset,
+)
+from nerfstudio.process_data.video_to_nerfstudio_dataset import VideoToNerfstudioDataset
 
 CONSOLE = Console(width=120)
-DEFAULT_COLMAP_PATH = Path("colmap/sparse/0")
 
 
 @dataclass
-class ProcessImages:
-    """Process images into a nerfstudio dataset.
-
-    This script does the following:
-
-    1. Scales images to a specified size.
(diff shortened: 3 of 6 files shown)


MalcolmMielle:
I ran the unit tests, but I haven't had time to test it with a real model building in case. If someone can test it, it would be great, otherwise I'll get to it probably next week. Or if someone as suggestion for unit tests, I would appreciate.

nerfstudio/process_data/base_converter_to_nerfstudio_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+
+from typing_extensions import Literal
+
+from nerfstudio.process_data import colmap_utils, hloc_utils, process_data_utils
+from nerfstudio.process_data.process_data_utils import CAMERA_MODELS
+from nerfstudio.utils import install_checks
+
+
+@dataclass
+class BaseConverterToNerfStudioDataset(ABC):
```
tancik, line 32: Here an elsewhere use lowercase `s` in Nerfstudio

nerfstudio/process_data/base_converter_to_nerfstudio_dataset.py
```diff
@@ -0,0 +1,219 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Base class to processes a video or image sequence to a nerfstudio compatible dataset."""
+
+import logging
```
tancik, line 17: Is there a reason you switched to logging from rich console? The rest of the codebase uses rich in order to get text effects, animations, emojis ect.

nerfstudio/process_data/base_converter_to_nerfstudio_dataset.py
MalcolmMielle: Sounds good :)

nerfstudio/process_data/base_converter_to_nerfstudio_dataset.py
MalcolmMielle: I'm not super familiar with rich console so I might have missed it in the documentation when I quckly read it, so I'll be happy to revert that in case. I made that change here so:

1. Using logging enable the user to set different level of severity for the prints. If the user wants only Warning and Errors it's ok. If they wants info it's also ok. If they want nothing it's also fine :).
2. The user can log in a file instead of the terminal which can be useful for tracking the logs and sharing.

I use nerfstudio as a library to run algorithms in Azure ML studio for my company, and using the console prints makes it a bit harder to log specific type of error/warning/infos, or to simply bypass them when we don't need them. What do you think?

I just realized from this comment that some warning are logged as info which is wrong. I'll correct that too.

tancik: I see, let's stick with rich for now. I think a larger change is probably needed across the codebase to accomplish what you are looking for. Here are two things that might be related - https://rich.readthedocs.io/en/stable/reference/console.html#rich.console.Console.capture
https://rich.readthedocs.io/en/stable/reference/logging.html

MalcolmMielle: Ok so I have made an update with the raise for the errors and using CONSOLE for "live logs". If you are against this, I can easily revert to using a `sys.exit` but I believe a raise is better for the user :). Up to you

Another advantage of using logging is that the code could be simplified by removing the use of a `summary_log`. Right now there is both a summary_log and a CONSOLE logging happening at the same time and we have two different process to log information and display them :). Using `logging` accross the board (to replace console and the summary_log gives more flexibility and control to the user for what is shown and when.

Let me know if it sounds interesting and I can submit another PR that does that for the IMage and Video dataset converter so you can have a look at what that would look like :).

MalcolmMielle: Fair enough, that's insight I didn't have!

What about adding a try...except in the `script/process_data.py` that will simply catch all raises and print the message that the contain? This should remove the stack trace and leave only the message, while giving programmer raises that they can handle :)?

tancik: That sounds good to me.

MalcolmMielle: I'll add this tomorrow. It's late now ;). Thanks for the feedback

MalcolmMielle: Just a side note: you could have the best of both world with [rich console handler for logging](https://rich.readthedocs.io/en/stable/logging.html). Put logging everywhere, get rich console formatting ;)

MalcolmMielle:
OK so in the latest version we have:

* Error are raised using RuntimeExpection or NotImplementedError
* Instead of using a try...except, I added an argument to the class to control the size of the traceback, with a default to 0. This way there is not need to touch the script to get the traceback back for debugging, but the non-expert user get the error message with no traceback :). Best of both world. To do this with `tyro` I didn't find any other solution than creating a new class

scripts/process_data.py
```diff
@@ -817,13 +389,53 @@ def main(self) -> None:
...
+    """Class to control the traceback size. Default to zero with no traceback shown"""
+
+    tracebacklimit: int = 0
+
+    def __post_init__(self):
+        sys.tracebacklimit = self.tracebacklimit
+
+
+@dataclass
+class StackVideoToNerfstudioDataset(StackConfig, VideoToNerfstudioDataset):
```
tancik, line 403: I wonder if it would make sense to have all of these subclass a `BaseProcessData` that also includes `output-dir` and `verbose`.

MalcolmMielle: I'm not sure I understand. Do you mean 'stackimage(stack, image2nerf, baseprocess)'? (Sorry for the poor formatting im on my Phone)

tancik: I mean something like:
```
@dataclass
class BaseProcess:
   output-dir: Path
   verbose: Boolean = False
   ...

class ProcessRecord3D(BaseProcess):
   ....
```

MalcolmMielle: That's a good suggestion. I'll update the code

MalcolmMielle: Done. However, there is a caviat. 

When inheriting dataclasses, the order of parameter can create error [due to their ordering](https://docs.python.org/3.10/library/dataclasses.html#re-ordering-of-keyword-only-parameters-in-init). This makes inheritance with dataclasses before python 3.10 quite difficult, already for the stack class, the order of import mattered. Thus, I made the dataclasses to be `keyword only` for all. I could have make only some parameters keyword only with `_: KW_ONLY` but, IMO, mixing keyword only and positional is going ot grow in cemplexity with time; it is clearer to hav all argument be keyword only :). I think what is important is the optional/non-optional and this is left untouched and work as before :).

But, to have `kw_only` with dataclasses, it is necessary to move to python 3.10. 

Let me know what you think. For now that last change is isolated in a commit so I can easily drop the commit if you don't like it :).

tancik: We do dataclass inheritance elsewhere in the codebase and it doesn't seem to cause issues, not sure what is different here. Is it just a matter of adding an empty `Path()` default? We want to avoid bumping to 3.10 (colab doesn't easily support it atm)

Also I think there was a little confusion in my suggestion, I was thinking more like
```
@dataclass()
class BaseConverterToNerfstudioDataset(ABC):
    """Base class to process images or video into a nerfstudio dataset"""

    data: Path = Path()
    """Path the data, either a video file or a directory of images."""
    output_dir: Path = Path()
    verbose: bool = False
    """If True, print extra logging."""
    tracebacklimit: int = 0
    """Control the traceback size. Defaults to no traceback shown"""

    def __post_init__(self) -> None:
        sys.tracebacklimit = self.tracebacklimit
        self.output_dir.mkdir(parents=True, exist_ok=True)
        self.image_dir.mkdir(parents=True, exist_ok=True)

    @property
    def image_dir(self) -> Path:  # pylint: disable=missing-function-docstring
        return self.output_dir / "images"

    @abstractmethod
    def main(self) -> None:
        """This method implements the conversion logic for each type of data"""
        raise NotImplementedError("the main method for conversion needs to be implemented")
```
Then
```
@dataclass
class ProcessRealityCapture(BaseConverterToNerfstudioDataset):
...
```

tancik: Thanks for taking the time to go back and forth on this. I agree its not ideal to hardcode default values. It looks like we don't actually need to for inheriting dataclasses, the following works in 3.8
```
@dataclass
class A:
    a: int
    
@dataclass
class B(A):
    c: int = 3
```

Regarding your second point, I was imagining that `BaseConverterToNerfstudioDataset` could exist in `process_data.py`, ie app level.

MalcolmMielle: No problem. That's what reviews are for ;). It's a good learning opportunity for me too, I get to understand your library better :).

I had tried that first and it didn't work, I think the problem was that `xml` in `ProcessMetashape` and `csv` in `ProcessRealityCapture` do not have a default value. But this problem made me realize something. The base class add a `data` argument that fit the exact definition of `xml` and `csv`, it is a Path to the data to use. So I removed `xml` and `csv` and replaced them by `data` and now all work as expected :D! I have updated the documentation so that no information is lost.

I would prefer having `BaseConverterToNerfstudioDataset` in the library level, so that all converter have a predictable interface based on this class :). As is the code now, it should be working with python 3.8 out of the box. Thanks again for all the suggestions.

nerfstudio/process_data/images_to_nerstudio_dataset.py
```diff
@@ -0,0 +1,106 @@
...
+
+from nerfstudio.process_data import equirect_utils, process_data_utils
+from nerfstudio.process_data.colmap_converter_to_nerfstudio_dataset import (
+    ColmapConverterToNerfstudioDataset,
+)
+
+CONSOLE = Console(width=120)
+
+
+@dataclass(kw_only=True)
```
tancik, line 30: `kw_only` still here

scripts/process_data.py
tancik: `ProcessMetashape` and `ProcessRealityCapture` will now fail if you don't provide the image data folder.

pyproject.toml
```diff
@@ -8,7 +8,7 @@ version = "0.1.19"
 description = "All-in-one repository for state-of-the-art NeRFs"
 readme = "README.md"
 license = { text="Apache 2.0"}
-requires-python = ">=3.7.3"
+requires-python = ">=3.10"
```
tancik, line 11: Needs to be reverted

MalcolmMielle: Yes! Sorry I missed that, thanks for letting me know. Will do tomorrow!

scripts/process_data.py
MalcolmMielle: Fair, il have a look to remove this parameter. I'll check if I can add unit tests to catch those kind of mishaps before it's send to review.

MalcolmMielle: Actually sorry I am confused :S. Wasn't it always the case?

Previously `ProcessRealityCapture` was defined as: 

```python
@dataclass
class ProcessRealityCapture:
    """Process RealityCapture data into a nerfstudio dataset.

    This script assumes that cameras have been aligned using RealityCapture. After alignment, it is necessary to
    export the camera poses as a `.csv` file using the `Internal/External camera parameters` option.

    This script does the following:

    1. Scales images to a specified size.
    2. Converts RealityCapture poses into the nerfstudio format.
    """

    data: Path
    """Path to a folder of images."""
    csv: Path
    """Path to the RealityCapture cameras CSV file."""
    output_dir: Path
    num_downscales: int = 3
    """Number of times to downscale the images. Downscales by 2 each time. For example a value of 3
        will downscale the images by 2x, 4x, and 8x."""
    max_dataset_size: int = 600
    """Max number of images to train on. If the dataset has more, images will be sampled approximately evenly. If -1,
    use all images."""
    verbose: bool = False
    """If True, print extra logging."""
```

and `data` had no default argument so needed to be defined and was not used in the code, as far as I understand? The new version has only `data` and no `csv` so you need only to defined `data` I believe.

Sorry for that this takes so much time in the end :/. I would make unit tests for it but I don't really see the problem at the moment and I don't have data/a workflow to test those classes.

tancik: The original code used `data` in the code here - https://github.com/nerfstudio-project/nerfstudio/blob/b85cb9f97982784d88abdcf08770d378f8960cb4/scripts/process_data.py#L777

MalcolmMielle: Thanks for the pointer! I believe it should be fixed now! I used the trick that I found documented for dataclasses pre-3.10: separate the default and non default argument in two different class and use inheritance in a clever order (as I did for the StackConfig actually). I made the two classes private and I think the names are descriptive enough that it should be understandable for new-comers _and_ easy to know that they should be removed when nerfstudio will switch to 3.10 :).

Let me know what you think, and again, thanks for taking the time.

scripts/process_data.py
```diff
@@ -732,7 +296,17 @@ def main(self) -> None:
 
 
 @dataclass
-class ProcessRealityCapture:
+class _NoDefaultrocessRealityCapture:
```
tancik, line 299: `_NoDefaultrocessRealityCapture` -> `_NoDefaultProcessRealityCapture:`

MalcolmMielle: Thanks for the comment, it's corrected.

tancik (approved):
LGTM

MalcolmMielle:
@tancik I rebased on the current main and added the `self.crop_bottom` to the colmap class to keep the new behavior. I also added a raise to make sure the parameter is within the correct bound but I didn't touch anything else :)

scripts/process_data.py
```diff
@@ -834,13 +383,53 @@ def main(self) -> None:
         CONSOLE.rule()
 
 
+@dataclass
+class StackConfig:
```
brentyi, line 387: I'm not a huge fan of the boilerplate here; if we don't want the full traceback, can we just use a try/except in the entrypoint?
```
    try:
        tyro.cli(Commands).main()
    except RuntimeError as e:
        CONSOLE.log("[bold red]" + e.args[0])
```

nerfstudio/process_data/images_to_nerstudio_dataset.py
```diff
@@ -0,0 +1,106 @@
...
+
+@dataclass
+class ImagesToNerfstudioDataset(ColmapConverterToNerfstudioDataset):
+    """Process images into a nerfstudio dataset.
+
+    1. Scales images to a specified size.
+    2. Calculates the camera poses for each image using `COLMAP <https://colmap.github.io/>`_.
+    """
+
+    def main(self) -> None:  # pylint: disable=R0915
```
brentyi, line 38: Not needed for this PR, but just wondering @MalcolmMielle do you think there's more refactoring that could be done to consolidate the image / video processing logic?

It'd also be nice to reduce the inheritance layers, but I don't have any concrete suggestions for how to do that cleanly 🤷

brentyi (approved):
Sorry for dropping in late @tancik @MalcolmMielle!

This PR just popped on my radar and looks great, just had some minor questions.

scripts/process_data.py
```diff
@@ -662,7 +207,17 @@ def main(self) -> None:
...
+    """
+    Private class to order the parameters of ProcessMetashape in the right order for default values
+    """
+
+    xml: Path
+    """Path to the Metashape xml file."""
+
+
+@dataclass
+class ProcessMetashape(BaseConverterToNerfstudioDataset, _NoDefaultProcessMetashape):
```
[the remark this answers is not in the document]
MalcolmMielle: I haven't used those classes and I have no refactored them so there is still redundant code between them. Furthermore, I have never used them and I'm afraid of moving them in the main pages because they have no unit tests associated.

nerfstudio/process_data/images_to_nerstudio_dataset.py
MalcolmMielle: I've tried my best hahaha. There might be more that can be done later but for now I don't see more :). 

With python 3.10 we could use 'kw_only' on the dataclasses to reduce the inheritance with the "default for order" classes.

scripts/process_data.py
brentyi: Got it, that makes sense! I have a medium preference for eliminating the extra complexity + layer of inheritance / redirection.

> However I wanted to keep the ability to have the stack trace for debugging purpose and enclosing in a try catch remove that possibility.

My feeling is that it may be good to disentangle errors from incorrect usage (like passing in an invalid argument or directory) vs actual errors (like bugs or dependency issues). To me it makes the most sense to suppress traces for the former case while retaining them for the latter.

MalcolmMielle:
@brentyi Thanks for the comments :). I have updated the PR to reflect them.

@tancik thanks for all thze feedback and feel free to merge whenever you want :D

brentyi:
I'm going ahead and merging; I did another pass on this PR last night and the only thing I noticed was some silly docs stuff. :)

Thanks for the PR @MalcolmMielle! There's an infinite amount of work to do in the cleanup/refactor direction but I'm glad we're taking steps.

PR #1702, opened 2023-04-05
Viewer backend refactor

brentyi:
cc @chungmin99 @kerrj @ethanweber @tancik 
TLDR: Snappier more extendable backend

This PR replaces the nerfstudio viewer backend with [viser](https://github.com/brentyi/viser). Currently the code is coppied to get it to fit into nerfstudio but may be migrated to use a visor pip package when viser is more developed.

Why:
+ Replacing the bridge server with a direct websocket connection is simpler, and will make interrupt-style callbacks easier. 
+ No longer need request/response-style polling when GUI elements get updated
+ Messages will be typed, which will be helpful for robustness + development.
+ We can leverage viser's infrastructure for Python-configurable GUI elements, which will be nice for integrating LERF and other projects.
+ Eventually, less code to maintain in `nerfstudio`. (it will require thought but probably `viser` can be used as a dependency)

Regressions:
+ [UI] Control panel buttons for training speed and resolution are no longer supported (could be added later)
+ The UI logic for time dependent controls have been removed for simplicity. Time dependent models can now implement `time` sliders directly in the model. However this does not support adding animated time to the camera path export. We will need to either think more deeply about enabling animation of custom gui elements or reimplement the time logic.


nerfstudio/viewer/server/viewer_param.py
```diff
@@ -1,27 +1,57 @@
...
+    def install(self, viser_server: ViserServer):
+        raise RuntimeError("Must implement install() for a ViewerElement")
+
+
+class ViewerButton(ViewerElement):
+    def __init__(self, name: str, callable: Callable, disabled=False):
+        super().__init__(name, disabled=disabled)
+        self.fn = callable
+
+    def _create_gui_handle(self, viser_server: ViserServer):
```
brentyi, line 32: We should probably add return annotations, eg `def _create_gui_handle(self, viser_server: ViserServer) -> None:`

kerrj: true

nerfstudio/viewer/server/viewer_utils.py
```diff
@@ -716,11 +719,11 @@ def _update_viewer_stats(self, render_time: float, num_rays: int, image_height:
             self.vis["renderingState/train_eta"].write("Paused")
             self.vis["renderingState/vis_train_ratio"].write("100% spent on viewer")
 
-    def _calculate_image_res(self, camera_object, is_training: bool) -> Optional[Tuple[int, int]]:
+    def _calculate_image_res(self, aspect_ratio: float, is_training: bool) -> Optional[Tuple[int, int]]:
         """Calculate the maximum image height that can be rendered in the time budget
 
         Args:
-            camera_object: the camera object to use for rendering
+            apect_ration: the aspect ratio of the current view
```
brentyi, line 734: typo 👀

nerfstudio/viewer/server/viewer_utils.py
```diff
@@ -90,57 +88,41 @@ def __exit__(self, ext_type, exc_value, traceback):
...
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        _ = sock.bind(("", port))
+        sock.close()
+        return True
+    except OSError:
+        return False
+
+
+def get_free_port(default_port: Optional[int] = None):
+    """Returns a free port on the local machine. Try to use default_port is possible.
```
brentyi, line 111: "is possible" => "if possible" 🕺

kerrj (approved):
happy birthday Matt!

PR #1709, opened 2023-04-07
Small updates for front-facing scenes

ayaanzhaque:
Added an option to use AABBBoxCollider in nerfacto, added a parameter to shift poses for training

### DIFF
--- a/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
+++ b/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 import math
 from dataclasses import dataclass, field
 from pathlib import Path, PurePath
-from typing import Optional, Type
+from typing import Optional, Tuple, Type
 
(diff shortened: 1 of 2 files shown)


nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -236,6 +238,9 @@ def _generate_dataparser_outputs(self, split="train"):
 
         poses[:, :3, 3] *= scale_factor
 
+        if self.config.shift_poses is not None:
+            poses[:, :3, 3] += torch.tensor(self.config.shift_poses)
```
tancik, line 242: The `transform_matrix` needs to be modified to take into account this shift. The matrix allows people to "undo" the  transforms applied by the dataparser (in this case by the transform applied by `auto_orient_and_center_poses`). It's tricky in this case though because shift is applied after the scale. Perhaps this shift should be moved into the `auto_orient_and_center_poses`

PR #1713, opened 2023-04-09
Add support for torch v2.0.0

jkulhanek:
This PR fixes the codebase such that it works with torch 2.0.0.

However IT DOES NOT change the pip dependencies for the following reason:
     torch >= 1.13 includes functorch, installing functorch pip package breaks imports (breaks in 2.0.0).
     torch < 1.13 does not include functorch and it needs to be installed.
Do we still need to support 1.12?

### DIFF
--- a/nerfstudio/data/utils/nerfstudio_collate.py
+++ b/nerfstudio/data/utils/nerfstudio_collate.py
@@ -23,7 +23,12 @@ from typing import Callable, Dict, Union
 
 import torch
 import torch.utils.data
-from torch._six import string_classes
+
+try:
+    from torch._six import string_classes
+except ImportError:
+    # _six deprecated in torch v2
+    string_classes = (str, bytes)
 
 from nerfstudio.cameras.cameras import Cameras
 


tancik:
Do the models still perform the same?

jkulhanek:
I haven't tested the models' performance under the new version.

nerfstudio/data/utils/nerfstudio_collate.py
```diff
@@ -23,7 +23,12 @@
 
 import torch
 import torch.utils.data
-from torch._six import string_classes
+
+try:
+    from torch._six import string_classes
+except ImportError:
+    # _six deprecated in torch v2
+    string_classes = (str, bytes)
```
tancik, line 31: Can we just edit ln 127 to `elif isinstance(elem, (str, bytes)):`?

jkulhanek: Sure

adithya-Avataar:
hi @jkulhanek I was wondering if you have tried using the torch.compile feature of 2.0 with nerfstudio. I had tried using the torch.compile feature, but it was unable to build the graph. Just curious to know if you have some info on this.

jkulhanek:
Hi, I don’t think it is possible to build the graph with tcnn and nerfacc custom cuda ops.

jkulhanek:
On another note, nerfacto seems not to be working with torch 2.0. I get the following error: https://github.com/Lightning-AI/lightning/issues/17245

adithya-Avataar:
Got it. Thank you

KamiCalcium:
> 

Hi,

Did you try the workaround mentioned in that thread? I'm having the same problem for torch 2.0.

jkulhanek:
No, lightning is not used, I was just pointing at other instances of the same error. There must be something in AMP optimisation which causes it to fail. Could the problem be caused by some gradients being computed inside of the custom tcnn kernels?

KamiCalcium:
> No, lightning is not used, I was just pointing at other instances of the same error. There must be something in AMP optimisation which causes it to fail. Could the problem be caused by some gradients being computed inside of the custom tcnn kernels?

Could be yeah. Didn't dig too much but the easiest workaround I found is to comment out that assertion and it worked..

gtangg12:
Few other additional things:

1) In nerfstudio/utils/profiler.py
`from torch.autograd.profiler import ContextDecorator `
ContextDecorator doesn't exist anymore. I think it may be _ContextDecorator.

2) One cannot index a cpu tensor with a cuda tensor anymore. In dataloaders, this is done (image is on cpu while image_index and its derivatives are on cuda).

jkulhanek:
1. Sorry, I have introduced this one.
2. Wasn’t this already the case in 1.13?

gtangg12:
Yea, I was on 1.12.1. Just a bit confused since image idx is on cuda but image cpu. Should I convert them all to cuda in the dataset?

jkulhanek:
I think in previous torch versions if the sampled tensor was on cpu the indices were moved to cpu as well. I would think that this would be more performant then moving both to cuda as you would have to move all cached images to cuda… could be quite expensive for large datasets. I guess this very much depends on the data, but sampling on cpu should be safer.

gtangg12:
Thanks, makes sense, thought this brings up another issue: in the dataloaders only image is excluded but mask, depth, and such are still being moved to cuda for the batch, which defeats the purpose of this ig.

jkulhanek:
True, in another PR I wanted to address this inconsistency. I didn’t have the time to test it though.

machenmusik:
Re: masks - see #1741?

jkulhanek:
Depth: #1730

PR #1717, opened 2023-04-10
Remove useless code in camera_paths

97littleleaf11:
The `camera_type` will be set in the following lines.


tancik (approved):
LGTM, thanks for cleaning up

PR #1718, opened 2023-04-10
Move model loading logic to the model class

MalcolmMielle:
A model can now be loaded from a pipeline state without having to load the dataloader or pipeline itself. This enable users to do inference on model without having to share the data used to create the model.


nerfstudio/pipelines/base_pipeline.py
```diff
@@ -383,11 +364,11 @@ def load_pipeline(self, loaded_state: Dict[str, Any], step: int) -> None:
...
-        state = {
-            (key[len("module.") :] if key.startswith("module.") else key): value for key, value in loaded_state.items()
-        }
-        self.model.update_to_step(step)
-        self.load_state_dict(state, strict=True)
+        self.model.load_model_from_pipeline_state(loaded_state, step)
+
+        state = Model.parse_pipeline_state(loaded_state)
+        pipeline_state = {key: value for key, value in state.items() if not key.startswith("_model.")}
+        self.load_state_dict(pipeline_state, strict=False)
```
ethanweber, line 371: Why is strict=False instead of strict=True now? Any reason in particular? I guess it's since we remove the model keys/values from the state. This could maybe lead to some undesired behavior if, for example, the state for optimized cameras silently fails. Let me know what you think.

tests/models/test_models.py
```diff
@@ -0,0 +1,225 @@
...
+    pipeline = VanillaPipeline(config, "cpu")
+
+    state_dict = pipeline.state_dict()
+    # Model weights are set to a known fixed value to compare the result of the model read
+    state_dict[  # pylint: disable=unsubscriptable-object,unsupported-assignment-operation
+        "_model.lpips.net.lin0.model.1.weight"
+    ] = torch.Tensor(
+        [
+            [
+                [[1]],
```
ethanweber, line 68: can you do list comprehension here? e.g., `[[[1]] for _ in range(N)]` may be cleaner

tests/models/test_models.py
```diff
@@ -0,0 +1,225 @@
...
+    # A new model is created with known weights of different value than the pipeline
+    config = VanillaModelConfig(_target=NeRFModel)
+    model = NeRFModel(config, scene_box=SceneBox(), num_train_data=1)
+    model_state_dict = model.state_dict()
+    model_state_dict[
+        "lpips.net.lin0.model.1.weight"
+    ] = torch.Tensor(  # pylint: disable=unsubscriptable-object,unsupported-assignment-operation
+        [
+            [
+                [[1]],
```
ethanweber, line 145: same comment as before

tests/models/test_models.py
```diff
@@ -0,0 +1,225 @@
+"""
+Test model
```
ethanweber, line 2: The filename `test_models.py` and this comment are misleading. Can you rename it to something to do with saving and loading from checkpoints? `test_models.py` implies the models are being tested. I know this deviates slightly from our convention but I think it'll be more clear in this case.

ethanweber (commented):
Thanks for the PR! Just left some comments.

nerfstudio/pipelines/base_pipeline.py
MalcolmMielle: Thanks for the review. So I know it looks like I did, and when proofreading I had the same question, but I actually didn't :).

There used to be an override of the method `load_state_dict` in `Pipeline` that would parse the model and then the pipeline itself by calling `super().load_state_dict(pipeline_state, strict=False)` at line 127. Since the model loading logic was moved to the `Model` class, we are left here with only `super().load_state_dict(pipeline_state, strict=False)` in the override and thus I have removed the override and I'm calling the torch `load_state_dict` directly here (the same one that previously was line 127). Hence, since the one line 127 was called with False, I've left the same logic here, but the git diff makes it look like I changed the call to the override :S

tests/models/test_models.py
MalcolmMielle: No problem I can change it, maybe to `test_load_model.py`? What do you think? For now I followed the convention on `test_vanilla_pipeline.py` which tests that the pipeline can be loaded similarly to this test :)

MalcolmMielle:
@ethanweber thanks! I've pushed the modifications :)

jkulhanek:
I think that moving the pipeline-loading logic to the model is a bad design. It couples the two and could lead to lots of hidden bugs. Besides, is there a need? When you need to load a checkpoint, cannot you just construct the pipeline while replacing the datamanager config with an empty one?

MalcolmMielle:
> I think that moving the pipeline-loading logic to the model is a bad design. It couples the two and could lead to lots of hidden bugs. Besides, is there a need? When you need to load a checkpoint, cannot you just construct the pipeline while replacing the datamanager config with an empty one?

Thanks for the feedback, I appreciate you taking the time to have a look.

My goal with this PR is to move toward decoupling the pipeline and the model loading so that the relevant part are in the relevant classes. I agree that having them coupled is a bad design. Right now, it is fully coupled in nerfstudio since all of the logic (incuding the model loading logic) is in the pipeline and a model cannot be loaded alone :). A nicer way IMO, is if the load_pipeline handles loading the pipeline elements and then calls a load_model that loads the model part. It is true that my PR was doing half of the job and so I've updated it to fully decouple the loading logic now :D.

To answer the question on giving dummy data to the method and the actual need for this PR:

* For loading, one need a valid `transforms.json` even though it serves no purpose for the model since there are checks on it.
* It's also confusing to the user to have to input dummy data to load a model through a pipeline object. It took my team longer to understand the code and test/make sure that there no side effect from the dummy data. it also makes the code harder to understand and maintain since everything is so strongly coupled. This PR makes it easier to understand the code since the different part to load are separated in logic units relevant to each classes.
* The code is also harder to tests since the logic is all lumped together in one function. Now it is possile to unit tests each step independently (which this PR adds)

I try to contribute in my free time, stuff that resolved pain point for my team :).

Thanks again for the comments @jkulhanek , I think they made this PR much better!

PS: can someone point me to what the `load_model` function do or where it is used? I couldn't find unit tests or other to understand and I would have liked to use that name instead of  :/.

jkulhanek:
I would still prefer not to introduce any parse methods on the model. I believe the model should be only responsible for loading its state and should never “see”the full checkpoint. IMHO, calling a private static method on the pipeline from the model doesn’t seem to me like a safe thing to do.

Besides, isn’t model itself useless? You always need pixel samplers and other code from the pipeline. Wouldn’t it be better to add an option to the pipeline not to instantiate the dataparser? But I guess it wouldn’t be that easy since parts of the code depend on dataparser results (eg the dataset metadata) and these data should be stored in the checkpoint itself?

Disclaimer: I have no authority in this regard  (I am not a maintainer) and I only share my opinions. It could be that I am wrong about this.
Perhaps someone else with a better knowledge could share their opinion?

MalcolmMielle:
@jkulhanek it's good to talk about those things :).

With this PR the model does not see the state of the checkpoint and only loads its state. Because the pipeline saw all the states and hte model saw nothing.  I've actually implemented what you are talking about ;). `Pipeline.parse_model_state` extract the model state form the pipeline state and the user can then feed it to the model class directly. The state extracted is a classic pytorch state dict, unless it's PPD and so the loading is only there to accomodate this other format. Hence everything model state related is now in the model itself. That means that later, if someone decides to save the model on its own using pytorch, the loading function will still work (I actually could unit test this). Calling a static method is actually the safe thing to do since it makes it clear that no Pipeline object's state is changed and that this method is thus free of side effects on the class.

Looking at the [rendering script](https://github.com/nerfstudio-project/nerfstudio/blob/main/scripts/render.py), the pipeline is first loaded and then fed into `_render_trajectory_video` where it's always used to called up the model through `pipeline.model` (hence the pipeline object is in itself not used apart from where it's used to get the device which you can also get from the model). IMO, the rendering script should be updated to remove the dependency to the Pipeline object if anything. This is what we successfully did on our production codebase :)

You actually have nailed it with your comment on not instanciating the dataparser being difficult: it is not used in the demo rendering script and it's difficult to either not instantiate or give dummy data :).

MalcolmMielle:
@ethanweber sorry to bother you :). Any update on this PR :)?

MalcolmMielle:
I converted to draft because there is problem with the appearance embeddings of nerfacto

PR #1736, opened 2023-04-14
return unscaled input in positional encoding

hturki:
Right now the [0, 1] in_tensor is getting multiplied to [0, 2pi], and then being returned in that range instead of being returned unmodified


hturki:
@tancik  I think the "Viewer Build and Deploy. / Build (pull_request)" test just failed (doubt this PR is causing it). In either case I don't think that I'm able to merge this PR myself

PR #1738, opened 2023-04-14
Respect wandb environment variables

jkulhanek:
Current implementation of wandb writer calls wandb.init ignoring WANDB_* environment variables as documented here:
https://docs.wandb.ai/guides/track/environment-variables

This PR changes the implementation such that these env. variables are respected


tancik (approved):
Lgtm, thanks for updating, I want aware those environment variables existed.

PR #1730, opened 2023-04-13
Reduce GPU memory usage in CacheDataloader

jkulhanek:
This PR implements functionality to allow Datasets to keep some tensors on CPU (same as "image") - reducing GPU memory requirements.
Currently, I moved the "depth_image" tensor to GPU.

Note, I haven't tested it yet. Which datasets should I test?

PR #1741, opened 2023-04-14
Allow masks with instant ngp dataparser

nepfaff:
### Changes
- Allow using masks with the `instant-ngp-dataloader`
- Move the entire dataset to the GPU when using masks. I feel like this is a reasonable compromise (see [here](https://github.com/nerfstudio-project/nerfstudio/issues/1465))

### DIFF
--- a/nerfstudio/data/dataparsers/instant_ngp_dataparser.py
+++ b/nerfstudio/data/dataparsers/instant_ngp_dataparser.py
@@ -65,6 +65,7 @@ class InstantNGP(DataParser):
             data_dir = self.config.data
 
         image_filenames = []
+        mask_filenames = []
         poses = []
         num_skipped_image_filenames = 0
         for frame in meta["frames"]:
@@ -87,6 +88,9 @@ class InstantNGP(DataParser):
                         meta["h"] = h
                 image_filenames.append(fname)
                 poses.append(np.array(frame["transform_matrix"]))
+                if "mask_path" in frame:
+                    mask_fname = data_dir / Path(frame["mask_path"])
+                    mask_filenames.append(mask_fname)
         if num_skipped_image_filenames >= 0:
             CONSOLE.print(f"Skipping {num_skipped_image_filenames} files in dataset split {split}.")
         assert (
@@ -144,6 +148,7 @@ class InstantNGP(DataParser):
             image_filenames=image_filenames,
             cameras=cameras,
             scene_box=scene_box,
+            mask_filenames=mask_filenames if len(mask_filenames) > 0 else None,
             dataparser_scale=self.config.scene_scale,
         )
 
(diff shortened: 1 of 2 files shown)


tancik:
This looks like a fine compromise to me. Curious to get feedback from @machenmusik since he has thought about this a bit.

machenmusik:
Personally I think this approach is reasonable as well, just two things
- can this be done more generally, and not just to individual methods?
- how does this interact, if at all, with #1730?

machenmusik:
If this makes masks work for ingp and nerfacto (maybe even semantic-nerfw and others?) then personally, I think we should merge

nepfaff:
> does this work more generally

We would need to add the same lines to every dataparser to make it work for that parser. There isn't a clean way around this without some refactoring. 

> how does this interact, if at all, with https://github.com/nerfstudio-project/nerfstudio/pull/1730?

It affects it in a way that this PR refactors this logic. We could keep this logic there, with an additional check of whether "mask" is already in  `exclude_batch_keys_from_device`. It would be cleaner to set `exclude_batch_keys_from_device` at a single place though, so I would add a similar check to where this variable is initialized (multiple places).

nepfaff:
The downside of this approach is that it is still easy to run out of GPU memory for decently large datasets as we are now loading all masks and images into GPU memory. However, not loading masks into the GPU leads to `nerfacto` taking 2h instead of 20min for the default iterations on my machine.

At a minimum, we should make sure that masks and images are on the same device to avoid an exception. This is not the case on master where masks are always loaded into GPU memory, even if images are not.

machenmusik:
Thanks.  If there were an overall policy setting (favor speed, favor min GPU mem) we might want to respect that here (images/masks both the GPU, or not, depending).  But I don't think there is one, and this is better than the earlier change which was reverted.

nepfaff:
> Is there a dataset to check this with after #1809 ?

I don't think this is related, as this PR modified the `instant-ngp` models but not the dataparser, which is model-agnostic. I mainly use this dataparser with `nerfacto`, as I wrote my data pre-processing pipeline for `instantNGP`.

Hence, training any model with data from this dataparser should be sufficient for testing. Should I share some results?

machenmusik:
> Hence, training any model with data from this dataparser should be sufficient for testing. Should I share some results?

That would be great, assuming they are good then I think we can merge.

nepfaff:
I rebased on master and trained on synthetic data of the YCB mustard bottle with masks:
`ns-train nerfacto --vis viewer+wandb instant-ngp-data --data data/mustard`. Maybe I shouldn't have pushed the rebase due to the checks...

Training took around 20min, and the results clearly show the effect of the masks. Noting that this is not necessarily the intended use case of the current masking implementation :)

image

Textured TSDF mesh:
nerfstudio_mustard_nerfacto_w_masks_tsdf_mesh

Are there any other results/ metrics that would be useful?

machenmusik:
LGTM, thanks!

Tao-11-chen:
Why don't you put the sampled indices from GPU to CPU which is both fast and memory efficient. #1819

PR #1743, opened 2023-04-15
Improve viewer element types

brentyi:
The main change is adding a type to the setter/getter for `.value`.

So `checkbox_element.value` will resolve to `bool`, `rgb_element.value` will resolve to `Tuple[int, int, int]`, etc.


tancik (approved):
Lgtm

PR #1748, opened 2023-04-17
Delete useless lists of tensors put to CUDA which may cause OOM.

Tao-11-chen:
This part of data is not used but very big(RaySamples and weights), may cause OOM while evaluate with large images(2000*2000).


tancik (approved):
LGTM, thanks for the optimization

PR #1757, opened 2023-04-18
Fix divide by zero when train util set to 1

dylanhu7:
## Changes
- In `nerfstudio/viewer/server/viewer_state.py`, adds the condition `self.control_panel.train_util != 1` for rendering while training. This prevents crashing as a result of division by 0 when Train Util is set to `1.00` and completely halts rendering during training, which I believe is reasonable behavior.

### DIFF
--- a/nerfstudio/viewer/server/viewer_state.py
+++ b/nerfstudio/viewer/server/viewer_state.py
@@ -303,11 +303,7 @@ class ViewerState:
         if self.camera_message is None:
             return
 
-        if (
-            self.trainer is not None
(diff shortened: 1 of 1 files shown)


brentyi (approved):
looks good to me!

brentyi:
FYI for next time `ns-dev-test` will automatically run all the tests / code formatters / linters / etc 🙂

dylanhu7:
Thanks! Will do in the future, first time contributing here so I overlooked that o.o

PR #1762, opened 2023-04-18
add individual grad_scaler config

wuzirui:
I am adding a nerfstudio implementation of the neural scene graph paper, and I tried to use separate nerf(nerfacto) models to represent the background and each foreground object, respectively. However, training such a complex model easily runs into nan issues at early iterations (~3K steps, which is obviously different from the ones in #873). The issue is solved by using grad_scaler only (w/o using amp).

### DIFF
--- a/nerfstudio/engine/trainer.py
+++ b/nerfstudio/engine/trainer.py
@@ -75,6 +75,8 @@ class TrainerConfig(ExperimentConfig):
     """Maximum number of iterations to run."""
     mixed_precision: bool = False
     """Whether or not to use mixed precision for training."""
+    use_grad_scaler: bool = False
+    """Use gradient scaler even if amp is disabled"""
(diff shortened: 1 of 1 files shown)


nerfstudio/engine/trainer.py
```diff
@@ -75,6 +75,8 @@ class TrainerConfig(ExperimentConfig):
     """Maximum number of iterations to run."""
     mixed_precision: bool = False
     """Whether or not to use mixed precision for training."""
+    use_grad_scaler: bool = False
+    """Use gradient scaler even if amp is disabled"""
```
tancik, line 79: Reword incase people don't know what amp is
"""Use gradient scaler even if mixed_precision is disabled"""

tancik (approved):
LGTM

PR #1767, opened 2023-04-18
Added docs wording change and also --crop-bottom

THE-COB:
Changed the "obscene mesh of human" wording and added `--crop-bottom [num]` arg to equal `--crop-factor 0.0 [num] 0.0 0.0`


tancik (approved):
LGTM

PR #1776, opened 2023-04-19
Set data path to launch viewer for DreamFusion

DveloperY0115:
The script _run_viewer.py_ crashes when it is provided a config for a pre-trained DreamFusion model.

It is because of _None_ passed to function _viewer_utils.setup_viewer_ via parameter _datapath_,  since generative models (i.e., DreamFusion) are trained without real images.

This commit modifies the code to distinguish _generative_ model from the other.

### DIFF
--- a/scripts/viewer/run_viewer.py
+++ b/scripts/viewer/run_viewer.py
@@ -63,16 +63,13 @@ class RunViewer:
         viewer_log_path = base_dir / config.viewer.relative_log_filename
 
         # NOTE: for generative pipeline(s), set the data path on-the-fly.
-        if pipeline.config.generative:
-            assert pipeline.datamanager.get_datapath() is None
-            CONSOLE.print(f"Generative models should not have data path. Setting it to {str(base_dir)}")
-            viewer_state, banner_messages = viewer_utils.setup_viewer(
-                config.viewer, log_filename=viewer_log_path, datapath=base_dir
-            )
-        else:
-            viewer_state, banner_messages = viewer_utils.setup_viewer(
-                config.viewer, log_filename=viewer_log_path, datapath=pipeline.datamanager.get_datapath()
-            )
+        viewer_state, banner_messages = viewer_utils.setup_viewer(
+            config.viewer,
+            log_filename=viewer_log_path,
+            datapath=pipeline.datamanager.get_datapath()
+            if pipeline.datamanager.get_datapath() is not None
+            else base_dir,
+        )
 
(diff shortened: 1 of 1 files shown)


DveloperY0115:
Just found out how to test my code before making PR on the documentation page.

Temporarily closing this PR.

tancik (dismissed):
LGTM

scripts/viewer/run_viewer.py
```diff
@@ -61,9 +61,18 @@ def main(self) -> None:
...
+        # NOTE: for generative pipeline(s), set the data path on-the-fly.
+        if pipeline.config.generative:
+            assert pipeline.datamanager.get_datapath() is None
+            CONSOLE.print(f"Generative models should not have data path. Setting it to {str(base_dir)}")
+            viewer_state, banner_messages = viewer_utils.setup_viewer(
+                config.viewer, log_filename=viewer_log_path, datapath=base_dir
+            )
+        else:
+            viewer_state, banner_messages = viewer_utils.setup_viewer(
+                config.viewer, log_filename=viewer_log_path, datapath=pipeline.datamanager.get_datapath()
```
tancik, line 74: I could potentially imagine generative methods that do rely on underlying data and therefore have a datapath. Maybe we should just default to `base_dir` if `pipeline.datamanager.get_datapath()` is `None`

DveloperY0115: Thank you for your comment.

Agree, a conditional or editing setup may expect data that could be used as a starting point.

scripts/viewer/run_viewer.py
```diff
@@ -63,16 +63,13 @@ def _start_viewer(self, config: TrainerConfig, pipeline: Pipeline):
...
-        else:
-            viewer_state, banner_messages = viewer_utils.setup_viewer(
-                config.viewer, log_filename=viewer_log_path, datapath=pipeline.datamanager.get_datapath()
-            )
+        viewer_state, banner_messages = viewer_utils.setup_viewer(
+            config.viewer,
+            log_filename=viewer_log_path,
+            datapath=pipeline.datamanager.get_datapath()
+            if pipeline.datamanager.get_datapath() is not None
+            else base_dir,
```
tancik, line 71: Can you make this if statement non-inline

DveloperY0115: Fixed. Thanks.

tancik (commented):
LGTM

PR #1779, opened 2023-04-20
Configurable ip address

tancik:
Change default address back to `0.0.0.0` to match nerfstudio functionality pre #1702 and allow users to specify custom addresses.


brentyi (approved):
Looks good to me, I just renamed the config variable to make the naming more consistent. `host` also seems a bit more accurate, since the value doesn't have to be an IP address.

PR #1792, opened 2023-04-21
Use current host in landing page

machenmusik:
When running locally, and not specifying websocket URL, the landing modal showed a hardcoded viewer.nerf.studio link that will take the user away from local viewer.

This PR uses current protocol and host to fix that issue.


tancik (approved):
Lgtm, thanks for making the change, this was annoying me.

PR #1794, opened 2023-04-21
Render in threejs viewer

machenmusik:
Bring the background image under threejs control in the ViewerWindow.


machenmusik:
(Perhaps more important than threejs niceties: after this change, the viewer including background is encapsulated in a single React element, which should be simpler to modify/replace.)

tancik:
This looks fine to me. I'm curious, how does this impact performance?

machenmusik:
From my light tests, negligible so far.

tancik (approved):
LGTM

PR #1795, opened 2023-04-21
Default threejs axes

machenmusik:
(Trying to unwind why the up axis was being set to backward (0, 0, 1) instead of default up (0, 1, 0), which was apparently causing viewer camera default to differ from render camera default.)


machenmusik:
(Closing for now, no point in running branch checks yet)

PR #1799, opened 2023-04-21
force resolutions to newer versions for security

### LINKED ISSUE #1773: Address Dependabot security warnings
It appears Dependabot has flagged some packages as requiring security updates.

machenmusik:
Fixes #1773 - so far, Dependabot is clean after this.
But needs broader testing of viewer functionality.


tancik (approved):
LGTM, viewer still works for me

PR #1761, opened 2023-04-18
Add PyTorch profiler to support tracing of forward/backward calls

jkulhanek:
This PR adds the support for PyTorch profiler (https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html)
It also changes the `time_function` decorator to act as a context manager - enabling blocks of code to be timed as well.

Trace example:
image

PR #1812, opened 2023-04-25
Fix profiler for PyTorch 2.0

jkulhanek:
Fixes bug mentioned here: https://github.com/nerfstudio-project/nerfstudio/pull/1761#issuecomment-1521174215

### DIFF
--- a/nerfstudio/utils/profiler.py
+++ b/nerfstudio/utils/profiler.py
@@ -21,12 +21,11 @@ import functools
 import os
 import time
 from collections import deque
-from contextlib import contextmanager
+from contextlib import ContextDecorator, contextmanager
 from pathlib import Path
 from typing import Callable, Dict, List, Optional, Tuple, Union
 
 from rich.console import Console
-from torch.autograd.profiler import ContextDecorator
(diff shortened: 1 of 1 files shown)


nerfstudio/utils/profiler.py
```diff
@@ -46,7 +45,7 @@
 class time_function(ContextDecorator):  # pylint: disable=invalid-name
     """Decorator/Context manager: time a function call or a block of code"""
 
-    def __init__(self, name: str):
+    def __init__(self, name):
```
tancik, line 48: Why was the type annotation removed?

jkulhanek: In VSCode the "@time_function" decorator was underlined as typing error as it didn't consider the "__new__" function I guess. I can also use the signature "str | Callable" for __init__ where the "Callable" won't ever be used. I think this would also fix the VSCode errors. What do you think? 

...I really don't know why the __new__ signature is not picked up.

tancik: Yea, I would go with that

tancik (approved):
LGTM

PR #1818, opened 2023-04-25
force resolution of yaml for security

machenmusik:
Resolves dependabot security warning about yaml package.


machenmusik:
Tested using localhost viewer, which works fine.

tancik (approved):
LGTM

Issue #1747, opened 2023-04-17
Any plan to improve speed of image masks in Nerfacto ?


Tao-11-chen:
Hello, i noticed that the current implementation of masking is inefficient.
In my use case, the situation is even worse, it may take more than three days to train a Nerfacto.
Are there any plans to improve speed of image masks? 
If no, is there any suggestions for me to make it faster in addition to using more threads to load data batch?
(I have ample memory but a relatively weak CPU)
Here is the running time of my program before and after using masks:

w/o mask:
CacheDataloader._get_batch_list: 0.2384
Trainer.train_iteration: 0.0339
VanillaPipeline.get_train_loss_dict: 0.0209

w/ mask:
CacheDataloader._get_batch_list: 0.6013
Trainer.train_iteration: 0.6127
VanillaPipeline.get_train_loss_dict: 0.5986

machenmusik:
Related to #1465 ?

Tao-11-chen:
Make pixel_sampler with masks run in parallel can be a help, i'll create a PR later.

Tao-11-chen:
Solved in new version.

PR #1819, opened 2023-04-25
fix_mask

Tao-11-chen:
Using masks to avoid sampling pixels on masked areas take times, so we should put masks to CUDA.
But this can lead to using indices on CUDA to query image pixels on CPU. 
Instead of putting all the images to CUDA, transform the indices tensor to CPU is more efficient.

### DIFF
--- a/nerfstudio/data/pixel_samplers.py
+++ b/nerfstudio/data/pixel_samplers.py
@@ -96,6 +96,7 @@ class PixelSampler:  # pylint: disable=too-few-public-methods
             indices = self.sample_method(num_rays_per_batch, num_images, image_height, image_width, device=device)
 
         c, y, x = (i.flatten() for i in torch.split(indices, 1, dim=-1))
+        c, y, x = c.cpu(), y.cpu(), x.cpu()
         collated_batch = {
             key: value[c, y, x] for key, value in batch.items() if key != "image_idx" and value is not None
         }


Tao-11-chen:
If it is right, should revert #1741 before merging.

Tao-11-chen:
In no case should all the images put to GPU cus masks are often used to deal with dynamic objects of outdoor scenes. These scenes are usually big and contain large images.

nepfaff:
Could you produce some results comparing this approach's training speed and memory consumption compared to https://github.com/nerfstudio-project/nerfstudio/pull/1741? Let me know if you need a small dataset with masks.

Tao-11-chen:
You could refer to #1747 for speed test and analysis. I'll give a more detailed result later.

machenmusik:
It would help if we could make available a larger dataset with masks for which this PR is tailored, so we could run some comparisons.

We should be able to determine size of images / datasets, if we have to choose behavior accordingly...

machenmusik:
So far, with https://github.com/machenmusik/nerfstudio/tree/mask2 which includes reverting #1741, I am seeing similar results with small `mustard` dataset using nerfacto, and perhaps some small improvement since not uploading images.  @nepfaff I believe you were using instant-ngp, can you confirm similar results?

machenmusik:
> On a side note, we would not want to revert all of https://github.com/nerfstudio-project/nerfstudio/pull/1741 as it also added masking capability to the instant-npg dataparser. I added them in the same PR as masks were broken, so adding them to a dataparser without fixing them did not make sense.

@nepfaff perhaps you can craft a PR that only reverts the images-to-GPU portion of #1741, so we can merge, and then merge this?

Tao-11-chen:
Thanks very much for testing it, glad to hear that it makes scense. In my situation, putting all the images to GPU causes OOM on RTX3090 untill i set image_batch_size to less than 100 out of 4000, which makes training slow with much of the time spent on I/O.

machenmusik (approved):
Thanks!

PR #1821, opened 2023-04-26
Revert loading entire dataset to GPU if using masks

nepfaff:
Reverting changes in #1741 as discussed in #1819.


machenmusik (approved):
LGTM, thanks!

Issue #1610, opened 2023-03-15
pytorch 2.0 no longer supports six


julianrendell:
First installation, trying to run 

```
ns-download-data nerfstudio --capture-name=poster
```

resulted in:
```
ImportError: cannot import name 'string_classes' from 'torch' (xxxx\nerfstudio\lib\site-packages\torch\__init__.py)
```

Found that `six` is now deprecated in pytorch (https://github.com/pytorch/pytorch/pull/94709.)

I edited `nerfstudio\nerfstudio\data\utils\nerfstudio_collate.py`
and removed `from torch._six import string_classes` and replaced the usage of `string_classes` with `str` in line 121.

I'm not sure if this is the correct fix, but the script now starts- but now crashes with  

```
.conda\envs\nerfstudio\lib\site-packages\torch\cuda\amp\grad_scaler.py", line 368, in step
    assert len(optimizer_state["found_inf_per_device"]) > 0, "No inf checks were recorded for this optimizer."
AssertionError: No inf checks were recorded for this optimizer.
```

julianrendell:
Note; hit this issue following the instructions for the Torch 1.13.1 with CUDA 11.7
dependencies. I think the installation instructions should read 
```pip install torch==1.13.1 torchvision functorch --extra-index-url https://download.pytorch.org/whl/cu117```

instead of:

```pip install torch torchvision functorch --extra-index-url https://download.pytorch.org/whl/cu117```

julianrendell:
Verified that adding  the explicit torch version gets things working (for me). Minor doc change: https://github.com/nerfstudio-project/nerfstudio/pull/1611

ksnzh:
same issue https://github.com/nerfstudio-project/nerfstudio/issues/1613

tancik:
Set the requirements to be < 2, should avoid this issue going forward.

PR #1830, opened 2023-04-27
[WIP] PyTorch 2.0 and torch.compile support

aiddun:
I'm looking into supporting PyTorch >=2.0 to see if the new compiler has any potential speedups - I've found what I think is the root cause of the bug preventing previous upgrade attempts and am seeking advice on the right fix. 

A naive upgrade leads to the same inf checking error in #1610. I believe I've narrowed down the bug to [this line in ray_samplers.py](https://github.com/nerfstudio-project/nerfstudio/blob/3f43e0d4c3885e51f3201cb72fc4cdb262466a29/nerfstudio/model_components/ray_samplers.py#L574), where after the tenth step, pytorch will no longer calculate gradients for the proposal networks, leading to the error. 

My current commit here avoids the error and continues to train by changing `self._steps_since_update > self.update_sched(self._step)` to `self._steps_since_update >= self.update_sched(self._step)` - for the example nerfacto poster command, this clause formerly always resolved to false (1 < 1). I don't know enough about pytorch's inf checks, why the updated flag is the way it currently is in ray_samplers.py, or why upgrading to 2.0 broke this line's usefulness to know if this is a proper solution, however I think it's a start if anyone would like to contribute.

### DIFF
--- a/nerfstudio/engine/schedulers.py
+++ b/nerfstudio/engine/schedulers.py
@@ -109,7 +109,7 @@ class ExponentialDecayScheduler(Scheduler):
 
     config: ExponentialDecaySchedulerConfig
 
-    def get_scheduler(self, optimizer: Optimizer, lr_init: float) -> lr_scheduler._LRScheduler:
+    def get_scheduler(self, optimizer: Optimizer, lr_init: float) -> lr_scheduler.LRScheduler:
         if self.config.lr_final is None:
             lr_final = lr_init
         else:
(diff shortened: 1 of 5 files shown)


jkulhanek:
There is another problem with PyTorch 2.0 #1883

SauravMaheshkar:
There is also work being done in https://github.com/nerfstudio-project/nerfstudio/pull/1922

jkulhanek:
The reason why it was broken after upgrading to torch 2.0 is the following:
In the nerfacto model the proposal network does not get updated every training step. It gets updated first 10 steps and then it is driven by a schedule. Changing the strictness of the equality check caused to be updated every iteration - which is wrong. But that was the reason why your code was working in torch 2.0.
The way the proposal network gets updated / or doesn't get updated is done by enabling/disabling gradients to flow through it. Unfortunately, grad-scaler is called on all the optimizers (including the one handling the proposal network) without checking if the gradients were computed. This worked fine in torch 1.13, but doesn't in 2.0. The reason is this:
https://github.com/pytorch/pytorch/pull/92731
... in torch 2.0 grads are reset to None instead of zeros to save memory (which is a breaking change).

In the alternative PR: #1922 the problem is fixed by  skipping grad_scaler.step for optimizers for which the grads weren't computed.

#1922 (for now) provides a compatibility with previous torch and addresses the same problem as this PR (except it does not include model.compile).

I suggest we close this PR?

jkulhanek:
Completer in #1922

PR #1834, opened 2023-04-28
fix video paths

takeru:
Fixed an issue where creating a custom camera path in the Viewer to create an mp4 would produce incorrect results.

### DIFF
--- a/nerfstudio/cameras/camera_utils.py
+++ b/nerfstudio/cameras/camera_utils.py
@@ -181,7 +181,7 @@ def get_interpolated_poses(pose_a: ArrayLike, pose_b: ArrayLike, steps: int = 10
         pose = np.identity(4)
         pose[:3, :3] = quaternion_matrix(quat)[:3, :3]
         pose[:3, 3] = tran
-        poses_ab.append(pose[:3])
+        poses_ab.append(pose)
     return poses_ab
 
 


tancik (approved):
LGTM. Could you provide some context on where you are running into this bug. It currently looks like this code is never actually used.

takeru:
https://discord.com/channels/1025504971606724712/1100421162317656074

Here it is.

I added a few cameras in the Viewer, output json, ran the command, and created a video file and got a funny viewpoint.

I thought there must be a bug somewhere in the recent changes, so I was looking for something related to the recent changes and found something suspicious.

tancik:
> https://discord.com/channels/1025504971606724712/1100421162317656074
> 
> Here it is.
> 
> I added a few cameras in the Viewer, output json, ran the command, and created a video file and got a funny viewpoint.
> 
> I thought there must be a bug somewhere in the recent changes, so I was looking for something related to the recent changes and found something suspicious. スクリーンショット 2023-04-29 21 34 40

I'm still a bit confused. The `get_interpolated_camera_path` is used in the render script, but only if you are not using a path generated from the viewer. Can you post the render command you are using and a stacktrace of the error?

takeru:
I checked carefully and it was a misunderstanding.

The original problem is the same as described in the DISCORD, which is that it is wrong when you create a video with a crop'd PATH.

takeru:
https://github.com/nerfstudio-project/nerfstudio/pull/1844

Issue #1835, opened 2023-04-28
Viewer crashes when using instant-ngp-bounded method


salykovaa:
**Describe the bug**
I use the latest version of nerfstudio (built from source). During training with `instant-ngp-bounded` VRAM consumption is normal (around 4GB). However if I use the viewer, more than 24GB of VRAM is needed, then I get CUDA out of memory error and the viewer crashes, **but the training continues.**

**To Reproduce**
1. `ns-train instant-ngp-bounded --data data/nerfstudio/poster/`
2. Open the viewer

**Screenshots**
![image](https://user-images.githubusercontent.com/63703454/235143499-606c1aca-db59-482e-a3ff-842449ac5000.png)

tancik:
@liruilong940607 is this related to the recent nerfacc bump to 0.5.2?

salykovaa:
Hi @tancik @liruilong940607 

Also I noticed that if I use `instant-ngp` method (**not** `instant-ngp-bounded`), training requires around 5GB of VRAM, but when I use the viewer, it suddenly jumps up to 15-17GB, but the viewer doesn't crash. I would assume all nerfacc-based methods have this problem.

liruilong940607:
Ah so I looked into it and this OOM is because of the [`cone_angle=0.004` -> `cone_angle=0.0`](https://github.com/nerfstudio-project/nerfstudio/pull/1809/files#diff-40f3d01f8379dea14417702e7b8a1a65e6efbb1e0d928262432d01abbe184b77R267) in the PR #1809 for `instant-ngp-bounded`. This leads to much more samples for `instant-ngp-bounded` than before. 

The reason for changing this default behavior is to match the NGP paper on how they train the bounded nerf-synthetic data.

For `instant-ngp` model, I reverted to pre-#1809 and the behavior is the same (15GB with viewer opened).

But note that the ultimate reason for such high GPU consumption with NGP-based methods, is NGP-based methods densely sample along the ray (~1000 or more samples per ray) and gradually reduce the number of samples by pruning the space during training. So in the early stage of the training, rendering is slow and memory intense because of the large amount of samples. But as the training goes the situation would become much much better. (e.g., If you open the viewer after like 20% of training, you wouldn't see OOM)

[...]

PR #1837, opened 2023-04-28
Reduce the GPU memory required in viewer for NGP methods

liruilong940607:
Fix the issue #1835:

- change the default `viewer.num_rays_per_chunk` from `1 << 15` to `1 << 12`.
- remove the `render_step_size=0.001` for `instant-ngp-bounded` so that it is automatically calculated based on the scene aabb (target on roughly 1000 samples). Preset `render_step_size` would lead to tons of samples if the scene aabb is large (or in different unit).


tancik (approved):
LGTM

PR #1838, opened 2023-04-28
`torchtyping` => `jaxtyping`

brentyi:
Draft at migrating from `torchtyping` to `jaxtyping`. This will fix static type checking.

Some notes:
- This will require bumping the minimum Python version to 3.8.
- I'm currently using `jaxtyping.Shaped[]` everywhere, but we should ideally be using `jaxtyping.Float[]`, `jaxtyping.Int[]`, etc.

cc @tancik


_torchtyping_to_jaxtyping.py
tancik, line 1: Should we keep this around in the scripts folder if others want to update their code?

brentyi: Nah let's not, I think it doesn't add much because changes still require so much manual labor

tancik (approved):
LGTM

brentyi:
Conversion script we used, for posterity: https://gist.github.com/brentyi/21d08fe4e90e0397b9484893cf1b9822

brentyi:
🎊 🎊 🎊 🎊 🎊 

cc @patrick-kidger thanks for adding torch support to your library :)

patrick-kidger:
Thank you for the tag! This is really great to see. :)

PR #1841, opened 2023-04-30
Support equirectangular sensor with metashape

f-dy:
fixes https://github.com/nerfstudio-project/nerfstudio/issues/1634 (see sample runs in that issue)

The COLMAP logic had to be changed slightly, as mentioned here: https://github.com/nerfstudio-project/nerfstudio/issues/1634#issuecomment-1529152231 : after computing the planar projections, the ns-process-data scripts now set the camera type to perspective (which makes sense, since all the rest of the processing is on perspective cameras).

### DIFF
--- a/docs/quickstart/custom_dataset.md
+++ b/docs/quickstart/custom_dataset.md
@@ -259,6 +259,18 @@ ns-train nerfacto --data {output directory}
 
 ## Metashape
 
+All images must use the same sensor type. If the images are using an equirectangular/spherical sensor, first convert those to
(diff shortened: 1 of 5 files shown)


f-dy:
Should I run `black`, which would change many more lines than what I touched?

tancik:
Run `ns-dev-test`, make sure to install the correct packages with `pip install -e .[dev]`

tancik (approved):
LGTM, thanks for fixing

PR #1844, opened 2023-05-01
fix crop_center

takeru:
The crop_center expression is wrong and has been corrected.
This fixes a problem with video files created from path.json generated with Crop set.

https://discord.com/channels/1025504971606724712/1100421162317656074


tancik (approved):
LGTM

PR #1851, opened 2023-05-02
Fix typos in Ray Samplers docs

MvWouden:
I fixed some typo's in the Ray Samplers docs, including:
- "... if the **sene** can be bounded ..." -> "... if the **scene** can be bounded ..."
- "... as **the the** eye can see ..." -> "... as **the** eye can see ..."
- "**Unfortunatly** ..." -> "**Unfortunately,** each ..."
- "... it is an arbitrary **funtion** ..." -> "... it is an arbitrary **function** ..."
- "... the **overal** distribution ..." -> "... the **overall** distribution ..."


tancik (approved):
LGTM

PR #1853, opened 2023-05-02
Fix images to nerfstudio filename

MvWouden:
I noticed that a filename had a typo, so I fixed it.
I also found a typo in the docs somewhere.


tancik (approved):
LGTM

Issue #1847, opened 2023-05-02
DistributedDataParallel's wrapped model was never used in training


jb-ye:
**Describe the bug**

In pipeline implementation, self.model refers to the self._model.module in multi-gpu setting. In fact, to support averaging of gradient across multiple gpus, self._model (DDP model) should be used instead of self._model.module (regular model). The current implementation will not average gradients across multiple gpus. See lines below:

https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/pipelines/base_pipeline.py#L161
https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/pipelines/base_pipeline.py#L278

**To Reproduce**
Training with batch size N with 2 gpu is currently generating the same behavior as training with batch size N with one gpu. 

**Expected behavior**
Training with batch size N with 2 gpu should generate the same behavior as training with batch size 2*N with one gpu.

tancik:
Interesting, good find! Would you be willing to make a PR with the requested change? I only have a single GPU atm so it's difficult for me to test.

jb-ye:
Sure, I can help with that.

PR #1856, opened 2023-05-02
Fix bug related to distributed data parallel training

jb-ye:
Address issue #1847


tancik:
If you run `ns-dev-test` it will fix the black errors. (Install the dev dependencies with `pip install -e .[dev]`)

ethanweber:
Hi @jb-ye, I'm not seeing a change when I run it on my end with 1 vs. 4 GPUs. I'm including the main branch in this plot as well. Do you see something differently? If so, could you share with us? Here I'm running `ns-train mipnerf --data data/blender/lego --machine.num-gpus 4 --vis wandb blender-data` vs `ns-train mipnerf --data data/blender/lego --machine.num-gpus 1 --vis wandb blender-data`.

image

jb-ye:
@ethanweber I realized you are plotting against loss vs. training time. My local experiment shows that the benefits of parallel saturate at two GPUs for most of my experiments. Because 4 GPUs, the communication overhead is too much.

jb-ye:
@tancik Do you expect to merge this change?

ethanweber:
Should [this line](https://github.com/nerfstudio-project/nerfstudio/blob/2f02479a2679c3212233bf9e2d9623766cbcab2c/nerfstudio/pipelines/base_pipeline.py#L142) be changed too? I don't think it's called anywhere right now but seems like the same issue. Before you merge, could you add an example in the docs along with vanilla nerf that uses nerfacto and trains faster with multiple GPUs? I think it would be great to get more nerfacto uses using multi GPU if it's working now.

jb-ye:
> Should [this line](https://github.com/nerfstudio-project/nerfstudio/blob/2f02479a2679c3212233bf9e2d9623766cbcab2c/nerfstudio/pipelines/base_pipeline.py#L142) be changed too? I don't think it's called anywhere right now but seems like the same issue. Before you merge, could you add an example command in the docs along with vanilla nerf that uses nerfacto and trains faster with multiple GPUs? I think it would be great to get more nerfacto uses using multi GPU if it's working now.

I noticed this line as well, but for class Pipeline, self._model is never instantiated in its constructor. So I thought it doesn't have self._model and doesn't support multi-gpu by design.

ethanweber (commented):
I just made these changes, and I think things look good to go! Thanks for the PR and documenting it!

jb-ye:
Looks good to me!

ethanweber (approved):

Issue #1724, opened 2023-04-12
Render depth video


ziiho08:
Hello,

Rendering a depth video with ns-render returns a grayscale video. Is there a way to get a colormapped depth video?

Thanks

tancik:
That logic currently does not exist in our render script.

filsus:
I am also to eager to have this option.

tancik:
Fixed in #1893

PR #1858, opened 2023-05-03
Specify Near/Far Plane for Render Depth

jaidevshriram:
Added the option to specify the near and far plane for depth rendering in addition to adding a colormap for the depth. Does this by calling `apply_depth_colormap` from `utils` along with new CLI options. Would address #1724.


tancik:
Closing since #1893 was merged which has a superset of this functionality.

PR #1868, opened 2023-05-04
Make VanillaDataManager generic

jkulhanek:
This PR makes VanillaDataManager a generic class which reduces the amount of boilerplate code when implementing different datasets:
- The VanillaDataManager now takes the Dataset type as a generic parameter
- Default generic parameter value is InputDataset (to preserve backward compatibility)
- Removes DepthDataManager, SemanticDataManager, SDFDataManager (VanillaDataManager is used for everything, Dataset type is specified)
- Added tests to ensure the correct dataset types are instantiated


jkulhanek:
I tested nerfacto and neus, both seems to work correctly.

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -426,22 +440,44 @@ def __init__(
 
         super().__init__()
 
-    def create_train_dataset(self) -> InputDataset:
+    def get_dataset_type(self) -> Type[TDataset]:
+        """Returns the dataset type passed as the generic argument"""
+        orig_class = get_orig_class(self, default_to__class__=True)
+        if orig_class is VanillaDataManager:
+            return TDataset.__default__  # typing: ignore
+        if hasattr(orig_class, "__origin__") and orig_class.__origin__ is VanillaDataManager:
```
brentyi, line 448: can we replace the manual access of `__origin__` and `__args__` with `typing.get_origin()` and `typing.get_args()`? or their backports in `typing_extensions`; I imagine it's the same in terms of functionality but might be more succinct + look less intimidating 🙂

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -426,22 +440,44 @@ def __init__(
...
+            return orig_class.__args__[0]
+        # For inherited classes, we need to find the correct type to instantiate
+        for base in getattr(self, "__orig_bases__", []):
+            if hasattr(base, "__origin__") and base.__origin__ is VanillaDataManager:
+                for value in base.__args__:
+                    if isinstance(value, ForwardRef):
```
brentyi, line 454: do we actually use forward refs, or is this just being thorough? if it's the latter I would vote to just add an `assert not isinstance(value, ForwardRef)` to keep things simple

nerfstudio/utils/misc.py
```diff
@@ -155,3 +166,87 @@ def strtobool(val) -> bool:
     FMI https://stackoverflow.com/a/715468
     """
     return val.lower() in ("yes", "y", "true", "t", "on", "1")
+
```
brentyi, line 169: most of the changes in this file can be dropped if we move to Python 3.8, right?

I spoke to @tancik and it seems like we can just bump the minimum version, it'll help with #1838 too

brentyi (commented):
this looks great to me, just a few questions!

nerfstudio/data/datamanagers/base_datamanager.py
jkulhanek: Currently the code does not use ForwardRefs. I will drop the support.

nerfstudio/utils/misc.py
jkulhanek: Dropping support for 3.7 wouldn't help much here. The problem is that __orig_class__ used to be populated before __init__ was called in 3.7, but in 3.8+ it is populated after the __init__ call. We need to access it from the __init__ so I used this workaround.

brentyi: Oops, yeah actually that makes a lot of sense!

To me this seems like what `__class_getitem__()` was designed for: https://docs.python.org/3/reference/datamodel.html#object.__class_getitem__

Incomplete, but as an alternative suggestion what do you think about a pattern that looks like this?


```python
from typing import Any, Generic, TypeVar

from typing_extensions import override

TDataset = TypeVar("TDataset")


class VanillaDataManager(Generic[TDataset]):
    def __init__(self) -> None:
        print(self.get_dataset_type())

    def get_dataset_type(self) -> Any:
        return "the default dataset type"

    def __class_getitem__(cls, key):
        class DataManager(VanillaDataManager):
            @override
            def get_dataset_type(self) -> Any:
                return key

        return DataManager


# Prints "int"
VanillaDataManager[int]()
```

nerfstudio/data/datamanagers/base_datamanager.py
jkulhanek: I have imported them in a new nerfstudio.utils.typing module. Do you agree? After dropping 3.7 we can import directly from typing.

nerfstudio/utils/misc.py
jkulhanek: I am not as experienced with how type inference is implemented in various tools, but perhaps returning a dynamically defined class may be problematic and less intuitive for users? Otherwise the pattern looks good and much simpler compared to my code.

jkulhanek: What would happen when you serialise the type with pyyaml (in TrainerConfig). Is it able to serialise and construct a dynamic type?

jkulhanek:
I have updated the code, the question is now if we want to use current code with "__orig_class__" or switch to "__class_getitem__"

nerfstudio/utils/misc.py
jkulhanek: The following code (used in storing and loading configs) does not work with dynamic types:
```
string = yaml.dump(VanillaDataManager[int]())
obj = yaml.load(string, Loader=yaml.UnsafeLoader)
reveal_type(obj)
```

brentyi: Nice catch! Maybe we can use `functools.partial`?

```python
import functools
from typing import Generic, cast

from typing_extensions import TypeVar, reveal_type

TDataset = TypeVar("TDataset", default=str)


class VanillaDataManager(Generic[TDataset]):
    def __init__(self, **kwargs) -> None:
        self.dataset_type = cast(TDataset, kwargs.get("dataset_type", str))
        print(self.dataset_type)

    def __class_getitem__(cls, key):
        return functools.partial(VanillaDataManager, dataset_type=key)


# prints "int"
x = VanillaDataManager[int]()

# type is revealed as Type[int]
reveal_type(x.dataset_type)


# prints "str"
y = VanillaDataManager()

# type is revealed as Type[str]
reveal_type(y.dataset_type)
```

jkulhanek: The object obtained after loading the yaml will be different from the original (not generic), but I guess it should be fine as long as we do not access __orig_class__ or similar.

jkulhanek: There is one more problem:
```
class tmp(VanillaDataManager[int]):
    pass
x = tmp()
print(x.dataset_type)
```
The following does not work (since the object returned from __class_getitem__ is not a class I guess?)

jkulhanek: But I can use this awful thing:
```
    def __class_getitem__(cls, key):
        return type(cls.__name__, (cls,), {"__module__": cls.__module__, "__init__": functools.partialmethod(cls.__init__, dataset_type=key)})
```

jkulhanek: I have pushed a new version which  uses this.

nerfstudio/data/datamanagers/base_datamanager.py
brentyi: Thanks!

nerfstudio/data/datamanagers/base_datamanager.py
brentyi: Looks good. If we get an OK from @tancik we could also expediting merging for the 3.8 requirement

nerfstudio/utils/misc.py
brentyi: Certainly awful, but it has my support 🙂

jkulhanek:
Can we merge this?

brentyi (approved):

brentyi:
Done, thanks for the bump @jkulhanek!

jkulhanek:
Thank you!

jkulhanek:
@assafbot, can you please try if you still experience the error with #2129 version?

PR #1869, opened 2023-05-04
Enable method config registration from env variable

jkulhanek:
Adds a new environment variable, "NERFSTUDIO_METHOD_CONFIGS", which enables registering custom methods without installing them as pip packages.

Tests added and docs updated.

Partially also addresses #1622.


tancik (approved):
LGTM, great idea

PR #1876, opened 2023-05-05
Bump minimum Python version to 3.8

brentyi:
Will help with #1868, #1838

cc @jkulhanek


brentyi:
For better or worse there are a lot of line changes from running an updated version of `black`, I can (manually) revert these if anybody feels strongly about keeping the diff clean...!

jkulhanek:
Not needed for #1868 anymore.

tancik (approved):
LGTM

PR #1879, opened 2023-05-06
Wandb/experiment improvements

ethanweber:
This PR fixes a consistent ordering for eval images when using a random seed. It also gives functionality for setting the wandb project name, setting the experiment name, and loading from a checkpoint path.


jkulhanek:
So wandb environment variables are no longer supported? 
https://docs.wandb.ai/guides/track/environment-variables

ethanweber:
@jkulhanek I'll fix that right now!

ethanweber:
If you want to double check at PR #1880, that would be great. Thanks!

Tao-11-chen:
Hi, I'm suffering from getting Null metrics_dict while the training step reached get_average_eval_image_metrics at the second time after this change.

ethanweber:
Thanks for fixing it!

PR #1880, opened 2023-05-06
fixing wandb envs

ethanweber:


nerfstudio/utils/writer.py
```diff
@@ -288,6 +289,12 @@ class WandbWriter(Writer):
 
     def __init__(self, log_dir: Path, experiment_name: str, project_name: str = "nerfstudio-project"):
         wandb.init(project=project_name, dir=str(log_dir), reinit=True, name=experiment_name)
```
jkulhanek, line 291: You call wandb.init twice here.

jkulhanek (changes requested):

jkulhanek (approved):

PR #1882, opened 2023-05-06
Avoids running test_train.py on external methods

ethanweber:
- Removes running tests for test_train.py (for a dry-run) when someone implements an external method.
- Adds a short description for the neus-facto method.


tancik (approved):
LGTM

Issue #1813, opened 2023-04-25
The latest version does not have viewer.start-train option


Barry-Zhou:
The latest version does not have viewer.start-train option,If I do not want it to train, how to do it?

tancik:
Use `ns-viewer` instead.

PR #1889, opened 2023-05-09
Fix start_train option

j3soon:
The `--viewer.start-train False` option is used in the docs but not available due to regressions in #1702.

This commit simply adds back the option.

Related: #1813

### DIFF
--- a/nerfstudio/configs/base_config.py
+++ b/nerfstudio/configs/base_config.py
@@ -131,6 +131,9 @@ class ViewerConfig(PrintableConfig):
 
     relative_log_filename: str = "viewer_log_filename.txt"
     """Filename to use for the log file."""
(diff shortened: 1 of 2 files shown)


tancik:
Out of curiosity, what is your use-case for this functionality? Is `ns-viewer` sufficient?

j3soon:
I haven't tried out `ns-viewer` yet, but I believe it is sufficient.

Since this option is only mentioned in the docs twice: https://github.com/search?q=repo%3Anerfstudio-project%2Fnerfstudio%20viewer.start-train&type=code,

another fix is to simply modify the docs to use `ns-viewer` instead.

tancik:
Closing in favor of #1892

PR #1891, opened 2023-05-09
Fix mask bug

Tao-11-chen:
In my case, mask here is a Something like this: [[True], [True], [False]....].
Using "if mask" will cause this issue:
RuntimeError: Boolean value of Tensor with more than one value is ambiguous.


tancik (approved):
LGTM

PR #1900, opened 2023-05-10
FIx random indices bug

Tao-11-chen:
Related to #1879
I'm suffering from getting Null metrics_dict while the training step reached get_average_eval_image_metrics at the second time after this change.
This code is not well tested in the newest version but it solved my problem above.
Merge this if it works.


wangmingyang4:
The Null metrics_dict still exists

PR #1901, opened 2023-05-10
Implement gradient scaling

jkulhanek:
This PR implements gradient scaling from [Radiance Field Gradient Scaling for Unbiased Near-Camera Training](https://gradient-scaling.github.io/)

GradientScaling is added to the nerfacto model and enabled using flag `--use-gradient-scaling=True`

### DIFF
--- a/nerfstudio/model_components/losses.py
+++ b/nerfstudio/model_components/losses.py
@@ -513,17 +513,18 @@ class GradientScaler(torch.autograd.Function):  # typing: ignore, pylint: disabl
     """
 
     @staticmethod
-    def forward(ctx, field_outputs, ray_dist):  # pylint: disable=arguments-differ
+    def forward(ctx, field_outputs, ray_samples):  # pylint: disable=arguments-differ
+        ray_dist = ray_samples.frustums.starts + ray_samples.frustums.ends / 2
         ctx.save_for_backward(ray_dist)
         return field_outputs, ray_dist
 
     @staticmethod
-    def backward(ctx, grad_field_outputs, grad_output_ray_dist):  # pylint: disable=arguments-differ
+    def backward(ctx, grad_field_outputs, grad_output_ray_samples):  # pylint: disable=arguments-differ
         (ray_dist,) = ctx.saved_tensors
         scaling = torch.square(ray_dist).clamp(0, 1)
         output_grad = dict(grad_field_outputs.items())
         for key in output_grad.keys():
             output_grad[key] = output_grad[key] * scaling
-        return output_grad, grad_output_ray_dist
+        return output_grad, grad_output_ray_samples
 
     vjp = backward
--- a/nerfstudio/model_components/losses.py
+++ b/nerfstudio/model_components/losses.py
@@ -513,12 +513,15 @@ class GradientScaler(torch.autograd.Function):
     """
 
     @staticmethod
-    def forward(ctx, colors, sigmas, ray_dist):
+    def forward(ctx, field_outputs, ray_dist):
         ctx.save_for_backward(ray_dist)
-        return colors, sigmas, ray_dist
+        return field_outputs, ray_dist
 
     @staticmethod
-    def backward(ctx, grad_output_colors, grad_output_sigmas, grad_output_ray_dist):
+    def backward(ctx, grad_field_outputs, grad_output_ray_dist):
         (ray_dist,) = ctx.saved_tensors
         scaling = torch.square(ray_dist).clamp(0, 1)
-        return grad_output_colors * scaling, grad_output_sigmas * scaling, grad_output_ray_dist
+        output_grad = dict(grad_field_outputs.items())
+        for key in output_grad.keys():
+            output_grad[key] = output_grad[key] * scaling
+        return output_grad, grad_output_ray_dist
--- a/nerfstudio/model_components/losses.py
+++ b/nerfstudio/model_components/losses.py
@@ -504,7 +504,7 @@ def tv_loss(grids: TensorType["grids", "feature_dim", "row", "column"]) -> Tenso
     return 2 * (h_tv / h_tv_count + w_tv / w_tv_count) / number_of_grids
 
 
-class GradientScaler(torch.autograd.Function):
+class GradientScaler(torch.autograd.Function):  # typing: ignore, pylint: disable=abstract-method
     """
     Scale gradients by the ray distance to the pixel
     as suggested in `Radiance Field Gradient Scaling for Unbiased Near-Camera Training` paper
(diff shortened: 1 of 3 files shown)


jkulhanek:
I am not sure it helps:
image

Here are the full runs, please take a look:
https://wandb.ai/kulhanek/nerfstudio-baselines?workspace=user-kulhanek

nerfstudio/model_components/losses.py
```diff
@@ -502,3 +502,28 @@ def tv_loss(grids: TensorType["grids", "feature_dim", "row", "column"]) -> Tenso
...
+class GradientScaler(torch.autograd.Function):  # typing: ignore, pylint: disable=abstract-method
+    """
+    Scale gradients by the ray distance to the pixel
+    as suggested in `Radiance Field Gradient Scaling for Unbiased Near-Camera Training` paper
+
+    Note, the scaling is applied on the interval of [0, 1] along the ray!
+    """
+
+    @staticmethod
+    def forward(ctx, field_outputs, ray_dist):  # pylint: disable=arguments-differ
```
tancik, line 516: What about inputing `RaySamples` instead of ray_dist

nerfstudio/model_components/losses.py
```diff
@@ -502,3 +502,29 @@ def tv_loss(grids: TensorType["grids", "feature_dim", "row", "column"]) -> Tenso
...
     w_tv = torch.pow((grids[:, :, :, 1:] - grids[:, :, :, :-1]), 2).sum()
     return 2 * (h_tv / h_tv_count + w_tv / w_tv_count) / number_of_grids
+
+
+class GradientScaler(torch.autograd.Function):  # typing: ignore, pylint: disable=abstract-method
+    """
+    Scale gradients by the ray distance to the pixel
+    as suggested in `Radiance Field Gradient Scaling for Unbiased Near-Camera Training` paper
+
+    Note, the scaling is applied on the interval of [0, 1] along the ray!
```
tancik, line 512: Add example of how to call in docstring. More info [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#:~:text=Example%3A%0A%20%20%20%20Examples%20can%20be%20given%20using%20either%20the%20%60%60Example%60%60%20or%20%60%60Examples%60%60%0A%20%20%20%20sections.%20Sections%20support%20any%20reStructuredText%20formatting%2C%20including%0A%20%20%20%20literal%20blocks%3A%3A%0A%0A%20%20%20%20%20%20%20%20%24%20python%20example_google.py)

nerfstudio/model_components/losses.py
```diff
@@ -502,3 +502,29 @@ def tv_loss(grids: TensorType["grids", "feature_dim", "row", "column"]) -> Tenso
...
+    """
+
+    @staticmethod
+    def forward(ctx, field_outputs, ray_samples):  # pylint: disable=arguments-differ
```
tancik, line 516: Can we type and add dosctrings? Or does it break since we are overwriting?

nerfstudio/model_components/losses.py
```diff
@@ -513,17 +513,18 @@ class GradientScaler(torch.autograd.Function):  # typing: ignore, pylint: disabl
     """
 
     @staticmethod
-    def forward(ctx, field_outputs, ray_dist):  # pylint: disable=arguments-differ
+    def forward(ctx, field_outputs, ray_samples):  # pylint: disable=arguments-differ
+        ray_dist = ray_samples.frustums.starts + ray_samples.frustums.ends / 2
```
salykovaa, line 517: should be `(ray_samples.frustums.starts + ray_samples.frustums.ends) / 2`

salykovaa:
@jkulhanek @tancik 

In the paper they set `near_plane` to 0 in all experiments. Does it make sense to do the same here?

tancik:
> @jkulhanek @tancik
> 
> In the paper they set `near_plane` to 0 in all experiments. Does it make sense to do the same here?

The near plane can be set with `--pipeline.model.near-plane`
If we set the gradient scaling on by default, then it makes sense to set the near plane to 0 by default. I'm fine with making this default, but we should test it on a few more scenes first to make sure that it better or equal to not using it.

ichsan2895:
Can't wait to be implemented in the next version of Nerfstudio 💯

ichsan2895:
> I am not sure it helps: image
> 
> Here are the full runs, please take a look: https://wandb.ai/kulhanek/nerfstudio-baselines?workspace=user-kulhanek

As stated in the paper, the PSNR will slightly change (not too much), but it more better qualitatively because less of floaters

nerfstudio/model_components/losses.py
jkulhanek: Typing and doctrings are not visible anywhere. I have added a wrapper function to provide the docstrings and typing support.

nerfstudio/model_components/losses.py
jkulhanek: Thanks!

tancik (approved):
LGTM. We should try experiment with it over the next few days. If we see better results on average, let's enable it by default.

ichsan2895:
image

image

Sorry why I can't use it?

ichsan2895:
> GradientScaling is added to the nerfacto model and enabled using flag `--use-gradient-scaling=True`

This seems works with `--pipeline.model.use_gradient_scaling True` instead of  `--use-gradient-scaling=True`
image

**Here is the show. I decreased the value of `num-ray-per-chunk` & others parameter to avoid CUDA OOM**

**Without Gradient Scale**, CHKPT 48 K, num-rays-per-chunk 16384, num-rays-per-batch 2048 , num-rays-per-batch 2048

without Grad Scaling mp4_snapshot_00 00_frame7

**With Gradient Scale**, CHKPT 48 K, num-rays-per-chunk 16384, num-rays-per-batch 2048 , num-rays-per-batch 2048

with Grad Scaling mp4_snapshot_00 00_frame7

jkulhanek:
Thanks for testing it on your scene! I guess we can optimize it to fit into the memory.

jkulhanek:
Does anyone know if gradients in autograd.function can me modified in-place?

ichsan2895:
> Thanks for testing it on your scene! I guess we can optimize it to fit into the memory.

Both test will OOM CUDA, so I think the problem of OOM is not by grad scaling. The problem of OOM CUDA come from my own potato GPU which it has low VRAM

jkulhanek:
Ok, I guess I will keep the current implementation then.

ichsan2895:
Another testing come here...

**Without Gradient Scaling**


https://github.com/nerfstudio-project/nerfstudio/assets/60387342/9d43f5f7-7a7b-4e17-98cc-98e7dc968aeb


**With Gradient Scaling**


https://github.com/nerfstudio-project/nerfstudio/assets/60387342/42f9f7ff-ee76-463f-8049-807e6d467792

PR #1905, opened 2023-05-10
Add button groups to viewer

tancik:
image

### DIFF
--- a/nerfstudio/viewer/app/src/modules/ConfigPanel/ConfigPanel.jsx
+++ b/nerfstudio/viewer/app/src/modules/ConfigPanel/ConfigPanel.jsx
@@ -55,6 +55,7 @@ function CustomLeva() {
     } else if (levaConf.type === 'BUTTON_GROUP') {
       // Add a button group.
       leafFolder[key] = buttonGroup({
+        label: levaConf.label,
         opts: levaConf.options.reduce((acc, opt) => {
           acc[opt] = () =>
             sendWebsocketMessage(viser_websocket, {
--- a/nerfstudio/viewer/app/src/modules/ConfigPanel/ConfigPanel.jsx
+++ b/nerfstudio/viewer/app/src/modules/ConfigPanel/ConfigPanel.jsx
@@ -55,8 +55,6 @@ function CustomLeva() {
     } else if (levaConf.type === 'BUTTON_GROUP') {
       // Add a button group.
       leafFolder[key] = buttonGroup({
-        label: `Train Speed`,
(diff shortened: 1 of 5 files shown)


brentyi:
I'm attempting to upstream this into `viser`, will come back to this after that's stable

nerfstudio/viewer/app/src/modules/ConfigPanel/ConfigPanel.jsx
```diff
@@ -51,6 +52,21 @@ function CustomLeva() {
           value: true,
         });
       }, levaConf.settings);
+    } else if (levaConf.type === 'BUTTON_GROUP') {
+      // Add a button group.
+      leafFolder[key] = buttonGroup({
+        label: `Train Speed`,
+        hint: 'Select the training speed, affects viewer render quality, not final render quality',
```
brentyi, line 59: @tancik does this `hint` field do anything? It's not in the TypeScript definition, which only has `label` and `opts`: https://github.com/pmndrs/leva/blob/d658bf0b3cb129e11465e7079dcaed82882eb1bc/packages/leva/src/types/public.ts#L59-L64

brentyi: also, any reason to put `Train Speed` in backticks but the hint in single quotes?

tancik: Here? https://github.com/pmndrs/leva/blob/d658bf0b3cb129e11465e7079dcaed82882eb1bc/packages/leva/src/types/public.ts#L147
No reason for different ticks

brentyi: Yeah, but the input of `buttonGroup` is just `ButtonGroupOpts`: https://github.com/pmndrs/leva/blob/d658bf0b3cb129e11465e7079dcaed82882eb1bc/packages/leva/src/helpers/buttonGroup.ts#L9-L11

And `ButtonGroupOpts` only has `label` and `opts` (as linked above)

=> I get an error when I try to assign `hint` to something:
Screenshot 2023-05-15 at 11 19 14 AM

tancik: aghh, that logic doesn't belong in the js. Fixed now.

brentyi (approved):

brentyi (approved):

PR #1906, opened 2023-05-10
fix metashape_to_json() if there is no 'components' tag in .xml

salykovaa:
If there is only 1 component, the 'components' tag is not present in the .xml file -> the variable `components` becomes `None` and not `iterable`. PR solves this problem. 

See the latest discussion started by `brycelynch#9743` in the troubleshooting channel.


tancik (approved):
LGTM, Thanks for fixing!

PR #1909, opened 2023-05-11
extract dataparser configs to a stand-alone file, support custom dataparsers

wuzirui:
I am working on a custom method that requires a customized dataparser. I followed the instructions in the docs but found out that it does not support doing this, cuz the dataparser dict is written in `base_datamanager.py`. So I followed the `registry.py` to add a `discover_dataparsers()` method.

Note that I didn't put the new methods & specification data types in the current types and registry scripts since it will cause a circular import error.


nerfstudio/plugins/registry_dataparser.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Discovers all dataparsers registered using the `nerfstudio.data_configs` entrypoint.
+    """
+    dataparsers = {}
+    discovered_entry_points = entry_points(group="nerfstudio.data_configs")
+    for name in discovered_entry_points.names:
+        specification = discovered_entry_points[name].load()
+        if not isinstance(specification, DataParserSpecification):
+            CONSOLE.print(
+                "[bold yellow]Warning: Could not entry point {n} as it is not an instance of DataParserSpecification"
```
tancik, line 55: Should be `f` string

nerfstudio/configs/data_configs.py
tancik, line 1: rename to `dataparser_configs`

tancik (commented):
Great idea, thanks for adding. Can you also update the documentation with this new functionality. It should be added [here](https://docs.nerf.studio/en/latest/developer_guides/pipelines/dataparsers.html). We should also add a note in [this](https://docs.nerf.studio/en/latest/developer_guides/new_methods.html#registering-custom-model-with-nerfstudio) section that informs developers that the similar functionality exists for dataparsers.

tancik (approved):
LGTM

Issue #1732, opened 2023-04-13
NeusFacto Custom Data Instructions


MvWouden:
**Is your feature request related to a problem? Please describe.**
Hi, I saw that NeusFacto got merged in the last day.
I tried running the model, but ended up with some issues in getting it to work on my data.
It seems it requires some pre-processing steps from sdfstudio, but I cannot find the proper documentation for this.

**Describe the solution you'd like**
A small piece of documentation or a (brief) explanation on how I can pre-process custom data to be used with the newly merged NeusFacto.

**Suggestion**
If provided with a (brief) explanation, I would be willing to document this process somewhere if that helps the project.

pablovela5620:
Hi so you are correct, currently the SDF based models (NeuS and NeuS Facto) require some pre-processing for custom data to work (as of this point they will only work with the sdfstudio-demo-data download). I had written a script for this over at the SDFStudio that you can find [here](https://github.com/autonomousvision/sdfstudio/blob/master/scripts/datasets/process_nerfstudio_to_sdfstudio.py). I plan on bringing this over and integrating it directly into ns-proccess-data (not totally clear as to how I'll do this yet). For now you're welcome to use the script over at the SDFstudio repo to get things preprocessed.

pablovela5620:
You don't need to but it helps in some cases (especially for indoor scenes), what command did you use exactly and what did the dataset/output look like? Most of my testing has been with indoor room enviroments

cian0:
Hi, how do I know what to put in pretrained_models and omnidata-path for that script? 

https://github.com/autonomousvision/sdfstudio/blob/master/scripts/datasets/process_nerfstudio_to_sdfstudio.py

MvWouden:
> Hi, how do I know what to put in pretrained_models and omnidata-path for that script?
> 
> https://github.com/autonomousvision/sdfstudio/blob/master/scripts/datasets/process_nerfstudio_to_sdfstudio.py

I believe that is only required when you want to extract mono-prior depths and normals.
If that's not needed, you can simply do:
```shell
python3 scripts/datasets/process_nerfstudio_to_sdfstudio.py --data <path/to/data/dir> --output-dir <path/to/output/dir> --data-type <colmap/polycam> --scene-type <indoor/object/unbound>
```

MvWouden:
@pablovela5620 I was able to train a Neusfacto model on custom data recently.
For this, I did not extract monocular depth and normal cues.
[...]

PR #1910, opened 2023-05-11
#1732 fix sdf dataparser no mono paths

MvWouden:
Related to issue: https://github.com/nerfstudio-project/nerfstudio/issues/1732
Related comment: https://github.com/nerfstudio-project/nerfstudio/issues/1732#issuecomment-1543654312

In summary, the [conversion script](https://github.com/autonomousvision/sdfstudio/blob/master/scripts/datasets/process_nerfstudio_to_sdfstudio.py) that is used before training Neus(facto) does not include the `mono_depth_path` and `mono_normal_path` keys for the frames by default. If monocular depth and normal cues are not extracted, some errors will be generated.

There are several ways in which this can be solved.
If another style/approach is preferred, please let me know, so I can change the PR accordingly.

### DIFF
--- a/nerfstudio/data/dataparsers/sdfstudio_dataparser.py
+++ b/nerfstudio/data/dataparsers/sdfstudio_dataparser.py
@@ -87,16 +87,16 @@ class SDFStudio(DataParser):
                 continue
 
             image_filename = self.config.data / frame["rgb_path"]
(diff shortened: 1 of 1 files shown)


tancik (approved):
LGTM

PR #1914, opened 2023-05-12
Replace python code tracing in ns viewer by explicit interrupts

SamDM:
Currently code tracing (which is usually reserved for debuggers) is done in the rendering pass of the nerfstudio viewer server (see [here](https://github.com/nerfstudio-project/nerfstudio/blob/79b06ee2bf637949a971d8cb95ab339c6d91787f/nerfstudio/viewer/server/render_state_machine.py#L205) and [here](https://github.com/nerfstudio-project/nerfstudio/blob/79b06ee2bf637949a971d8cb95ab339c6d91787f/nerfstudio/viewer/server/viewer_utils.py#L70)).

It seems like debugger functionality is abused to be able to check for an interrupt of the rendering pass after every line of python code. However, replacing the tracing function clashes with debugging tools (VSCode in my case) that also use tracing to step trough python code.

I removed code tracing, and just sprinkled a few interrupt checks in the render function of the viewer. It seems to work fine (tested on a NVIDIA GeForce RTX 2080 Ti). I don't notice any delays in the viewer.

Other alternatives (not in this PR) would be:

-  to launch the render pass in a new process, and interrupt it if needed. That may reduce the latency of an interrupt, but increase overhead due to multiprocessing. Not sure if this will result in a noticeable UX improvement.
- keep using code tracing as an interrupt mechanism, but don't throw out the existing debugger. And on every line, first execute the behavior of the existing debugger, and only then interrupt. And finally restore the original debugger after the render pass is done.

### DIFF
--- a/nerfstudio/scripts/render.py
+++ b/nerfstudio/scripts/render.py
@@ -71,11 +71,7 @@ def _render_trajectory_video(
     seconds: float = 5.0,
     output_format: Literal["images", "video"] = "video",
     camera_type: CameraType = CameraType.PERSPECTIVE,
-    colormap: colormaps.Colormaps = "turbo",
-    normalize: bool = True,
-    colormap_min: float = 0.0,
-    colormap_max: float = 1.0,
-    invert: bool = False,
+    colormap_options: colormaps.ColormapOptions = colormaps.ColormapOptions(),
 ) -> None:
     """Helper function to create a video of the spiral trajectory.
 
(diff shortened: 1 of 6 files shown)


SamDM:
Some further experimentation showed that despite a warning from VSCode:
```
PYDEV DEBUGGER WARNING:
sys.settrace() should not be used when the debugger is being used.
This may cause the debugger to stop working correctly.
If this is needed, please check: 
http://pydev.blogspot.com/2007/06/why-cant-pydev-debugger-work-with.html
to see how to restore the debug tracing back correctly.
Call Location:
  File "/home/sam/Safe/OCL3D/tools/nerfstudio/nerfstudio/viewer/server/viewer_utils.py", line 68, in __exit__
    sys.settrace(None)
```

The debugger keeps working fine, so closing this thread.

PR #1921, opened 2023-05-14
Add "on_train_end" method logic similar to Pytorch Lightning

maturk:
- Added on_train_end() optional custom user method to BaseModel.
- Added on train end logic to trainer. 

Let me know if there are perhaps further use cases similar to this method that could be added to either the model or trainer logic and whether or not my implementation is the best option. Happy to learn and get some feedback.

### DIFF
--- a/nerfstudio/engine/trainer.py
+++ b/nerfstudio/engine/trainer.py
@@ -284,6 +284,9 @@ class Trainer:
         # save checkpoint at the end of training
         self.save_checkpoint(step)
 
+        # optional model method called at the end of training
+        self.on_train_end()
+
         # write out any remaining events (e.g., total train time)
         writer.write_out_storage()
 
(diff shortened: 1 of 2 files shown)


nerfstudio/engine/trainer.py
```diff
@@ -503,3 +506,11 @@ def eval_iteration(self, step: int) -> None:
         if step_check(step, self.config.steps_per_eval_all_images):
             metrics_dict = self.pipeline.get_average_eval_image_metrics(step=step)
             writer.put_dict(name="Eval Images Metrics Dict (all images)", scalar_dict=metrics_dict, step=step)
+
+    def on_train_end(self) -> None:
+        """Calls optional on_train_end() model method at training end"""
+        if callable(self.pipeline.model.on_train_end()):
+            try:
```
jkulhanek, line 513: This is bad, it hides the exception. Why did you add the try-catch in the first place?

maturk: I put the try-catch in case a user implements something that fails within the function, so it wouldn't break the trainer.

maturk:
Thanks for the comments @jkulhanek, it helps. The use case for the "at train end" logic is to run custom evaluations and functions on the outputs of a trained nerf, which would include postprocessing images (I am working on color correction for example) and rendering images from specific views. This is mainly something I would like implemented/automated for my own experiments that I am working on, not sure if others would find it useful too. You are right, having this inside the pipeline would expose the most control for the user.

Feel free to close the pull request, I will try to implement a better solution within the pipeline and make a PR for further comments.

PR #1922, opened 2023-05-15
Switching to Torch 2.0 by default.

jkulhanek:
This PR drops support for torch 1.12 and adds support for torch 2.0. Can we open a discussion if this is a good thing to do or not?

Note: we can try torch 2.0 compile to see if there are any speed improvements to be gained.


nerfstudio/engine/schedulers.py
```diff
@@ -25,6 +25,14 @@
 
 from nerfstudio.configs.base_config import InstantiateConfig
 
+try:
+    from torch.optim.lr_scheduler import (  # pylint: disable=ungrouped-imports
+        LRScheduler,
+    )
+except ImportError:
+    # Backward compatibility for PyTorch 1.x
```
jkulhanek, line 33: This offers backward compatibility for PyTorch 1.x. Should we keep it here or drop it?

Dockerfile
```diff
@@ -114,8 +114,8 @@ SHELL ["/bin/bash", "-c"]
 
 # Upgrade pip and install packages.
 RUN python3.10 -m pip install --upgrade pip setuptools pathtools promise pybind11
-# Install pytorch and submodules (Currently, we still use cu116 which is the latest version for torch 1.12.1 and is compatible with CUDA 11.8).
-RUN python3.10 -m pip install torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116
+# Install pytorch and submodules
+RUN python3.10 -m pip install torch==2.0.1+cu117 torchvision==0.15.2+cu117 --extra-index-url https://download.pytorch.org/whl/cu118
```
tancik, line 118: Why is wheel `cu118` but requested version is `cu117`?

jkulhanek: Sorry for this one.

igozali:
Curious if this PR would include support for CUDA 12 as well?

jkulhanek:
Sure. For now only if you compile PyTorch from source: https://discuss.pytorch.org/t/pytorch-for-cuda-12/169447

tancik:
> Sure. For now only if you compile PyTorch from source: https://discuss.pytorch.org/t/pytorch-for-cuda-12/169447

Not sure if Tinycudann and nerfacc would work though.

tancik:
Test renders using nerfacto:
1.13:
https://github.com/nerfstudio-project/nerfstudio/assets/3310961/3f4f53b4-dc83-40a2-a106-d017d43570dd
2.0:
https://github.com/nerfstudio-project/nerfstudio/assets/3310961/ee158711-0a98-48c0-b605-d65a4a3b1633

No noticeable quality differences.

tancik:
The sphinx error is `/home/docs/checkouts/readthedocs.org/user_builds/plenoptix-nerfstudio/checkouts/1922/docs/quickstart/installation.md:199: WARNING: 'myst' reference target not found: tiny-cuda-syntax-error`

tancik:
There appears to be an issue with `instant-ngp`, it is running much slower, but the quality is equivalent. Maybe @liruilong940607 has an idea?

x-axis seconds:
image

x-axis step:
image

nerfstudio/engine/schedulers.py
tancik: Lets keep it for now, but print a warning to the user that they should update to pytorch 2.0

liruilong940607:
Interesting, is this with torch2.0 + `torch.compile()` or simply switching to torch2.0?

tancik:
> Interesting, is this with torch2.0 + `torch.compile()` or simply switching to torch2.0?

Just switching to torch 2.0

jkulhanek:
I am installing a fresh environment to test it.

jkulhanek:
torch.compile breaks in @time_function wrapper.

jkulhanek:
I am getting similar speeds, but I am running it on V100 (cuda compute 70)
image

jkulhanek:
Here is the project:
https://wandb.ai/kulhanek/nerfstudio-torch-comparison?workspace=user-kulhanek

jkulhanek:
@tancik what GPUs are you using? Did torch2.0 printed any warnings?

liruilong940607:
I'm also testing it on TITAN RTX

jkulhanek:
Can anyone confirm if the following commit is safe?
https://github.com/nerfstudio-project/nerfstudio/pull/1922/commits/477a6a0abbecf584a0948513e7e3ed3f12a267e4

It is a fix for one (of many) torch.compile errors.

liruilong940607:
> I'm also testing it on TITAN RTX

I somehow get a different conclusion -- the 2.0 version runs faster than 1.13 on TITAN RTX.

 - step
Screen Shot 2023-05-16 at 1 53 51 PM
- relative wall time
Screen Shot 2023-05-16 at 1 53 25 PM

It's a fresh env with the comment:
```
CUDA_VISIBLE_DEVICES=1 python nerfstudio/scripts/train.py instant-ngp --data data/nerfstudio/poster/ --vis wandb --max-num-iterations 5000
```

tancik:
Im using a 4090 in a fresh env with cuda 1.8

jkulhanek:
> Im using a 4090 in a fresh env with cuda 1.8

Ok, I was using CUDA 11.7 for both torch 1.13 and torch 2.0. I will try CUDA 11.8 with torch 2.0

jkulhanek:
> I am getting similar speeds, but I am running it on V100 (cuda compute 70) image

Sorry, I ran nerfacto here!

tancik:
I've been able to pin the slowness down to the `radial_and_tangential_undistort`
https://github.com/nerfstudio-project/nerfstudio/blob/d542646c72e2f7c8cd03f6677b7489334c1d92c8/nerfstudio/cameras/cameras.py#L631

Perhaps @liruilong940607 tested with a dataset that doesn't have distortion params.

| version | model | speed|
| -- | | -- | -- |
| 1.13 | nerfacto | 🐇  |
| 2.0 | nerfacto | 🐇  |
| 1.13 | instant-ngp | 🐇  |
| 2.0 | instant-ngp | 🐢   |

I wonder if it has to do with the dynamic batch sizes used in `ingp`? Did 2.0 change when it chooses to recompile for JIT? Is it now being triggered each iteration due do the input size changes?

jkulhanek:
Interesting! Can you try replacing torch.jit.script with torch.compile?

liruilong940607:
The poster scene I was testing on also has the distortion parameters. Which scene were you running on? I can also test that out.

liruilong940607:
It is possible that the JIT of that function makes timing weird. What if you just disable the JIT (also do not do torch compile)? I'm not familiar with either but at least we can know if it has something to do with them.

liruilong940607:
> > The poster scene I was testing on also has the distortion parameters. Which scene were you running on? I can also test that out.
> 
> https://data.nerf.studio/nerfstudio/floating-tree.zip

Ok I can reproduce this slowness with floating tree. The GPU stats are very different between these two runs.

Screen Shot 2023-05-17 at 10 57 31 AM
Screen Shot 2023-05-17 at 10 57 17 AM

tancik:
I did some rough timings

| model | opt method | 1.13 time | 2.0 time |
| -- | -- | -- | -- |
| nerfacto | jit script | 66 | 73 |
| nerfacto | jit script | 66 | 73 |
| nerfacto | jit script | 66 | 73 |

jkulhanek:
I have tested 2K iters on A100, CUDA 11.8, PyTorch 2.0.1:
| torch.jit.script | 300s                        |
|------------------|-----------------------------|
| eager mode       | 100s                        |

jkulhanek:
Actually for me torch.compile is the fastest. But the compilation takes a long time at the beginning:
image

tancik:
So is eager the way to go for now? @jkulhanek were these tests with the `ingp` and the floating tree dataset?

jkulhanek:
@tancik , can you please try @torch.compile(dynamic=True) . This is the fastest for me now. I will still play with the configuration for a bit, but I believe this could be it.

jkulhanek:
> So is eager the way to go for now? @jkulhanek were these tests with the `ingp` and the floating tree dataset?

Yes, ingp with floating tree dataset. I wonder if A100 is so much slower compared to 4090 or what is the problem?

jkulhanek:
Alternatively there is also `mode="reduce-overhead"` but it doesn't help much in my case. I would need to do more precise profiling, and there is the tradeoff if we want to spend more time on compile at the beginning of training or have slower iterations.

I suggest we stick with `torch.compile(dynamic=True)` only for now.

liruilong940607:
I can confirm that `torch.compile(dynamic=True, mode="reduce-overhead")` is the best option:

Screen Shot 2023-05-17 at 12 41 02 PM

Additionally I tested using [`nerfacc.cameras.opencv_lens_undistortion()`](https://github.com/KAIR-BAIR/nerfacc/blob/910488e38f94f7c00a14736f4c42a8009740e4f2/nerfacc/cameras.py#LL13C5-L13C29) to replace the `radial_and_tangential_undistort()`, which basically fuse all computation into a single CUDA kernel. It seems like the `torch.compile(dynamic=True, mode="reduce-overhead")` give similar speed with nerfacc's explicit fuse, only with a little bit compiling overhead in the first place (nerfacc needs compiling too but did not reveal here):

Screen Shot 2023-05-17 at 12 45 12 PM

Test with INGP, TITAN RTX, CUDA 11.7, 5000 iters, floating-tree scene.

jkulhanek:
Cool! @liruilong940607 can you please also try with additional backend=“eager”? That setup is the fastest for me.

liruilong940607:
`torch.compile(dynamic=True, mode="reduce-overhead", backend="eager")` is not that good for me. (The `compile-eager` line in the plot)
Screen Shot 2023-05-17 at 1 02 59 PM

tancik:
Here are some results on `nerfacto`:
image

`@torch.compile(dynamic=True, mode="reduce-overhead", backend="eager")`
`@torch.compile(dynamic=True, mode="reduce-overhead")`
`@torch.compile(dynamic=True, backend="eager")`
`@torch.compile(dynamic=False, mode="reduce-overhead", backend="eager")`
`main branch, pytorch 1.13`

They are all basically the same, with the exception of `@torch.compile(dynamic=True, mode="reduce-overhead")` which is worse.

liruilong940607:
Oh I didn't know torch 1.13 also supports `torch.compile`. Isn't it a feature introduce in torch 2.0?

tancik:
> Oh I didn't know torch 1.13 also supports `torch.compile`. Isn't it a feature introduce in torch 2.0?

It doesn't I'm just plotting the main branch for reference (It uses @torch.jit.script)

tancik:
Here are the same plots for `ingp`

liruilong940607:
I see. Does this mean torch.compile behaves basically the same with jit.script, in the case of static shape input? And they only behave differently for dynamic shape (and that’s when all these argument starts to matter)?

I’m curious what’s the logic behind different choices. Some toy examples might be helpful. (Or is there a tutorial somewhere to explain these things?)

Seems like we are ending up in a situation that different model (or even data?) should have different optimization strategy, which kinda makes sense to me.

jkulhanek:
Here is some nice info:
https://pytorch.org/get-started/pytorch-2.0/#user-experience

also here:
https://pytorch.org/docs/stable/dynamo/faq.html

jkulhanek:
It would be cool if we could then torch.compile the whole model to see if we can speed it up. Currently, there are several places where it breaks for nerfacto and other models. I expected the compilation to take much longer, but if it is too much we can cache the compiled model to drive (currently API quite experimental https://pytorch.org/get-started/pytorch-2.0/#user-experience ) so this will only happen once - same as nerfacc kernels I guess.

tancik (approved):
LGTM

PR #1923, opened 2023-05-15
Split up render commands

tancik:
The CLI for `ns-render` was getting messy. This PR splits up different rendering actions as different commands.


brentyi:
Does this need a viewer version bump?

tancik:
> Does this need a viewer version bump?

Yes, good catch

PR #1924, opened 2023-05-15
Fix static type for dataparser configs

brentyi:
https://github.com/nerfstudio-project/nerfstudio/pull/1909 adds support for custom dataparsers via a type that's dynamically generated from `tyro.extras.subcommand_type_from_defaults()`, but unfortunately this type can't be used as an annotation. (in pyright / pylance it gets resolved to `Unknown`)

Not a great solution, but the best one I can think of, curious about thoughts from @jkulhanek / @tancik.


jkulhanek:
I don't think there is anything wrong with it. I tried to achieve the same using `typing.cast(typing.Type[DataParserConfig], ...)`. The type gets resolved correctly, but pylance complains.

jkulhanek (approved):

tancik (approved):

PR #1932, opened 2023-05-16
Added at_train_end logic to base pipeline

maturk:
Here I have added at_train_end() logic to the base pipeline. To accommodate various different user needs, I added similar methods to the base_datamanager and base_model which are callable with **kwargs. 

Here is a small example of how I am personally using it within my custom pipeline for color correction:

```@profiler.time_function
    def at_train_end(self) -> None:
        self.eval()
        camera_ray_bundle, batch = self.datamanager.at_train_end()
        preds, refs = self.model.at_train_end(camera_ray_bundle=camera_ray_bundle, batch=batch)
        cc_images = self.color_correct(img=preds.cpu(), ref=refs.cpu())
        save_image((cc_images *255).astype(np.uint8), self.save_path, log=True)

### CODE REVIEW

File: nerfstudio/data/datamanagers/base_datamanager.py [hunk 1]
```diff
@@ -286,6 +286,9 @@ def get_param_groups(self) -> Dict[str, List[Parameter]]:  # pylint: disable=no-
         """
         return {}
 
+    def at_train_end(self, **kwargs: Any) -> Optional[Any]:  # pylint: disable=unused-argument disable=no-self-use
+        """Called at end of training for optional datamanager outputs."""
+
```
tancik, line 291: This is never called, I worry that users will override this function and expect it to be called.

nerfstudio/models/base_model.py
```diff
@@ -219,3 +219,6 @@ def update_to_step(self, step: int) -> None:
         Args:
             step: training step of the loaded checkpoint
         """
+
+    def at_train_end(self, **kwargs: Any) -> Optional[Any]:  # pylint: disable=unused-argument disable=no-self-use
+        """Called at end of training for optional model outputs."""
```
tancik, line 224: Also never called, same potential confusion as above.

tancik (commented):
We should either make sure that the datamanager and model `train_end` are called, or remove them and leave it to the user to add such a function and call it in the pipeline.

maturk:
Thanks for the feedback again. I have added a new callback type OnTrainEndCallback. I removed unnecessary on_train_end calls from base model and base pipeline. This callback has access to the pipeline object.

Here is an example of how I am using the new callback type within my custom pipeline similar to my previous implementation:

````
    def get_on_train_end_callbacks(self) -> List[OnTrainEndCallback]:
        on_train_end_callbacks = []
        def color_correct_images():
            self.eval()
            camera_ray_bundle, batch = self.datamanager.at_train_end() # No longer in base datamanager, just a user defined function
            preds, refs = self.model.at_train_end(camera_ray_bundle=camera_ray_bundle, batch=batch) # No longer in base model, just a user defined function
            preds = self.color_correct(img=preds.cpu(), ref=refs.cpu())
            save_image((preds *255).astype(np.uint8), self.cc_save_paths, log=True)
        
        on_train_end_callbacks.append(
            OnTrainEndCallback(func=color_correct_images)
        )
        return on_train_end_callbacks

**Comment by jkulhanek** [comment]
I am very sorry I didn't explain it clearly. I meant the already existing callback infrastructure. ...registering the callback here: https://github.com/nerfstudio-project/nerfstudio/blob/8b9574bb5a33a21bd1efae9bd777328ee566c0a7/nerfstudio/engine/callbacks.py#L42 , etc...

**Comment by jkulhanek** [comment]
I think that would integrate in a more concise way with the rest of the code.

**Comment by maturk** [comment]
Whew, no problem :) Happy to learn, maybe third time is the the charm! So I integrated it into the existing callback locations. Only thing I am not sure about is regarding the "step" var when calling run_callback_at_location(step,location) at the end of training. There is really no way to know if step==max-train-iterations inside the TrainingCallback class since this is not exposed. I have set step=None for now to signify that this is being called at train end.

File: nerfstudio/engine/callbacks.py [hunk 3]
```diff
@@ -44,6 +44,7 @@ class TrainingCallbackLocation(Enum):
 
     BEFORE_TRAIN_ITERATION = auto()
     AFTER_TRAIN_ITERATION = auto()
+    ON_TRAIN_END = auto()
```
jkulhanek, line 47: Can we name it AFTER_TRAIN_END?

nerfstudio/engine/trainer.py
```diff
@@ -297,6 +297,10 @@ def train(self) -> None:
         table.add_row("Checkpoint Directory", str(self.checkpoint_dir))
         CONSOLE.print(Panel(table, title="[bold][green]:tada: Training Finished :tada:[/bold]", expand=False))
 
+        # on train end callbacks
+        for callback in self.callbacks:
+            callback.run_callback_at_location(step=None, location=TrainingCallbackLocation.ON_TRAIN_END)
```
jkulhanek, line 302: Can you pass the actual step here?

nerfstudio/engine/callbacks.py
```diff
@@ -91,8 +92,10 @@ def run_callback(self, step: int) -> None:
         elif self.iters is not None:
             if step in self.iters:
                 self.func(*self.args, **self.kwargs, step=step)
+        elif step is None:
+            self.func(*self.args, **self.kwargs, step=step)
 
-    def run_callback_at_location(self, step: int, location: TrainingCallbackLocation) -> None:
+    def run_callback_at_location(self, step: Union[int, None], location: TrainingCallbackLocation) -> None:
```
jkulhanek, line 98: Please keep the signature as int

nerfstudio/engine/callbacks.py
```diff
@@ -71,15 +72,15 @@ def __init__(
...
+        ), f"'step: Union[int, None]' must be an argument in the callback function 'func': {func.__name__}"
         self.where_to_run = where_to_run
         self.update_every_num_iters = update_every_num_iters
         self.iters = iters
         self.func = func
         self.args = args if args is not None else []
         self.kwargs = kwargs if kwargs is not None else {}
 
-    def run_callback(self, step: int) -> None:
+    def run_callback(self, step: Union[int, None]) -> None:
```
jkulhanek, line 83: Please keep int here.

nerfstudio/engine/callbacks.py
```diff
@@ -71,15 +72,15 @@ def __init__(
     ):
         assert (
             "step" in signature(func).parameters.keys()
-        ), f"'step: int' must be an argument in the callback function 'func': {func.__name__}"
+        ), f"'step: Union[int, None]' must be an argument in the callback function 'func': {func.__name__}"
```
jkulhanek, line 75: Same here, please keep int

nerfstudio/engine/callbacks.py
jkulhanek: Or perhaps "AFTER _TRAIN"?

tancik (approved):
LGTM

PR #1934, opened 2023-05-16
Discover external methods from CLI

jkulhanek:
This PR allows external methods to be discovered and installed from cli (ns-*).
New methods appear under `ns-train --help`. Upon running `ns-train externalmethodname` a message appears and optionally allows the external dependencies to be installed.

### DIFF
--- a/nerfstudio/configs/external_methods.py
+++ b/nerfstudio/configs/external_methods.py
@@ -31,8 +31,13 @@ class ExternalMethod:
     """External method class. Represents a link to a nerfstudio-compatible method not included in this repository."""
 
     instructions: str
+    """Instructions for installing the method. This will be printed to
+    the console when the user tries to use the method."""
     configurations: List[Tuple[str, str]]
+    """List of configurations for the method. Each configuration is a tuple of (registered slug, description)
+    as it will be printed in --help."""
     pip_package: Optional[str] = None
+    """Specifies a pip package if the method can be installed by running `pip install <pip_package>`."""
 
 
 external_methods = []
--- a/nerfstudio/configs/external_methods.py
+++ b/nerfstudio/configs/external_methods.py
@@ -18,7 +18,7 @@
 import subprocess
 import sys
 from dataclasses import dataclass, field
-from typing import Any, Dict, Optional, Tuple, cast
+from typing import Any, Dict, List, Optional, Tuple, cast
 
 from rich.prompt import Confirm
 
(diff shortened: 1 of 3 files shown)


nerfstudio/configs/external_methods.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+@dataclass
+class ExternalMethod:
+    """External method class. Represents a link to a nerfstudio-compatible method not included in this repository."""
+
+    slug: str
+    model_description: str
+    instructions: str
+    install_commands: Optional[str] = None
```
tancik, line 36: Instead of install command, what if we just have the user specify the git repo

nerfstudio/configs/external_methods.py
```diff
@@ -0,0 +1,160 @@
...
+            "Instruct-NeRF2NeRF. Half prevision with no LPIPS",
+            in2n_install_instructions,
+            in2n_install_script,
+        ),
+        ExternalMethod(
+            "in2n-tiny",
+            "Instruct-NeRF2NeRF. Half prevision with no LPIPS",
+            in2n_install_instructions,
+            in2n_install_script,
+        ),
```
tancik, line 70: Duplicate

nerfstudio/configs/external_methods.py
```diff
@@ -0,0 +1,160 @@
...
+    ]
+)
```
tancik, line 72: Not sure if it is any better, but what about something more concise, like

```
# Instruct-NeRF2NeRF
in2n_install_instructions = """
[bold yellow]Instruct-NeRF2NeRF[/bold yellow]
For more information visit: https://docs.nerf.studio/en/latest/nerfology/methods/in2n.html
To enable Instruct-NeRF2NeRF, you must install it first by running:
  [grey]pip install git+https://github.com/ayaanzhaque/instruct-nerf2nerf[/grey]
""".strip()
external_methods.append(
       ExternalMethod(
            methods={
                   "in2n": "Instruct-NeRF2NeRF. Full model, used in paper",
                   "in2n-small": "Instruct-NeRF2NeRF. Half precision model", 
                   "in2n-tiny": "Instruct-NeRF2NeRF. Half prevision with no LPIPS", 
             }
             instructions: in2n_install_instructions,
             install_repo: "https://github.com/ayaanzhaque/instruct-nerf2nerf"
)
```

tancik (commented):
This is quite handy, thanks for adding.
We should also include documentation for the feature and make it a requirement when adding a new method.

+ I just tested it and ran into a couple of issues, `ns-train --help` does not register `--help` and provides the same response as just running `ns-train`
+ I ran `lerf` then confirmed the github install, but it still get the install instructions after it's been installed.

jkulhanek:
@tancik Thank you very much for spotting the issue!

nerfstudio/configs/external_methods.py
jkulhanek: Is it better now?

nerfstudio/configs/external_methods.py
jkulhanek: Perhaps not all methods in future will be installed from git? Some may be on pip or installed from a git subdirectory? What do you think of the current alternative?

tancik: What about `pip_package`?

nerfstudio/configs/external_methods.py
```diff
@@ -0,0 +1,131 @@
...
+from nerfstudio.utils.rich_utils import CONSOLE
+
+
+@dataclass
+class ExternalMethod:
+    """External method class. Represents a link to a nerfstudio-compatible method not included in this repository."""
+
+    instructions: str
+    configurations: List[Tuple[str, str]]
+    pip_package: Optional[str] = None
```
tancik, line 35: Add argstrings

nerfstudio/configs/external_methods.py
tancik: Yea, I like it more.

tancik:
I am still getting the issue where it doesn't think the package is installed, 
image

Also can you add documentation tot the webdocs about this feature.

jkulhanek:
I am not able to reproduce it. For me, it works fine... after installing in2n, it does not show the message anymore and runs in2n. LERF does not work for me. After I install LERF it breaks nerfstudio:
```
AssertionError: `pipeline.datamanager.dataparser` was provided a default value of type <class 'nerfstudio.data.dataparsers.nerfstudio_dataparser.NerfstudioDataParserConfig'> but no matching subcommand was found.
```

What is the output of 'ns-train --help` after you install LERF?

tancik:
> I am not able to reproduce it. For me, it works fine... after installing in2n, it does not show the message anymore and runs in2n. LERF does not work for me. After I install LERF it breaks nerfstudio:
> 
> ```
> AssertionError: `pipeline.datamanager.dataparser` was provided a default value of type <class 'nerfstudio.data.dataparsers.nerfstudio_dataparser.NerfstudioDataParserConfig'> but no matching subcommand was found.
> ```
> 
> What is the output of 'ns-train --help` after you install LERF?

image
 I also tried `in2n` which has the same issue.

jkulhanek:
That is strange. After installing `in2n` I am getting the following:
image

Installed methods shouldn't have descriptions starting with [External] and uninstalled methods should be at the end of the list, not in the middle. What is your commit SHA?

tancik:
Silly me, I was on an old commit

tancik (approved):
LGTM

PR #1939, opened 2023-05-17
Switch to `ruff`

brentyi:
Let's move to `ruff`!

Runtimes on my machine:
- `pylint nerfstudio tests`: 31.64 seconds
- `isort --profile=black nerfstudio docs tests`: 0.28 seconds
- `ruff check nerfstudio docs tests`: 0.01 seconds

Next steps:
- `pylint` rules are still a work-in-progress in `ruff` (only about 1/3 of the rules are implemented), but I'd personally vote to just move over because: it seems like most of the important ones are there, it's nice to be forward-looking, type checking in the CI whenever it lands will handle most robustness issues, and `pylint` is super slow. If you disagree @tancik I can just revert the `pylint` parts of this PR, and keep only the `isort` bits.
- There are some VS Code settings that have `pylint` enabled in them. This could be moved to the `ruff` extension? Probably best handled by someone who uses VS Code.
- Ruff has a lot of rules that I haven't turned on, because they result in a slew of errors. The `pydocstyle` and `numpy` rules in particular could be worth exploring in the future. https://beta.ruff.rs/docs/rules/

### DIFF
--- a/.github/workflows/core_code_checks.yml
+++ b/.github/workflows/core_code_checks.yml
@@ -32,13 +32,10 @@ jobs:
       - name: Check notebook cell metadata
         run: |
           python ./nerfstudio/scripts/docs/add_nb_tags.py --check
-      - name: Run ruff
+      - name: Run Ruff
         run: ruff docs/ nerfstudio/ tests/
       - name: Run Black
         run: black docs/ nerfstudio/ tests/ --check
-      - name: Python Pylint
-        run: |
-          pylint nerfstudio tests
       - name: Test with pytest
         run: |
           pytest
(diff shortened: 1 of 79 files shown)


SauravMaheshkar:
> If linting is fast enough, can we add pre-commit (https://pre-commit.com/) hook? It is used in PyTorch repos and I think it can work well here.

Yes, I'd love for there to be pre-commit hooks. In most PRs there are some commits purely for making linting fixes. Having pre-commit hooks will surely bring those down and make the git history cleaner.

tancik (approved):
LGTM, soo much faster!

PR #1940, opened 2023-05-17
Fix some typing problems

jkulhanek:
This PR fixes some typing issues in models, model_components, fields. Now pyright, vscode (pylance) do not report any errors for these subpackages.

### DIFF
--- a/nerfstudio/models/nerfplayer_nerfacto.py
+++ b/nerfstudio/models/nerfplayer_nerfacto.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 
 import functools
 from dataclasses import dataclass, field
-from typing import Dict, List, Literal, Type
+from typing import Dict, List, Literal, Type, cast
 
 import numpy as np
 import torch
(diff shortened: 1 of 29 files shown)


jkulhanek:
@brentyi can you take a look?

nerfstudio/models/nerfplayer_nerfacto.py
```diff
@@ -119,7 +119,8 @@ def populate_modules(self):
         self.density_fns = []
         num_prop_nets = self.config.num_proposal_iterations
         # Build the proposal network(s)
-        self.proposal_networks = torch.nn.ModuleList()
+        # The following line uses a typing workaround (may not be the best way to do this)
+        self.proposal_networks = cast(List[TemporalHashMLPDensityField], torch.nn.ModuleList())
```
jkulhanek, line 123: This line is not good. Any suggestions?

SauravMaheshkar: The best I can think of is a `PROPOSAL_NETWORKS_TYPE = List[TemporalHashMLPDensityField]` line at the top and then just do a `self.proposal_networks = cast(PROPOSAL_NETWORKS_TYPE, torch.nn.ModuleList())`

brentyi: I'm okay with what you originally had, but I pushed an alternative solution where we instead cast the `ModuleList` instance to `Sequence[TemporalHashMLPDensityField]`. This seems a little bit more "correct" since `Sequence` is a protocol, although in practice either seem OK?

tancik (approved):
LGTM

brentyi (approved):

brentyi:
FYI I'm going to try and get the rest of the type errors resolve today, and (hopefully) get `pyright` into the the CI

PR #1942, opened 2023-05-17
bump pyproject torch

tancik:


jkulhanek:
1. If torch version is bumped, torchvision needs to be bumped as well to 1.15.2
2. can I now remove the backward-compatible code if 1.13 is no longer supported?

PR #1948, opened 2023-05-18
Drop backward compatibility for Torch 1.13

jkulhanek:


tancik (approved):
LGTM, decided kinda last minute to only support 2.0, thanks for making these residual changes.

PR #1951, opened 2023-05-18
[WIP] Reducing the time for RayGenerator 

yuerout:
Profiling with PyTorch profiler give:
image 
A lot of time is spent inside https://github.com/nerfstudio-project/nerfstudio/blob/9584fa9b39bd276c97830a4753791a94afd8e9f3/nerfstudio/model_components/ray_generators.py#L56
2 ideas for reducing the time needed to run this:
- precompute ray bundle and load from disk 
- multithreading with torch dataloader

### DIFF
--- a/nerfstudio/data/datamanagers/base_datamanager.py
+++ b/nerfstudio/data/datamanagers/base_datamanager.py
@@ -497,6 +497,7 @@ class VanillaDataManager(DataManager, Generic[TDataset]):
         assert isinstance(image_batch, dict)
         batch = self.train_pixel_sampler.sample(image_batch)
         ray_indices = batch["indices"]
+        # this takes some time to run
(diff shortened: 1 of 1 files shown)


jkulhanek:
Was the trace generated with `CUDA_LAUNCH_BLOCKING=1`?

jkulhanek:
Also, which step is recorded here?

yuerout:
> Was the trace generated with `CUDA_LAUNCH_BLOCKING=1`?

no, I think CUDA_LAUNCH_BLOCKING=0

> Also, which step is recorded here?

ns-train

jkulhanek:
Can you try with blocking=1?

i meant what step number is recorded. I would advise to record some later iteration.

tancik:
Closing due to inactivity.

PR #1959, opened 2023-05-19
generalize mask collation to other image-like data

decrispell:
Variable resolution image collation currently looks for the special keys "image" and "mask" only.
This change is a generalization that looks for "image", and then treats any other keys with the same first two dimensions as "image" as auxiliary data to be collated accordingly.


tancik (approved):
LGTM for now, I think long-term we should think about how to give the user more control over what is and isn't loaded / the ability to load data at different scales

PR #1965, opened 2023-05-21
Add option for model normals on ply export

CameronFraser:
This PR makes the arguments related to normals for point cloud export consistent with the arguments related to normals for Poisson export. Doing this adds support for the use case where a user may want to export just a point cloud with predicted normals instead of o3d estimated normals to use in their own Poisson meshing pipeline.

Additionally, the viewer is updated to reflect the changes.

### DIFF
--- a/nerfstudio/scripts/exporter.py
+++ b/nerfstudio/scripts/exporter.py
@@ -59,6 +59,30 @@ class Exporter:
     """Path to the output directory."""
 
 
+def validate_pipeline(normal_method, normal_output_name, pipeline: Pipeline) -> None:
+    """Check that the pipeline is valid for this exporter."""
+    if normal_method == "model_output":
+        CONSOLE.print("Checking that the pipeline has a normal output.")
+        origins = torch.zeros((1, 3), device=pipeline.device)
+        directions = torch.ones_like(origins)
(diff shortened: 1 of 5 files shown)


nerfstudio/scripts/exporter.py
```diff
@@ -59,6 +59,30 @@ class Exporter:
     """Path to the output directory."""
 
 
+def validate_pipeline(normal_method, normal_output_name, pipeline: Pipeline) -> None:
```
tancik, line 62: Add types and documentation for the args.

tancik (commented):
Thanks for adding. Sorry for the delayed review.

nerfstudio/scripts/exporter.py
CameronFraser: np, I just updated that file

ethanweber:
LGTM. Can you merge `main` now and then we can merge into `main`?

CameronFraser:
@ethanweber updated branch with main

ethanweber (approved):
Great thanks! LGTM, and I tried out the viewer changes too.

CameronFraser:
worth noting that in my case when using nerfstudio with aerial imagery the nerfacto predicted normals and the resulting mesh were significantly better looking than using open3d

PR #1970, opened 2023-05-22
Update installation docs

tancik:


docs/quickstart/installation.md
```diff
@@ -38,14 +32,15 @@ python -m pip install --upgrade pip
...
 ### pytorch
 
 ::::{tab-set}
 :::{tab-item} Torch 2.0.1 with CUDA 11.8
 
 - To install 2.0.1 with CUDA 11.8:
 
-Note that if a pytorch version prior to 2.0.1 is installed, 
+Note that if a pytorch version prior to 2.0.1 is installed,
 it should be uninstalled first to avoid upgrade issues (e.g. with functorch)
```
machenmusik, line 44: My understanding is that functorch may need to be explicitly uninstalled
`pip uninstall functorch`
because just uninstalling torch doesn't seem to uninstall functorch.
However, that may also impact other projects.

docs/quickstart/installation.md
```diff
@@ -63,7 +58,7 @@ pip install torch==2.0.1+cu118 torchvision==0.15.2+cu118 --extra-index-url https
 
 - To install 2.0.1 with CUDA 11.7:
 
-Note that if a pytorch version prior to 2.0.1 is installed, 
+Note that if a pytorch version prior to 2.0.1 is installed,
```
machenmusik, line 61: My understanding is that functorch may need to be explicitly uninstalled
`pip uninstall functorch`
because just uninstalling torch doesn't seem to uninstall functorch.
However, that may also impact other projects.

docs/quickstart/installation.md
```diff
@@ -157,50 +161,76 @@ docker run --gpus all \                                         # Give the conta
...
 - [Installation errors, File "setup.py" not found](pip-install-error)
 - [Runtime errors, "len(sources) > 0".](cuda-sources-error)
 
  <br />
 
+(tiny-cuda-mismatch-arch)=
+
+**ImportError: DLL load failed while importing \_89_C**
+
+tiny-cuda-nn was installed for the wrong hardware architecture.
```
machenmusik, line 220: I was going to do a PR with this description of the problem
`This occurs with certain GPUs that have CUDA architecture versions (89 in the example above) for which TinyCUDA does not automatically compile support.`

docs/quickstart/installation.md
```diff
@@ -157,50 +161,76 @@ docker run --gpus all \                                         # Give the conta
...
+
+**Solution**:
+
+Reinstall tiny-cuda with the following command:
+
+```bash
+TCNN_CUDA_ARCHITECTURES=XX pip install git+https://github.com/NVlabs/tiny-cuda-nn/#subdirectory=bindings/torch
```
machenmusik, line 227: This doesn't work in Windows Command Prompt, for which the command is `set TCNN_CUDA_ARCHITECTURES=XX`

machenmusik (changes requested):
A few suggestions, but the main difference is how to set the environment variable if using Windows Command Prompt.  It may be better to have a tabset for Linux and Windows

docs/quickstart/installation.md
tancik: I'm confused what your recommendation is here, the following line uninstalls functorch

tancik: Updated to make more clear

machenmusik (approved):
LGTM, only thing I would say is that we are inconsistent about tiny-cuda-nn, TinyCUDA, etc. in the docs

PR #1971, opened 2023-05-22
Fix mask check in PatchPixelSampler

lukasHoel:
Small fix in `PatchPixelSampler` in case a mask is present.
Tested this while experimenting with InstructNerf2Nerf and a custom dataset that has masks.


nerfstudio/data/pixel_samplers.py
```diff
@@ -278,7 +278,7 @@ def sample_method(
         mask: Optional[Tensor] = None,
         device: Union[torch.device, str] = "cpu",
     ) -> Int[Tensor, "batch_size 3"]:
-        if mask:
+        if isinstance(mask, torch.Tensor):
```
tancik, line 281: nit: we import Tensor, so 
`if isinstance(mask, Tensor):`
or
`if mask is not None:`
would be preferred

tancik (approved):
LGTM

PR #1982, opened 2023-05-23
Add warnings for (slower) Torch 1.x and Windows not supporting torch.compile

jkulhanek:


tancik (approved):
LGTM, thanks for adding.

PR #1944, opened 2023-05-18
Final typing fixes + add Pyright to CI

tancik:

PR #1990, opened 2023-05-24
[WIP] Fix model loading collate_fn non-static

jkulhanek:
After fixing typing errors #1944  I believe there as a bug introduced which prevents trained checkpoints to be loaded.

Similar issue here:
https://github.com/pydantic/pydantic/issues/1596

Can anyone please confirm that loading of trained checkpoints is broken and this PR fixes it?


tancik:
Hmm, Im able to load checkpoints.

Train for 2k iters
`ns-train nerfacto --data data/nerfstudio/test/`

Load checkpoint
`ns-train nerfacto --data data/nerfstudio/test/ --load-dir outputs/test/nerfacto/2023-05-24_104341/nerfstudio_models/step-000002000.ckpt`

jkulhanek:
I believe the error happens when you load pre 0.3.0 checkpoint in >=0.3.0 nerfstudio.
I will verify it.

jkulhanek:
Yes, you can reproduce the error if you remove the collate_fn field manually from config.json:
image

jkulhanek:
What I suggested here is also not good. I don't think collate_fn gets registered as a field now.

jkulhanek:
I guess the problem is in the yaml loader. Do you think it is worth fixing?

tancik (approved):
LGTM

PR #1993, opened 2023-05-24
add deepfloyd model

terrancewang:


nerfstudio/generative/deepfloyd.py
```diff
@@ -0,0 +1,207 @@
...
+        # Return magniture of gradient, not the actual loss.
+        return torch.mean(gt_grad**2) ** 0.5
+
+    @staticmethod
+    @custom_bwd
+    def backward(ctx, grad):  # pylint: disable=arguments-differ
+        del grad
+        (gt_grad,) = ctx.saved_tensors
+        batch_size = len(gt_grad)
+        return gt_grad / batch_size, None
```
tancik, line 41: Create a new utils folder in `generative/` that has this so that It can be shared between SD and deepfloyd

nerfstudio/generative/deepfloyd.py
```diff
@@ -0,0 +1,207 @@
...
+
+class DeepFloyd(nn.Module):
```
tancik, line 43: Add documentation for class and methods that are currently missing docs.

nerfstudio/generative/deepfloyd.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def sds_loss(
+        self,
+        text_embeddings, 
+        image, 
+        guidance_scale, 
+        grad_scaler,
+    ) -> torch.Tensor:
+        with torch.autocast(device_type="cuda", enabled=False):
+            image = F.interpolate(image.half(), (IMG_DIM, IMG_DIM), mode="bilinear", align_corners=False)
```
tancik, line 130: Does `IMG_DIM` always need to be 64? If not, we might want to add it as an argument.

nerfstudio/generative/deepfloyd.py
tancik, line 43: Add types to all methods

tancik (commented):
Thanks for adding!

nerfstudio/generative/deepfloyd.py
terrancewang: We're only using the first model of Deepfloyd for SDS, which is trained to generate 64x64 images. Not sure what the results would look like if we used a different image size though. 

the other two models in the pipeline are the super resolution models that allow it to generate higher resolution images

terrancewang:
unsure how to resolve the last 3 pyright errors

nerfstudio/generative/deepfloyd.py
tancik: ok, feel free to ignore my comment

nerfstudio/generative/gradient_utils.py
tancik, line 1: I would make the filename just `utils.py`. I can imagine more generative util functions in the future, but probably few if any `gradient` specific utils.

terrancewang:
i tried resolving the last 3 pyright errors but im unsure how to fix the typing for them

nerfstudio/generative/deepfloyd.py
```diff
@@ -0,0 +1,233 @@
...
+        assert isinstance(negative_embeds, FloatTensor)
+        assert isinstance(prompt_embeds, FloatTensor)
+        return torch.cat([negative_embeds, prompt_embeds])
+
+    def sds_loss(
+        self,
+        text_embeddings,
+        image,
+        guidance_scale,
+        grad_scaler,
```
tancik, line 136: Missing types

tancik (approved):
Lgtm

PR #1994, opened 2023-05-24
style(ruff): provide better integration with ruff vscode extension

SauravMaheshkar:
Makes a few changes to better integrate with the ruff vscode extension.


tancik (approved):
LGTM

PR #1996, opened 2023-05-24
pass grad sclaer from trainer to pipeline

terrancewang:


nerfstudio/pipelines/base_pipeline.py
```diff
@@ -235,6 +236,7 @@ def __init__(
         test_mode: Literal["test", "val", "inference"] = "val",
         world_size: int = 1,
         local_rank: int = 0,
+        grad_scaler: Optional[GradScaler] = None,
```
tancik, line 239: Add `grad_scalar` to arg doc

PR #2001, opened 2023-05-25
Add Metadata Handling to `Cameras` Collation Function

MingwuZheng:
This pull request addresses an issue in the `nerfstudio_collate` function where the metadata of `Cameras` objects is not properly accounted for during batch processing.

In the existing implementation, the metadata of `Cameras` objects is not included during collation, leading to the assignment of `None` to the metadata field. This may cause downstream issues for components that rely on this metadata for processing.

The proposed changes aim to ensure that the metadata of each `Cameras` object in a batch is properly collated. The updates involve the creation of a metadata dictionary that collates the metadata from each `Cameras` object in the batch. This dictionary is then passed to the `Cameras` constructor during batch assembly.

The changes ensure that all `Cameras` objects have the same metadata keys before proceeding with the collation. This ensures consistency across the batch and prevents potential errors from occurring due to mismatched or missing metadata keys.


tancik:
Can you provide a PR description. Also look [here](https://docs.nerf.studio/en/latest/reference/contributing.html#requirements) for installing `pre-commit` to avoid these failing checks.

MingwuZheng:
Hey @tancik,
Just wanted to let you know that I've passed the tests and updated the PR. Could you take another look when you have a moment?

tancik (approved):
LGTM

PR #2011, opened 2023-05-28
Update metashape_utils.py

henrypearce4D:
Changed the ValueError message when using a camera.xml that contains multiple sensors. Per-frame intrinsics can be defined, the previous message said that they were not supported.

### DIFF
--- a/nerfstudio/process_data/metashape_utils.py
+++ b/nerfstudio/process_data/metashape_utils.py
@@ -68,7 +68,7 @@ def metashape_to_json(
     if sensor_type.count(sensor_type[0]) != len(sensor_type):
         raise ValueError(
             "All Metashape sensors do not have the same sensor type. "
-            "nerfstudio does not support per-frame camera_model."
+            "nerfstudio does support per-frame camera intrinsics, but at this current time the utility does not account for this."
         )
     data = {}
     if sensor_type[0] == "frame":


brentyi (approved):
Looks correct to me!

f-dy:
I don't understand: the previous message was correct, and it was triggered only when using several camera_model values in the metashape project (eg fisheye + perspective).

The limitation does not come from this utility but from nerfstudio itself, see [here](https://github.com/nerfstudio-project/nerfstudio/blob/43c399e71ead6a9a733e29efa1c5966da3ff4629/docs/quickstart/data_conventions.md?plain=1#LL44C168-L44C210):  "Per-frame `camera_model` is not supported." - which is exactly the sentence I used in the error message.

f-dy:
@henrypearce4D @brentyi I think this should be reverted, as the new message is more confusing.

brentyi:
Yeah, on a re-read the original error message makes sense. @henrypearce4D can you comment/confirm?

henrypearce4D:
@brentyi -
@f-dy Is correct, it can be reverted perhaps with a little clarification added as another user recently mentioned this was a little confusing;

`nerfstudio does not support per-frame camera_model types, only one sensor type can be used. OPENCV, OPENCV_FISHEYE, EQUIRECTANGULAR.`

brentyi:
@henrypearce4D that sounds reasonable to me!

Issue #1498, opened 2023-02-24
Training with masks results in artefacts outside the mask regions


nepfaff:
Hi, I want to train a NeRF of a masked object for mesh + texture map extraction. However, I'm getting unexpected results where the renders include artefacts outside the masked regions.

I'm using the same dataset (download [here](https://drive.google.com/file/d/16F6hcki0UorqbGB9OCOn7kFDPHtJ3Bw6/view?usp=sharing)) for all the following results. I additionally copied the mask parsing logic from the `nerfstudio_dataparser` to the `instant_ngp_dataparser`.

I built nerfstudio from source, commit 4d333f4cb6fba7c1cb0b478f40a20c6fb06edef1.

`instant-ngp-bounded` is giving me the most similar results to the official instantNGP implementation. However, I'm still getting artefacts at the top of the scene bounding cube.

Are there any settings that I could change to obtain my desired results?

### Desired results (Using the official instantNGP implementation with masked images and `black_transparent=True`)

![Screenshot from 2023-02-24 10-52-13](https://user-images.githubusercontent.com/53228351/221258835-fccf90ae-8425-48f2-8ca0-63e5f4a2a763.png)

### Nerfstudio Nerfacto results (default settings)

```ns-train nerfacto instant-ngp-data --data data/mustard```

![Screenshot from 2023-02-24 10-49-02](https://user-images.githubusercontent.com/53228351/221258911-dbf4767b-6d9b-447b-974a-7b00fdb21557.png)

### Nerfstudio instant-ngp-bounded results (default settings)

```ns-train instant-ngp-bounded instant-ngp-data --data data/mustard```

![Screenshot from 2023-02-24 13-06-05](https://user-images.githubusercontent.com/53228351/221259076-ca53faa0-7f1f-4cf5-8d0e-981a473a58d5.png)

nepfaff:
This does not seem to solve my problem as it would only remove the artefacts completely if I could get the ROI bounding box to be exactly my object of interest.

I looked more into the difference to the official instantNGP implementation and realized that its `black_transparent` works very differently. It aims to achieve zero density at coordinates corresponding to black image pixels. This is more powerful than only using masked pixels for rendering. instantNGP also provides the weaker mask option as nerfstudio does, as this is the desired approach if images contain, for example, dynamic objects that should be ignored.

In this case, can I convert this into a feature request to also support transparency enforcement?

jkulhanek:
[...]

PR #2025, opened 2023-06-01
add transparency carving to nerfacto

SamDM:
there is a very simple trick not implemented in Nerfacto that significantly reduces floaters in synthetic scenes with few cameras. I implemented it, below is a small comparison with and without this addition.

Without:
plant-no-alpha-carve

With:
plant-with-alpha-carve

The trick only works if you have foreground masks, it works as follows:

1. Make everything except for the foreground object transparent in each image.
2. While rendering, pick a random color in each iteration, then use this random color as background during alpha compositing.
3. When computing the RGB loss, make the color of transparent pixels match the randomly chosen color of the current iteration used for rendering.

The end result is that floaters will get punished because then can't keep up with the constantly changing background color, therefore pushing them towards transparency.

### DIFF
--- a/nerfstudio/data/dataparsers/base_dataparser.py
+++ b/nerfstudio/data/dataparsers/base_dataparser.py
@@ -57,6 +57,8 @@ class DataparserOutputs:
     """Camera object storing collection of camera information in dataset."""
     alpha_color: Optional[Float[Tensor, "3"]] = None
     """Color of dataset background."""
+    use_alpha_channel: bool = False
+    """Keep the alpha channel if present (or add one) when loading image data"""
     scene_box: SceneBox = SceneBox(aabb=torch.tensor([[-1, -1, -1], [1, 1, 1]]))
     """Scene box of dataset. Used to bound the scene or provide the scene scale depending on model."""
     mask_filenames: Optional[List[Path]] = None
--- a/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
+++ b/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
@@ -26,11 +26,7 @@ from PIL import Image
 
 from nerfstudio.cameras import camera_utils
 from nerfstudio.cameras.cameras import CAMERA_MODEL_TO_TYPE, Cameras, CameraType
-from nerfstudio.data.dataparsers.base_dataparser import (
-    DataParser,
-    DataParserConfig,
-    DataparserOutputs,
-)
+from nerfstudio.data.dataparsers.base_dataparser import DataParser, DataParserConfig, DataparserOutputs
 from nerfstudio.data.scene_box import SceneBox
 from nerfstudio.utils.io import load_from_json
 from nerfstudio.utils.rich_utils import CONSOLE
@@ -296,9 +292,15 @@ class Nerfstudio(DataParser):
             applied_scale = float(meta["applied_scale"])
             scale_factor *= applied_scale
 
+        if "use_alpha_channel" in meta:
+            use_alpha_channel = bool(meta["use_alpha_channel"])
+        else:
+            use_alpha_channel = False
+
         dataparser_outputs = DataparserOutputs(
             image_filenames=image_filenames,
             cameras=cameras,
+            use_alpha_channel=use_alpha_channel,
             scene_box=scene_box,
             mask_filenames=mask_filenames if len(mask_filenames) > 0 else None,
             dataparser_scale=scale_factor,
@@ -306,6 +308,7 @@ class Nerfstudio(DataParser):
             metadata={
                 "depth_filenames": depth_filenames if len(depth_filenames) > 0 else None,
                 "depth_unit_scale_factor": self.config.depth_unit_scale_factor,
+                "use_alpha_channel": use_alpha_channel,
             },
         )
         return dataparser_outputs
--- a/nerfstudio/models/nerfacto.py
+++ b/nerfstudio/models/nerfacto.py
@@ -29,11 +29,7 @@ from torchmetrics.functional import structural_similarity_index_measure
 from torchmetrics.image.lpip import LearnedPerceptualImagePatchSimilarity
 
 from nerfstudio.cameras.rays import RayBundle, RaySamples
-from nerfstudio.engine.callbacks import (
-    TrainingCallback,
-    TrainingCallbackAttributes,
-    TrainingCallbackLocation,
-)
+from nerfstudio.engine.callbacks import TrainingCallback, TrainingCallbackAttributes, TrainingCallbackLocation
 from nerfstudio.field_components.field_heads import FieldHeadNames
 from nerfstudio.field_components.spatial_distortions import SceneContraction
 from nerfstudio.fields.density_fields import HashMLPDensityField
@@ -46,16 +42,8 @@ from nerfstudio.model_components.losses import (
     pred_normal_loss,
     scale_gradients_by_distance_squared,
 )
-from nerfstudio.model_components.ray_samplers import (
-    ProposalNetworkSampler,
-    UniformSampler,
-)
-from nerfstudio.model_components.renderers import (
-    AccumulationRenderer,
-    DepthRenderer,
-    NormalsRenderer,
-    RGBRenderer,
-)
+from nerfstudio.model_components.ray_samplers import ProposalNetworkSampler, UniformSampler
+from nerfstudio.model_components.renderers import AccumulationRenderer, DepthRenderer, NormalsRenderer, RGBRenderer
 from nerfstudio.model_components.scene_colliders import NearFarCollider
 from nerfstudio.model_components.shaders import NormalsShader
 from nerfstudio.models.base_model import Model, ModelConfig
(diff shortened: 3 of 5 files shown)


machenmusik:
Interesting.   How does this do with mustard dataset from https://github.com/nerfstudio-project/nerfstudio/issues/1498#issue-1599115766 I wonder...

f-dy:
I think we should have the two kinds of masks:
1. ignore masks to ignore parts of the image (eg transients): do not sample rays outside of these masks
2. transparency masks to say everything outside of these masks should be transparent: adds the transparency loss

nepfaff:
> I think we should have the two kinds of masks:
> 
> 1. ignore masks to ignore parts of the image (eg transients): do not sample rays outside of these masks
> 2. transparency masks to say everything outside of these masks should be transparent: adds the transparency loss

The first one is already implemented in Nerfstudio. The second one is what this PR is trying to do and what instantNGP is doing. I actually planned to implement something like this once I'm back from my vacation, so I'm excited that this PR already got opened. Thanks! I will try this on the masked mustard dataset soon and compare it with instantNGP.

nepfaff:
@SamDM would you mind providing the data and training command that you used for your testing? It would be interesting to compare your results with my nerfacto results with default settings above.

nepfaff:
I added the same logic as done in this PR to `NGPModel` and run it with the default settings of `instant-ngp-bounded`. The results now closely correspond to the ones when using `instantNGP` with alpha transparent training:

https://github.com/nerfstudio-project/nerfstudio/assets/53228351/b4fa76e2-216e-4df0-9bf7-429168ba58bb

**These results are great, and I would support merging this.**
 I'm still unsure why `nerfacto` did not work for this dataset when using alpha transparent training. If someone has any insights here, it would be great to mention something in the docs about when alpha transparent training is expected to work (model, parameters, etc.).

nerfstudio/models/nerfacto.py
```diff
@@ -327,15 +322,20 @@ def get_outputs(self, ray_bundle: RayBundle):
...
     def get_loss_dict(self, outputs, batch, metrics_dict=None):
         loss_dict = {}
         image = batch["image"].to(self.device)
-        loss_dict["rgb_loss"] = self.rgb_loss(image, outputs["rgb"])
+        if outputs.get("background_color", None) is not None and image.shape[-1] == 4:
+            background_color = outputs["background_color"]
+            image_ = image[:, :3] * image[:, 3:] + background_color.to(image.device) * (1 - image[:, 3:])
+        else:
+            image_ = image[:, :3]
+        loss_dict["rgb_loss"] = self.rgb_loss(image_, outputs["rgb"])
```
nepfaff, line 338: Can you add the same logic to [NGPModel](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/models/instant_ngp.py)? I found that this model works (at least with the `instant-ngp-bounded` default settings) works particularly well with alpha transparent training (at least for the masked mustard dataset). It might also be beneficial for other models but maybe not worth adding for the others until it becomes useful for people.

f-dy:
Just one concern: the option names and documentation. `use_alpha_channel` is a dataloader option, but we don't know what alpha is being used for. And the transparency loss is implemented only in nerfacto, and undocumented.

Alternative:
- alpha channel should be always loaded (so no use_alpha_channel option)
- models that don't implement transparency loss should discard it
- nerfacto should have an option to specifically use alpha for transparency loss
- other models can progressively implement transparency loss

f-dy:
Hum, this is not really implementing transparency carving, is it? Shouldn't there be a loss on the accumulation? What prevents solid stuff of the same color as the background to be reconstructed outside of the masks?

nepfaff:
The background color is random. The 

> Hum, this is not really implementing transparency carving, is it? Shouldn't there be a loss on the accumulation? What prevents solid stuff of the same color as the background to be reconstructed outside of the masks?

That is correct. I missed that before...
There shouldn't be a loss on the accumulation (as is done for a constant background color), but the background color should be added to both the target pixels (as done in this PR) and to the network prediction (not done yet). This then encourages the network to predict zero density to let the background color shine through.
I'm now quite surprised why this works so well with the mustard dataset and the `instant-ngp-bounded` model, even if this part of the implementation is missing. Rendering the accumulation in the viewer, the model indeed learns zero density where alpha is zero.

instantNGP does this [here](https://github.com/NVlabs/instant-ngp/blob/HEAD/src/testbed_nerf.cu#L1499).

SamDM:
Hi, I'd like to clear out a few things:

## About alpha carving:

> There shouldn't be a loss on the accumulation (as is done for a constant background color), but the background color should be added to both the target pixels (as done in this PR) and to the network prediction (not done yet). This then encourages the network to predict zero density to let the background color shine through.

Indeed, in both the ray color prediction as the target image, you need a matching random color that changes in every training evaluation. 

- setting the random background color for the network prediction (i.e. rendering) step happens [here](https://github.com/SamDM/nerfstudio/blob/7776355651a16f1e90242e083166f5ae47156e17/nerfstudio/models/nerfacto.py#L280C1-L285).
- setting the matching background color for the target image happens [here](https://github.com/SamDM/nerfstudio/blob/7776355651a16f1e90242e083166f5ae47156e17/nerfstudio/models/nerfacto.py#L335).

I have also tested penalizing the accumulation on transparent pixels, and got about the same results. The only downside of penalizing accumulation compared to using changing background colors is that you introduce another hyperparam for setting the weight of the accumulation loss.

## My training settings:

The options I used for training were:
```
--pipeline.datamanager.camera-optimizer.mode off
--pipeline.model.disable-scene-contraction True
--pipeline.model.background-color white
--viewer.image-format png
--logging.local-writer.max-log-size 0
nerfstudio-data
--scale-factor 1.0
--scene-scale 0.12
--center-method none
--orientation-method none
--auto-scale-poses False
--train-split-fraction 1.0
```

Also, you must add `use_alpha_channel=True` as a global option to the `cameras.json` file.

Unfortunately, I am not allowed to share the training data, as it is private company data.

nepfaff:
@SamDM You're right. I missed the rendering argument :) Looks good to me functionality-wise!

Code quality-wise, I agree with @f-dy's points. What do you think?

> One concern: the option names and documentation. `use_alpha_channel` is a dataloader option, but we don't know what alpha is being used for. And the transparency loss is implemented only in nerfacto, but undocumented.
> 
> Alternative:
> 
> * alpha channel should be always loaded (so no use_alpha_channel option)
> * models that don't implement transparency loss should discard it
> * nerfacto should have an option to specifically use alpha for transparency loss
> * other models can progressively implement transparency loss

nepfaff:
@SamDM I'd be willing to create a new PR with these changes if you don't want to work on it. Just let me know :) It would be nice to get this merged to be up to date with main.

nerfstudio/models/nerfacto.py
```diff
@@ -288,14 +276,21 @@ def get_outputs(self, ray_bundle: RayBundle):
...
         ray_samples_list.append(ray_samples)
 
-        rgb = self.renderer_rgb(rgb=field_outputs[FieldHeadNames.RGB], weights=weights)
+        if self.kwargs["metadata"].get("use_alpha_channel", False) and self.training:
+            background_color = torch.rand((1, 3))
+        else:
+            background_color = None
+        rgb = self.renderer_rgb(
+            rgb=field_outputs[FieldHeadNames.RGB], weights=weights, background_color=background_color
+        )
```
jkulhanek, line 285: I would suggest to use:
```
background_color = self.renderer_rgb.get_background_color(field_outputs[FieldHeadNames.RGB][:, 0, ...].shape)
rgb = self.renderer_rgb(
      rgb=field_outputs[FieldHeadNames.RGB], weights=weights, background_color=background_color
)
```

nerfstudio/models/nerfacto.py
```diff
@@ -327,15 +322,20 @@ def get_outputs(self, ray_bundle: RayBundle):
...
 
     def get_loss_dict(self, outputs, batch, metrics_dict=None):
         loss_dict = {}
         image = batch["image"].to(self.device)
-        loss_dict["rgb_loss"] = self.rgb_loss(image, outputs["rgb"])
+        if outputs.get("background_color", None) is not None and image.shape[-1] == 4:
+            background_color = outputs["background_color"]
+            image_ = image[:, :3] * image[:, 3:] + background_color.to(image.device) * (1 - image[:, 3:])
+        else:
+            image_ = image[:, :3]
```
jkulhanek, line 337: Can you please move the code to the renderer and call `self.renderer_rgb.blend_background(...)`?

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -296,16 +292,23 @@ def _generate_dataparser_outputs(self, split="train"):
             applied_scale = float(meta["applied_scale"])
             scale_factor *= applied_scale
 
+        if "use_alpha_channel" in meta:
```
jkulhanek, line 295: I think this can be removed from the dataparser and the code would always load the alpha-channel if present (only when alpha_color is set it will blend before training - not sure if it will be necessary though).

nerfstudio/data/dataparsers/base_dataparser.py
```diff
@@ -57,6 +57,8 @@ class DataparserOutputs:
     """Camera object storing collection of camera information in dataset."""
     alpha_color: Optional[Float[Tensor, "3"]] = None
     """Color of dataset background."""
+    use_alpha_channel: bool = False
```
jkulhanek, line 60: Please always keep alpha_channel if present and alpha_color is None.

SamDM:
> @SamDM I'd be willing to create a new PR with these changes if you don't want to work on it. Just let me know :) It would be nice to get this merged to be up to date with main.

Hi, I don't have much time to work on this now, so please go ahead, I'd be much obliged!

PR #2030, opened 2023-06-02
installation.md: Use https instead of ssh for git clone.

SharkWipf:
Https git clone can be run by anyone directly, without account, whereas ssh git on Github requires having a Github account set up.


tancik (approved):
LGTM

PR #2033, opened 2023-06-04
Add antialias parameter as default None

maturk:
I was getting some annoying warning messages about the functional.resize() function. Apparently pytorch is changing the default value of the 'antialias' parameter from None to True. Adding the parameter as None might be useful.

Link: https://pytorch.org/vision/main/generated/torchvision.transforms.functional.resize.html'

> "The current default is None but will change to True in v0.17 for the PIL and Tensor backends to be consistent."


tancik (approved):
LGTM

PR #2037, opened 2023-06-05
Optimizing camera undistortion

DreekFire:
Only runs Newton's method until sub-pixel accuracy is reached instead of all 10 iterations. Then resamples images with bilinear sampling to get pixel color at points with error.

Also uses analytical formulas for pixel areas instead of distorting coordinates of neighboring pixels to get pixel area.


DreekFire:
Need example of datasets with different camera types to finish testing.

nerfstudio/cameras/camera_utils.py
```diff
@@ -413,40 +417,226 @@ def radial_and_tangential_undistort(
...
+    if distortion_params.shape[-1] == 0:
+        return coords, torch.eye(2, device=coords.device), coords
+
+    if distortion_params.shape[-1] < 8:
+        distortion_params = F.pad(distortion_params, (0, 8 - distortion_params.shape[-1]), "constant", 0.0)
+    assert distortion_params.shape[-1] == 8
+
+    resolution = resolution.to(coords.device)
+    # n_samples = coords.shape[0]
+    # n_iters = 0
```
tancik, line 449: Remove commented out code, here and elsewhere

tancik (commented):
Nice! Added some initial comments.

nerfstudio/data/datamanagers/base_datamanager.py
tancik, line 1: Many of the changes in this file don't seem related to this PR

liruilong940607:
Thanks @DreekFire for the updates. This is not an easy one.

My main question is on the fisheye formula. Have you run any tests to verify the fisheye formula is correct? 

The reason I'm asking this is that the old version actually have a correct fisheye undistortion formula by combining 
https://github.com/nerfstudio-project/nerfstudio/blob/43c399e71ead6a9a733e29efa1c5966da3ff4629/nerfstudio/cameras/camera_utils.py#L411

and 

https://github.com/nerfstudio-project/nerfstudio/blob/43c399e71ead6a9a733e29efa1c5966da3ff4629/nerfstudio/cameras/cameras.py#L724-L735

And this PR seems to change the `radial_and_tangential_undistort` function but not the second part for fisheye. I'm not sure if the formula is correct anymore.

kerrj:
what machine are you testing on? is there any particular sort of hardware this implementation is optimized for? eg would you see stronger benefits on weaker CPUs or GPUs, etc

tancik:
@kerrj Will this conflict with any of the datamanager stuff that you are working on?

DreekFire:
Ended up removing the resampling because it didn't appear to be worth it - once a ray was already within sub-pixel accuracy, Newton's method usually achieved very good accuracy in just one more iteration. This makes the code simpler as well. Now, the key optimizations are early stopping for undistortion and analytical formulas for pixel areas instead of numerical ones (should also be more accurate because it correctly accounts for skewed pixels).

The speed fluctuates throughout training, but it is generally about 10% faster at any given iteration than the main branch on puppy.

Not sure why the tests are failing - seems to be something going wrong with torch.compile. It passes on both my local machine as well as my env on the puppy machine.

DreekFire:
Hi @anc2001, nerfstudio does not use all of the distortion parameters simultaneously. For perspective cameras, it uses two radial and two tangential distortion parameters, while for fisheye cameras, it uses 4 radial distortion parameters. The old formula produces results identical to the formulas implemented in OpenCV (OpenCV uses different formulas for perspective and fisheye cameras) as long as you do not try to use the third or fourth radial distortion parameters for a perspective camera, or any tangential distortion parameters for a fisheye camera.

anc2001:
Hi @DreekFire thanks for the reply. At the time of my original comment I was unsure that the current implementation of fisheye undistortion matched up with the math described in the OpenCV documentation. I worked it out and understand why it's correct now. Are you referring to the implementation on this branch or the current implementation on the main branch? To my understanding the current implementation uses 4 radial distortion parameters and 2 tangential where the 2 tangential are just 0 when the camera is fisheye. For perspective undistortion under the OpenCV distortion specification it will be incorrect if the 4th radial distortion parameter is included, but shouldn't it be correct up to the 3rd radial distortion parameter? Also shouldn't nerfstudio support the full OpenCV perspective camera model (maybe minus the thin prism coefficients)?

PR #2041, opened 2023-06-06
Fix bug when copying images when input_dir = output_dir

maturk:
Two things in this bug fix: 
1. Remove the output_directory iff it is not the same as the input_directory containing images (we don't want to delete valuable user data accidentally). 
2. Bypass shutil.SameFileError when copying files with a simple try catch. Works fine when input_dir = output_dir.

For some reason my ns-dev-test reformatted some of the longer arguments. I have a recent version of nerfstudio so not sure why it did that.

### DIFF
--- a/nerfstudio/process_data/process_data_utils.py
+++ b/nerfstudio/process_data/process_data_utils.py
@@ -262,11 +262,7 @@ def copy_and_upscale_polycam_depth_maps_list(
     depth_dir.mkdir(parents=True, exist_ok=True)
 
     # copy and upscale them to new directory
-    with status(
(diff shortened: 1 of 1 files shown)


tancik:
Have you rerun `pip install -e .[dev]`?

maturk:
> Have you rerun `pip install -e .[dev]`?

Yep. I think the problem was that I also had a format on save setting also applied to ruff in my vscode settings. I think this was messing with all the formatting and running ns-dev-test did not fix it for some reason. My local ruff was adding some more formatting it seems.

tancik (approved):
LGTM

PR #2047, opened 2023-06-08
Quick fix for eval script failing for depth-nerfacto (and other methods with a non-standard dataset)

jMysliwiec:
Quick fix for issue pointed out here:
https://github.com/nerfstudio-project/nerfstudio/issues/1992

It replaces the target loaded from the yaml config with the one with the correct generic type.


tancik (approved):
LGTM

Issue #2050, opened 2023-06-08
ns-eval, ns-render, ns-export load the whole training data (including images and transforms.json)


f-dy:
**Describe the bug**
The three commands `ns-eval`, `ns-render`, `ns-export` should not need the training data:
- `ns-eval` only need to load the test data (10% of the dataset by default)
- `ns-render` and `ns-export` do not need any training data

This is inefficient, especially when rendering or exporting from a pre-trained model on a different machine, where one may not want to transfer all training data with the checkpoint.

**To Reproduce**
any output of these three commands starts with:
```
Setting up training dataset...
Caching all 270 images.
```

**Expected behavior**
Only load data that is required :
- for ns-eval, transforms.jso + the test set
- for ns-render and ns-export, no transforms.json nor images

tancik:
I'm a little confused how the provided changes solves the issue?

f-dy:
commented in the PR. What confused me was why images were required at inference time :-)
(and btw huge congrats!)

PR #2051, opened 2023-06-08
nerfstudio dataparser: avoid loading images at inference time

f-dy:
Also remove the image skipping logic, which seems to be a remnant of some old code (all images should exist).
This allows running inference (s-render or ns-export) without the images.
Fixes https://github.com/nerfstudio-project/nerfstudio/issues/2050

### DIFF
--- a/nerfstudio/data/datamanagers/base_datamanager.py
+++ b/nerfstudio/data/datamanagers/base_datamanager.py
@@ -388,6 +388,8 @@ class VanillaDataManager(DataManager, Generic[TDataset]):
         else:
             self.config.data = self.config.dataparser.data
         self.dataparser = self.dataparser_config.setup()
+        if test_mode == "inference":
+            self.dataparser.downscale_factor = 1  # Avoid opening images
         self.includes_time = self.dataparser.includes_time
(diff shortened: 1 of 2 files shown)


tancik:
I'm a bit confused, I don't see how these changes solves the issue.

f-dy:
if the downscalefactore is None, _get_fname tries to open the image to compute it from the image resolution, here  https://github.com/nerfstudio-project/nerfstudio/blob/5511f5f4edf45a163a19ce2ff41e8b5fe2b9771e/nerfstudio/data/dataparsers/nerfstudio_dataparser.py#L322

f-dy:
and this check for file existence here https://github.com/nerfstudio-project/nerfstudio/blob/5511f5f4edf45a163a19ce2ff41e8b5fe2b9771e/nerfstudio/data/dataparsers/nerfstudio_dataparser.py#L112

I don't think it's any useful. all images should be available for training anyway

tancik (approved):
LGTM

pwais:
@f-dy this had been bugging me for a while, thank you so much for fixing!! 🎉

PR #2053, opened 2023-06-09
Add kwargs to dataparser outputs

maturk:
I found in my research that I needed some additional kwargs beyond "split" to generate dataparser outputs. Added kwargs as optional arguments.


tancik (approved):
LGTM

PR #2058, opened 2023-06-09
colormaps: add "gray"

f-dy:
- Add the trivial "gray" colormap, useful to eg extract accumulation.
- Don't show "pca" in the viewer menu (it crashes the server)


tancik:
The `PCA` was there for typing reasons. [This](https://github.com/nerfstudio-project/nerfstudio/blob/d01fef76502cf26eaaefe4283dfab60e57753de7/nerfstudio/viewer/server/control_panel.py#L349-L368) logic would also need to change . Maybe there should be a separate 1d colomaps list. Currently "PCA" and "Default" are weird.

f-dy:
@tancik I updated to re-add "pca" to the list, and fixed the viewer code to not show the "pca" option for single-channel images

tancik (approved):
LGTM

PR #2059, opened 2023-06-09
fix "ns-render interpolate", so that it can render the input views, add crop box

f-dy:
Previously, `ns-render interpolate`:
- Ignored distortion parameters
- Used an interpolated focal length, but principal point was from the first camera
- Rendered all images with the same size, even if input views were of different sizes
- Forgot to render the last frame
- There was no default support for crop box and background color, except when rendering from a camera path, where these would exclusively be fetched from the camera_path.json

Now:
- There is an command-line option `--fixed-intrinsics` to keep fixed intrinsics through the whole sequence, but they are all interpolated by default
- Everything is interpolated, even size, so that `ns-render interpolate --interpolation-steps 1` should render exactly the input images
- The last frame is rendered
- All render modes have the options `--use-bounding-box` `--background-color` `--bounding-box-min` `--bouding-box-max` (same syntax as ns-export), which can override the json settings when rendering a camera path.

Works great with the gray colormap from https://github.com/nerfstudio-project/nerfstudio/pull/2058 to render object masks.


tancik:
This is great!
Can you post example videos before and after.
Also how are the videos saved when the resolution changes?

tancik:
Hi, just checking in on this PR, looking forward to it being incorporated.

PR #1936, opened 2023-05-16
Ability to use vanilla pytorch instead of tcnn for Nerfacto 

ethanweber:
- [x] `--pipeline.model.implementation tcnn` is default but `--pipeline.model.implementation torch` can be used for vanilla pytorch. This will only work for the Nerfacto method at the moment.
- [x] Removed "nerfacto" from the blacklisted github actions tests.

The vanilla pytorch implementation is maybe ~4x slower. Looking into numerical equivalence is a good next step.

Example commands are as follows,

```
ns-train nerfacto --vis viewer+wandb --experiment-name aspen-tcnn --data data/nerfstudio/aspen --pipeline.model.implementation tcnn
```

and

```
ns-train nerfacto --vis viewer+wandb --experiment-name aspen-torch --data data/nerfstudio/aspen --pipeline.model.implementation torch
```

The quality looks very similar.

The main branch and this PR (which is "tcnn" here and still the default for nerfacto) should be equivalent. Here is a small sanity check.

image

PR #2063, opened 2023-06-12
Encoder + MLP combo

ethanweber:
In PR #1936, I removed the functionality of `tcnn.NetworkWithInputEncoding` and used `torch.nn.Sequential` instead. Here I revert that with a new class called `EncoderAndMLP`. This should be faster when using tccn with nerfacto. See "combo" below where I show some runs, though it doesn't show a significant (if any) improvement.

Screenshot 2023-06-11 at 11 05 14 PM
Screenshot 2023-06-11 at 11 05 19 PM


nerfstudio/field_components/mlp.py
```diff
@@ -102,38 +103,41 @@ def __init__(
             self.build_nn_modules()
         elif implementation == "tcnn" and not TCNN_EXISTS:
             print_tcnn_speed_warning("MLP")
+            self.build_nn_modules()
```
ethanweber, line 106: I fixed this in this PR too!

nerfstudio/fields/nerfacto_field.py
```diff
@@ -124,24 +124,27 @@ def __init__(
...
+                "min_res": base_res,
+                "max_res": max_res,
+                "log2_hashmap_size": log2_hashmap_size,
+                "features_per_level": features_per_level,
+            },
+            mlp_params={
+                "num_layers": num_layers,
+                "layer_width": hidden_dim,
+                "out_dim": 1 + self.geo_feat_dim,
+                "activation": nn.ReLU(),
```
kerrj, line 143: it seems like this should be passing nn.ReLU instead of an instance, is there a reason it needs to call the constructor here?

ethanweber:
I cleaned up the code after seeing issue #2348. I tested with the dozer scene from the Nerfstudio Dataset. Running nerfacto, I get these memories with the combined MLP and hash encoder vs. sequential.
- 2901MB with combo
- 2937MB with torch sequential

tancik (approved):

yurkovak:
Nice that it's supposed to be faster! Just note that it breaks older checkpoints, and the default unstrict loading causes silent skipping of parts of the model, leading to renders predictions that are hard to explain.

PR #2066, opened 2023-06-12
Monodepth Regularization

AdamRashid96:
Changes
- Fixed up `DepthDataset` so that it will now generate monodepth images outputted by [ZoeDepth](https://github.com/isl-org/ZoeDepth) if no depth images are found in the dataset, otherwise just loads them.
- Added a `PairPixelSampler` that will sample pairs of rays within a radius of distance from each other, useful for pair-based losses.
- Added a `depth_ranking_loss` that distills local depth ranking priors from coarse depth maps into the NeRF, implemented from [SparseNeRF](https://sparsenerf.github.io/).

Notes:
- Added the depth generation and regularization to nerfacto, should probably be another method or a flag to disable the depth stuff. Personally, I think a flag makes more sense but open to any opinions.
- ZoeDepth requires an older version of `timm`, I am using `timm==0.6.7` in `pyproject.toml`
- Currently hardcoding the cuda device in the `DepthDataset`, which could cause problems if the device is specified for training, because I'm not sure of an elegant solution to get the device without adding an argument. 
- Currently I replaced the `PatchPixelSampler` in `_get_pixel_sampler` with `PairPixelSampler`, but we probably want to add a flag to specify what type of sampler we use, or even a samplertype class where you can specify the type and argument for sampler.

Example:
RGB
Screen Shot 2023-06-12 at 3 54 47 PM

Before Depth Ranking
Screen Shot 2023-06-12 at 3 42 34 PM

After Depth Ranking
Screen Shot 2023-06-12 at 3 50 43 PM


tancik:
FYI, this is the issue from the Read the Docs failure
```
/home/docs/checkouts/readthedocs.org/user_builds/plenoptix-nerfstudio/checkouts/2066/nerfstudio/data/pixel_samplers.py:docstring of nerfstudio.data.pixel_samplers.PairPixelSampler:3: WARNING: Definition list ends without a blank line; unexpected unindent.
```

nerfstudio/models/nerfacto.py
```diff
@@ -290,12 +294,15 @@ def get_outputs(self, ray_bundle: RayBundle):
 
         rgb = self.renderer_rgb(rgb=field_outputs[FieldHeadNames.RGB], weights=weights)
         depth = self.renderer_depth(weights=weights, ray_samples=ray_samples)
+        with torch.no_grad():
+            depth_med = self.renderer_depth_med(weights=weights, ray_samples=ray_samples)
```
tancik, line 298: It doesn't look like this is ever used

nerfstudio/data/datasets/depth_dataset.py
```diff
@@ -31,14 +40,50 @@ class DepthDataset(InputDataset):
...
-            "depth_filenames" in dataparser_outputs.metadata.keys()
-            and dataparser_outputs.metadata["depth_filenames"] is not None
-        )
+        # if there are no depth images than we want to generate them all with zoe depth
+        device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        if len(dataparser_outputs.image_filenames) > 0 and (
+            "depth_filenames" not in dataparser_outputs.metadata.keys()
+            or dataparser_outputs.metadata["depth_filenames"] is None
+        ):
+            depth_paths = []
```
tancik, line 51: We should add a warning  so that users know this is happening,
```
CONSOLE.print("[bold yellow] No depth data found!")

File: nerfstudio/data/datasets/depth_dataset.py [hunk 3]
```diff
@@ -31,14 +40,50 @@ class DepthDataset(InputDataset):
         scale_factor: The scaling factor for the dataparser outputs.
     """
 
-    exclude_batch_keys_from_device = InputDataset.exclude_batch_keys_from_device + ["depth_image"]
-
     def __init__(self, dataparser_outputs: DataparserOutputs, scale_factor: float = 1.0):
```
tancik, line 43: Add to the class description that the depth dataset will generate data using the Zoe model if it is not found.

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -432,7 +432,7 @@ def create_eval_dataset(self) -> TDataset:
     def _get_pixel_sampler(self, dataset: TDataset, *args: Any, **kwargs: Any) -> PixelSampler:
         """Infer pixel sampler to use."""
         if self.config.patch_size > 1:
-            return PatchPixelSampler(*args, **kwargs, patch_size=self.config.patch_size)
+            return PairPixelSampler(*args, **kwargs, radius=self.config.patch_size)
```
tancik, line 435: I don't think we want to make the assumption that the user wants the Pair sampler over the Patch sampler.

nerfstudio/configs/method_configs.py
```diff
@@ -96,6 +96,8 @@
     mixed_precision=True,
     pipeline=VanillaPipelineConfig(
         datamanager=VanillaDataManagerConfig(
+            _target=VanillaDataManager[DepthDataset],
```
tancik, line 99: This should be added to `depth-nerfacto` unless we find that it is consistently better and the speed impact is negligible.

tancik (commented):
Results look great!

nerfstudio/models/nerfacto.py
AdamRashid96: I added this because I need average depths for the depth rank loss, but median depth is more useful to visualize. It's being added to the outputs, for visualization.

nerfstudio/configs/method_configs.py
AdamRashid96: Regarding speed I find that it doesn't really affect the speed at all, since the loss is only like 4 vector operations.

This is an example of training a scene and the depth rank loss kicks in at 500 training steps. You can see here the average speed doesn't really change, maybe a few more ms per step.

Screen Shot 2023-06-13 at 11 30 35 PM

Regarding the quality, I have found from the scenes that we have used it on that it has only helped, but our scenes were primarily tabletop scenes that lack a lot of detail. From the few other scenes I've tried it on, it can smooth out fine features, like the bristles in the brush that I sent above are slightly smoothed out. Definitely requires more testing to get an understanding of how it affects quality.

nerfstudio/data/datamanagers/base_datamanager.py
AdamRashid96: Yeah I agree, do you have any suggestions of the best way to switch between the two samplers. It could be a flag or another class defined in the config file.

kerrj: I think one option is just defining a new SamplingConfig which is future proof to all the random sampling changes we might want (eg this would make it easier to implement sampling wrt image frequency, edges, etc)

kerrj: this might involve refactoring the way get_pixel_sampler works in favor of making a tyro config

nerfstudio/configs/method_configs.py
kerrj: Yea it's better to keep it constrained to depth-nerfacto, it'll wreck performance on things like transparent scenes, fluid/smoke effects, maybe thin features like leaves/foliage, and no clue how it would do on strange objects/scenes.

nerfstudio/models/nerfacto.py
```diff
@@ -352,6 +359,11 @@ def get_loss_dict(self, outputs, batch, metrics_dict=None):
                 loss_dict["pred_normal_loss"] = self.config.pred_normal_loss_mult * torch.mean(
                     outputs["rendered_pred_normal_loss"]
                 )
+
+            if "depth_image" in batch and self.step >= 500:
```
kerrj, line 363: Might want to get rid of the 500+ thing, it's a strange hyperparam (scheduling from 0 to 2000 or so is reasonable though)

nerfstudio/models/nerfacto.py
kerrj: We should probably change the behavior so that "depth" is median depth and "expected_depth" is the expected one which gradients go through, otherwise downstream code could break like pointcloud generation

nerfstudio/models/nerfacto.py
kerrj, line 1: Philosophically we should probably not change nerfacto.py and just edit depth_nerfacto.py. Maybe there we can have a string which toggles between ranking distillation and metric depth distillation? That file is also sort of outdated, could use a bit of cleanup

nerfstudio/configs/method_configs.py
AdamRashid96: Moved the loss to depth_nerfacto and removed it from nerfacto, added a check to switch between metric depth loss and rank depth loss, if depth is generated or loaded by zoe_depth

nerfstudio/models/nerfacto.py
AdamRashid96: Made it scheduling from 0 to 2000 and looks like the results are good

nerfstudio/models/nerfacto.py
AdamRashid96: Moved the loss to depth_nerfacto and removed it from nerfacto, added a check to switch between metric depth loss and rank depth loss, if depth is generated or loaded by zoe_depth

nerfstudio/data/datamanagers/base_datamanager.py
AdamRashid96: Added `PixelSamplerConfig` to define which pixel sampler to use in the datamanager, removed `EquirectangularPixelSampler` and moved its sample method to `PixelSampler` and added a check that if all images are equirectangular then use equirectangular sampling.

nerfstudio/configs/sampling_config.py
```diff
@@ -0,0 +1,32 @@
...
+from dataclasses import dataclass
+from typing import Optional
+
+from nerfstudio.configs.base_config import (
+    InstantiateConfig,
+)
+
+
+@dataclass
+class SamplingConfig(InstantiateConfig):
```
kerrj, line 26: I would move this inside the pixel_samplers.py so they're in the same file as the configs, /configs is mostly used for larger method config files

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -338,8 +338,8 @@ class VanillaDataManagerConfig(DataManagerConfig):
     """The scale factor for scaling spatial data such as images, mask, semantics
     along with relevant information about camera intrinsics
     """
-    patch_size: int = 1
-    """Size of patch to sample from. If >1, patch-based sampling will be used."""
+    sampling_config: PixelSamplerConfig = PairPixelSamplerConfig()
```
kerrj, line 341: This should default to the standard PixelSampler

nerfstudio/configs/sampling_config.py
kerrj: Actually it doesn't seem like this is used? is it residual from an older version?

AdamRashid96: ya I should just delete that file

ethanweber (commented):
Cool PR! Leaving a comment on naming to be consistent with our method definitions.

ethanweber:
Closing in favor of PR https://github.com/nerfstudio-project/nerfstudio/pull/2284 with the GitHub actions issue fixed.

cv-lab-x:
> Changes
> 
> * Fixed up `DepthDataset` so that it will now generate monodepth images outputted by [ZoeDepth](https://github.com/isl-org/ZoeDepth) if no depth images are found in the dataset, otherwise just loads them.
> * Added a `PairPixelSampler` that will sample pairs of rays within a radius of distance from each other, useful for pair-based losses.
> * Added a `depth_ranking_loss` that distills local depth ranking priors from coarse depth maps into the NeRF, implemented from [SparseNeRF](https://sparsenerf.github.io/).
> 
> Notes:
> 
> * Added the depth generation and regularization to nerfacto, should probably be another method or a flag to disable the depth stuff. Personally, I think a flag makes more sense but open to any opinions.
> * ZoeDepth requires an older version of `timm`, I am using `timm==0.6.7` in `pyproject.toml`
> * Currently hardcoding the cuda device in the `DepthDataset`, which could cause problems if the device is specified for training, because I'm not sure of an elegant solution to get the device without adding an argument.
> * Currently I replaced the `PatchPixelSampler` in `_get_pixel_sampler` with `PairPixelSampler`, but we probably want to add a flag to specify what type of sampler we use, or even a samplertype class where you can specify the type and argument for sampler.
> 
> Example: RGB Screen Shot 2023-06-12 at 3 54 47 PM
> 
> Before Depth Ranking Screen Shot 2023-06-12 at 3 42 34 PM
> 
> After Depth Ranking Screen Shot 2023-06-12 at 3 50 43 PM

HI, thanks for your work, when i test the depth-nerfacto with depthranking loss on my custom datasets, the depthranking loss increases all the time, i use the zeodepth as default, --pipeline.model.depth-loss-mult 0.2 as the sparsenerf used.  what config params do you recommend for depth-nerfacto with depthranking loss. Looking forward to your reply! ps:training only with rgb is ok. @AdamRashid96

PR #2080, opened 2023-06-15
Add non-trainable camera indices to CameraOptimizer

maturk:
Here I added the possibility for a user to manually tell the Camera Optimizer not to train certain camera indices. This is needed when you know prior that only certain indices should be trained in the camera training_dataset.

### DIFF
--- a/nerfstudio/cameras/camera_optimizers.py
+++ b/nerfstudio/cameras/camera_optimizers.py
@@ -123,7 +123,7 @@ class CameraOptimizer(nn.Module):
             outputs.append(exp_map_SE3(self.pose_adjustment[indices, :]))
         else:
             assert_never(self.config.mode)
-        # Detach non-trainable indices from training by setting to identity transform.
+        # Detach non-trainable indices by setting to identity transform
         if self.non_trainable_camera_indices is not None:
             outputs[0][self.non_trainable_camera_indices] = torch.eye(4)[:3, :4].to(self.device)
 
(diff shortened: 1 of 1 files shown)


nerfstudio/cameras/camera_optimizers.py
```diff
@@ -121,11 +123,13 @@ def forward(
             outputs.append(exp_map_SE3(self.pose_adjustment[indices, :]))
         else:
             assert_never(self.config.mode)
+        # Detach non-trainable indices by setting to identity transform
+        if self.non_trainable_camera_indices is not None:
+            outputs[0][self.non_trainable_camera_indices] = torch.eye(4)[:3, :4].to(self.device)
```
tancik, line 128: Create on device
`torch.eye(4, device=self.device)`

kerrj:
LGTM with Matt's change, useful in robotics when you want to prevent the coord frame from drifting too much thru optimization. I wonder if eventually it could be good to optionally input a covariance matrix for each camera, then you can even have a per-camera confidence :). Not sure how you'd incorporate that into the optimization, but maybe just some simple regularization like weighting loss higher on lower confidence cameras

maturk:
Regarding the camera optimizer, on the top of the file it says `Pose and Intrinsics Optimizers` but there is no intrinsics optimization currently right? I might look into that as well since I think I will need it...

kerrj:
Yes there is no intrinsic optimization. We're also thinking about refactoring camera optimization to move it out of the datamanager and into the Model, since this allows more natural development of different methods (BARF etc), and also would help speed up dataloading (right now ray generation cannot be parallelized because gradients need to flow from model to datamanager). Might be good to keep this in mind, and if you have thoughts on a refactor we'd love to hear them too

kerrj (approved):
lgtm

PR #2082, opened 2023-06-15
reduce nerfacto-huge memory

kerrj:
Reduced memory usage of nerfacto-huge to fit on a 24GB GPU.
Screen Shot 2023-06-15 at 4 06 39 PM


brentyi (approved):

machenmusik:
A little late, but I think that there is also a 16GB threshold that would be useful for those on consumer GPUs that max out there.  Maybe a nerfacto-lesshuge

kerrj:
nerfacto-big is supposed to target that audience, but I think it could use some more tuning to reach the threshold.

PR #2090, opened 2023-06-17
Fix get_num_frames_in_video

olegpokhilchenko:
Tests have shown that there can be 2 exact same numbers in the output, separated by non-digit symbols. I think it's better to just find the first number.

### DIFF
--- a/nerfstudio/process_data/process_data_utils.py
+++ b/nerfstudio/process_data/process_data_utils.py
@@ -17,6 +17,7 @@
 import math
 import shutil
 import sys
+import re
 from enum import Enum
 from pathlib import Path
 from typing import List, Literal, Optional, OrderedDict, Tuple, Union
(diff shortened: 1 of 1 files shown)


tancik:
What are you seeing as the output of ffmpeg? I wan't to get a better idea about when it fails.

olegpokhilchenko:
Yes, in some cases ffprobe can give results like this: 123123\n\n123123.
For example, the AVCHD codec in MTS format does this.

tancik:
This doesn't work for me. I ran a test where it should return 635 frames, but it only returned 6.

olegpokhilchenko:
There was an error in the code: `output[0]` was referencing the initial command output, which is not what we want. I changed it to `number_match[0]`. This means "take the regex match and return the group 0 of it (whole match)".

PR #2092, opened 2023-06-18
Move camera optimization out of datamanager and parallelize dataloading

kerrj:
WIP for removing camera optimization from datamanger and moving it to the model. This will potentially allow significant speedups in dataloading as several people have already noted, since the most time consuming step in dataloading is ray generation, which currently depends on gradients from the camera optimizer and thus cannot be parallelized.

Currently this
- Removes all camera optimization code from datamanager
- Adds camera optimization to nerfacto.py and refactors param groups, optimizer config accordingly

Things that need to happen:
- Think about how to merge this PR without breaking everyone's custom method configs which rely on the camera optimizer being inside the datamanager
- Think about proper inheritance for using camera optimizer (currently it sits in nerfacto.py, but it might be better to move it inside base_model.py)
- Parallelize all the ray generation code inside a proper dataloader


maturk:
Hey @kerrj, I would like to help you with this PR. I have a few questions:
1. Regarding the pose regularizer, in the previous nerfstudio version there was no explicit loss on the poses and the main loss was just the nerf reconstruction loss. I like the idea of L2 regularization on the poses, did you see any improvements in the training when using it? I am guessing it is quite negligible in most use cases.
2. Regarding the actual training of the camera optimizer, how would one add their own scheduler and optimizer for the CameraOptimizer class? Several papers use different schedules/steps when training the poses which is a bit separate from the main nerf training pipeline.

I don't think I am able to directly commit/make changes in this branch but I will test out the changes at least.

nerfstudio/viewer_beta/viewer.py
```diff
@@ -190,16 +211,25 @@ def init_scene(self, dataset: InputDataset, train_state: Literal["training", "pa
...
+                name=f"/camera_{idx:05d}",
+                fov=2 * np.arctan(camera.cx / camera.fx[0]),
+                scale=1,
                 aspect=float(camera.cx[0] / camera.cy[0]),
                 image=image_uint8,
                 wxyz=R.wxyz,
                 position=c2w[:3, 3] * VISER_NERFSTUDIO_SCALE_RATIO,
             )
 
+            @camera_handle.on_click
```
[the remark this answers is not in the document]
tancik: We shouldn't assume locally build viser in nerfstudio code. The viser version should be bumped and then updated in the pyproject here.

nerfstudio/models/nerfacto.py
```diff
@@ -163,6 +166,9 @@ def populate_modules(self):
             implementation=self.config.implementation,
         )
 
+        self.camera_optimizer: CameraOptimizer = self.config.camera_optimizer.setup(
+            num_cameras=self.num_train_data, device="cpu"  # for now initialize on cpu, these will be moved later
```
maturk, line 170: This is very confusing for me. The camera_optimizer.device attribute is set at initialization as "cpu". But during training, attributes like self.pose_adjustment are moved to the gpu, so internal attributes are no longer on the "self.device" device. Is this intended? Should we somehow check that self.device changes inside the camera optimizer when e.g. self.pose_adjustment tensors get moved to the gpu?

maturk: > Yes this device behavior has always been confusing to me too, I'm not sure what happens to the device parameter when you call .cuda() on the an nn.Module, does it also get updated to the gpu?

All parameters within a module will go to "gpu" but this user defined "self.device" param will stay as "cpu". It is not so important, just noticed that I could not rely on the self.device var when considering where I should send (gpu/cpu) my "non_trainable_indices" ints. I will check out your parallel data manager, awesome work!

nerfstudio/data/datamanagers/parallel_datamanager.py
```diff
@@ -0,0 +1,220 @@
...
+            CONSOLE.print("[bold yellow]Warning: Some cameras are equirectangular, but using default pixel sampler.")
+        return PixelSampler(*args, **kwargs)
+
+    def next_train(self, step: int) -> Tuple[RayBundle, Dict]:
+        self.train_count += 1
+
+        # Fetch the next batch in an executor to parallelize the queue get() operation
+        # with the train step
+        bundle, batch = self.batch_fut.result()
+        self.batch_fut = self.executor.submit(self.data_q.get)
```
pwais, line 202: will this work for multi-gpu distributed data parallel (ddp) mode?  (do you need multi-gpu testing support?)   

can this class now just be a torch DataLoader, which I think would automatically support ddp?  i guess i thought the original `DataManager` was not a DataLoader and is a `nn.Module` to afford for camera optimizer / gradients

maturk:
It's fast! Tested on Nvidia 4090 GPU on the poster scene with camera opt enabled.

Parallel: avg. train rays/sec = 266460, gpu process memory allocation: 22.17%
Base: avg. train rays/sec = 197794, gpu process memory allocation: 29.23%

Parallel manager is about 36% faster and uses less memory! Can we make it even faster :D

baseVSparallel

MvWouden:
Could you perhaps try it out for e.g. TensoRF with and without CO?
Currently the CO is broken for these methods, I'm curious to see if this PR manages to fix it (whether intended or unintended).

Related issue: https://github.com/nerfstudio-project/nerfstudio/issues/2116

maturk:
@tancik & @kerrj. This PR currently breaks backward compatibility with methods registered with nerfstudio if older methods have camera optimization inside the datamanager (within the VanillaDatamanagerConfig). How should we handle this incompatibility? Is it okay to break backwards compatibility with this PR?

kerrj:
Getting back to this, I think it's ok to break backwards compatability in config files as long as we give a clear example of how to reformat your configs (in practice it's very similar to the old version, just requires moving some chunks of config). @tancik @brentyi  thoughts?

tancik:
I think it is something we want to support. We can bump to v0.4.0 and make it clear that previous models are not supported. Alternatively, we can add logic that converts old configs if they are detected.

brentyi:
> Alternatively, we can add logic that converts old configs if they are detected.

If this (or at least an informative message) is possible I'd put a strong vote for it. It's nice to not break external methods, particularly given the recent push to migrate them out of the core repo.

Ilyabasharov:
@maturk I see that you added BARF in [NerfactoField](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/fields/nerfacto_field.py). Do you think that we should also add bundle adjustment in [HashMLPDensityField](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/fields/density_fields.py)?

maturk:
> @maturk Hello! I see that you added BARF in [NerfactoField](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/fields/nerfacto_field.py). Do you think that we should also add bundle adjustment in [HashMLPDensityField](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/fields/density_fields.py)?

I don't think so. My barf implementation was really just a small test I wanted to play with, I think it does not belong in this pr since it does not improve psnr on already good colmap poses + it only works for really bad poses + it further breaks backward compatibility for subclassed nerfactos.

maturk:
Deprication warning added to VanillaDataManagerConfig. I believe there is a more elegant solution using @dataclass(kw_arg=True) or [dataclasses.KW_ONLY](https://docs.python.org/3.10/library/dataclasses.html#dataclasses.KW_ONLY) but this is only available in python 3.10. Right now, a red warning is given to the user if they initialize a VanillaDatamanagerConfig with a camera_optimizer field. Since the camera optimizer has been moved away from this class, camera optimization does not occur and this is also mentioned in the warning. This does not entirely fix backwards compatibility, because camera opt will indeed be OFF even if older methods want it to be ON, but at least it does not break nerfstudio and pylance is happy...

ksnzh:
https://github.com/nerfstudio-project/nerfstudio/blob/justin/camera_opt_refactor/nerfstudio/pipelines/base_pipeline.py#L363 may be fixed to `assert isinstance(self.datamanager, (VanillaDataManager, ParallelDataManager))`

maturk:
I think this is ready for merge. Tested with Lerf repository out of the box, and warnings appeared as they should without breaking the pipeline. Users would have to pip install nerfstudio once again after this update, since viser version has been bumped up.

brentyi (approved):

maturk:
pytest failed :/

PR #2101, opened 2023-06-20
Fix quickstart doc for remote host viewing setup

jlightfoot:
SSH command should specific localport:remotehost:remoteport rather than localport:localhost:remoteport in -L switch.  Also changed description for clarity.


tancik (approved):
LGTM

PR #2108, opened 2023-06-20
expose downscale-factor

machenmusik:
This PR exposes downscale-factor as an adjustable parameter `--pipeline.datamanager.downscale-factor
` to allow experimentation (since it is not exposed / effective via pipeline.datamanager.dataparser)

### DIFF
--- a/nerfstudio/data/datamanagers/base_datamanager.py
+++ b/nerfstudio/data/datamanagers/base_datamanager.py
@@ -113,6 +113,8 @@ class DataManagerConfig(InstantiateConfig):
     """Specifies the camera pose optimizer used during training. Helpful if poses are noisy."""
     masks_on_gpu: Optional[bool] = None
     """Process masks on GPU for speed at the expense of memory, if True."""
+    downscale_factor: Optional[int] = None
+    """Downscale factor to use, automatic if not specified."""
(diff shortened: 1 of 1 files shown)


jkulhanek:
Is this a shortcut for pipeline.datamanager.dataparser.downscale_factor?

machenmusik:
That is where it should be IMO, but when I try using that with `ns-train nerfacto` it gives errors (doesn't work), thus the datamanager override

tancik:
It should be accessible via `ns-train nerfacto ... nerfstudio-data --downscale-factor 2`.
The base dataparser doesn't require `downscale-factor` to be implemented, so this could fail for some dataparsers, https://github.com/nerfstudio-project/nerfstudio/blob/ef1350bd3627afc515ed86a379ceea676ed5850a/nerfstudio/data/dataparsers/base_dataparser.py#L115-L121

machenmusik:
My recollection is that other dataparsers support it as well, although not required (maybe copied from 
nerfstudio loader), and so this could provide a uniform method to access the parameter.  Perhaps this would not be necessary or useful if nerfstudio-data were default config rather than blender...

Feel free to close if you prefer current methodology.

PR #2110, opened 2023-06-21
Add option to process masks on GPU for speed at the expense of memory

### LINKED ISSUE #2075: masked (mustard) dataset much slower with main than PR2025
On my machine
- using the mustard dataset with masks (see https://github.com/nerfstudio-project/nerfstudio/issues/1498#issue-1599115766 ) with PR #2025 had default nerfacto iteration times around 80 to 90 ms
- using same dataset with main (currently https://github.com/nerfstudio-project/nerfstudio/commit/d18588d5d43fafa87aa819cdb3dc5b4e09f05549 ) has default nerfacto iteration times around 400 ms
so 4-5x worse.

machenmusik:
Closes #2075 and friends, when correct option value is supplied.
`--pipeline.datamanager.masks-on-gpu True`


tancik (approved):
LGTM

PR #2111, opened 2023-06-21
Fixing the image_dir if using skip_image_processing

Khoa-NT:
If we use `--skip-image-processing` with the `data=folder_A/images` and `output_dir=folder_B`, then both colmap and hloc will still assume that the input images are located in `folder_B/images`, even though it does not exist.


tancik (approved):
LGTM

Issue #2081, opened 2023-06-15
Error in spherical harmonics implementation


thetianshuhuang:
**Description**

The spherical harmonics implementation has some math errors in components (4,-1) and (4,4).

[Component (4,-1)](https://github.com/nerfstudio-project/nerfstudio/blob/8fad92c39af4e2d992dfdf950d4cff4f4478c3c9/nerfstudio/utils/math.py#LL83C1): a straightforward missing variable.
```
components[..., 19] = 0.6690465435572892 * y * (7 * zz - 3)
# should be
components[..., 19] = 0.6690465435572892 * y * z * (7 * zz - 3)
```
[Component (4,4)](https://github.com/nerfstudio-project/nerfstudio/blob/8fad92c39af4e2d992dfdf950d4cff4f4478c3c9/nerfstudio/utils/math.py#L88C1): the coefficient appears to be the imaginary coefficient `3/16 * sqrt(35 / (2pi)` instead of the real coefficient `3/16 * sqrt(35 / pi)`. 
```
components[..., 24] = 0.4425326924449826 * (xx * (xx - 3 * yy) - yy * (3 * xx - yy))
# should be
components[..., 24] = 0.6258357354491761 * (xx * (xx - 3 * yy) - yy * (3 * xx - yy))
```

**To Reproduce**

This can be checked through numerical integration; I wrote a quick test harness (pardon the modified code -- I found this bug only when working on a different project *based* on this code, not actually using it), which can be used to verify that the spherical harmonic coefficients are orthonormal:
```
from matplotlib import pyplot as plt
from jax import numpy as jnp
import numpy as np
from jax import vmap


def spherical_harmonics(dx, harmonics: int = 25):
    """Compute spherical harmonic coefficients.

    Parameters
    ----------
    dx: offset direction from the point to the sensor (on the unit sphere).
    harmonics: number of coefficients to compute.

    References
    ----------
    https://en.wikipedia.org/wiki/Table_of_spherical_harmonics
    https://github.com/nerfstudio-project/nerfstudio:
        nerfstudio/utils/math.py
    """
    x, y, z = dx
    xx = x**2
    yy = y**2
    zz = z**2

    coef = [0.28209479177387814]
    if harmonics >= 4:
        coef += [
            0.4886025119029199 * y,
            0.4886025119029199 * z,
            0.4886025119029199 * x]
    if harmonics >= 9:
        coef += [
            1.0925484305920792 * x * y,
[...]

PR #2112, opened 2023-06-21
Fix spherical harmonics #2081

jkulhanek:
Implements fix from #2081

Thanks @thetianshuhuang for finding the bug and providing the tests!


tancik (approved):
Lgtm

PR #2113, opened 2023-06-21
Allow newer pycolmap for newer python

jkulhanek:
PyColmap version 0.3.0 seems not to be supported in Python >=3.10, preventing nerfstudio[dev] from installing. This PR allows the newer version 0.4.0 to be installed instead.


tancik (approved):
Lgtm

PR #2119, opened 2023-06-22
allow ns-process-data to use CR2 raw images

machenmusik:
(Note, this may allow other raw image formats to work as well, but I only have some CR2 that I can't share at present.)

### DIFF
--- a/nerfstudio/process_data/process_data_utils.py
+++ b/nerfstudio/process_data/process_data_utils.py
@@ -38,6 +38,7 @@ ALLOWED_RAW_EXTS = [".cr2"]
 """Suffix to use for converted images from raw."""
 RAW_CONVERTED_SUFFIX = ".tiff"
 
+
 class CameraModel(Enum):
     """Enum for camera types."""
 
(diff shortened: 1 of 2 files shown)


nerfstudio/process_data/process_data_utils.py
```diff
@@ -207,7 +210,15 @@ def copy_images_list(
             CONSOLE.log(f"Copying image {idx + 1} of {len(image_paths)}...")
         copied_image_path = image_dir / f"frame_{idx + 1:05d}{image_path.suffix}"
         try:
-            shutil.copy(image_path, copied_image_path)
+            # if CR2 raw, we want to read raw and write TIFF, and change the file suffix for downstream processing
+            if str(image_path).lower().endswith(".cr2"):
```
tancik, line 214: `image_path.suffix().lower() == ".cr2"`

tancik (commented):
Seems worthwhile, thanks for adding. Added a minor comment.

machenmusik:
Do others have non-Canon raw image samples that should also be supported?  CR2 is one of the Canon suffixes, but others such as Nikon and Sony should work as well (in theory).

machenmusik:
Made it easy to add other suffixes to treat as raw images.

tancik (approved):
LGTM

PR #2124, opened 2023-06-23
Fix to fall back to torch if tcnn not found

ethanweber:
`self.build_nn_modules()` was not present where it should be, so this adds it back. The logic is to fall back to torch implementation if tcnn is specified but not installed/found.


tancik (approved):
Lgtm

PR #2125, opened 2023-06-23
ns-render with depth colormap support

ethanweber:
This adds the ability to use a depth colormap and specify the min and max depth values to consider (optionally). I also move `rendered_output_names` to the `BaseRender` config.


nerfstudio/scripts/render.py
```diff
@@ -149,14 +153,27 @@ def _render_trajectory_video(
                         )
                         sys.exit(1)
                     output_image = outputs[rendered_output_name]
-                    output_image = (
-                        colormaps.apply_colormap(
-                            image=output_image,
-                            colormap_options=colormap_options,
+                    if "depth" in rendered_output_name:
```
tancik, line 156: I'm not a huge fan of adding special logic to outputs like this. It could be confusing to users why the "depth" output looks different than a custom output like "depth_median". Maybe we can add logic in the colormap options where a boolean could exist for `is_depth_colormap`.

ethanweber: In this code `rendered_output_name` is a string, so maybe writing `if rendered_output_name.find("depth") != -1:` would be less confusing. Do you think that change would suffice? It follows similar logic we have throughout the codebase, although of course it's not ideal. By default, the colormaps should look the same since the depth min and max are set to None.

ethanweber: I'm going to do a check with `is_depth = rendered_output_name.find("depth") != -1`. I think modifying ColormapOptions isn't so easy now that it's passed throughout `render.py`.

ethanweber:
Here are 3 different results using different near and far planes for the depth colormap.

`ns-render ... --rendered_output_names rgb depth --downscale_factor 4`
https://github.com/nerfstudio-project/nerfstudio/assets/16965789/a067131f-9bad-48d3-a5f3-4650700a134a

`ns-render ... --rendered_output_names rgb depth --downscale_factor 4 --depth_near_plane 0.25 --depth_far_plane 2.0`
https://github.com/nerfstudio-project/nerfstudio/assets/16965789/d2ee3ccd-141a-4cfd-bbb0-9c12f9d2b31c

`ns-render ...  --rendered_output_names rgb depth --downscale_factor 4 --depth_near_plane 1.5 --depth_far_plane 2.0`
https://github.com/nerfstudio-project/nerfstudio/assets/16965789/ab012eac-e52d-4f41-acc8-f93cf551e509

PR #2139, opened 2023-06-28
add `build-arg` cuda version to dockerfile

Ilyabasharov:
faced with the fact that a certain version of сuda was unavailable for me (11.8), so I set the parameter editable from outside


tancik (approved):
LGTM

PR #2145, opened 2023-06-29
Fix unnecessary encodings calculation in tensorf field

kst179:
Found this bug in tensorf field: if using shperical harmonics (`use_sh=True`), then directions/features encodings are calculated but not used, hence wasting resources.


tancik (approved):
Good find. LGTM

PR #2146, opened 2023-06-29
Simplify external projects, unify plugin API

jkulhanek:
- Moves the AnnotatedDataParserConfig out of VanillaDataManagerConfig which simplifies custom methods as there won't be circular imports when DataParser is in the same file as the Model.
- Moves Annotated method configs out of nerfstudio.config.method_configs and users can extend existing configuration files in external methods.
- Unifies the discover_methods, discover_dataparsers API.

### DIFF
--- a/nerfstudio/configs/annotated_types.py
+++ b/nerfstudio/configs/annotated_types.py
@@ -1,0 +1,139 @@
+# Copyright 2022 the Regents of the University of California, Nerfstudio Team and contributors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import dataclasses
+from collections import OrderedDict
+from typing import TYPE_CHECKING
+
+import tyro
+
+from nerfstudio.configs.dataparser_configs import dataparser_configs
+from nerfstudio.configs.external_methods import get_external_methods
+from nerfstudio.configs.method_configs import descriptions as method_descriptions
+from nerfstudio.configs.method_configs import method_configs
+from nerfstudio.data.dataparsers.base_dataparser import DataParserConfig
+from nerfstudio.engine.trainer import TrainerConfig
+from nerfstudio.plugins.registry import discover_dataparsers, discover_methods
+
+
+def merge_methods(methods, method_descriptions, new_methods, new_descriptions, overwrite=True):
+    """Merge new methods and descriptions into existing methods and descriptions.
+    Args:
+        methods: Existing methods.
+        method_descriptions: Existing descriptions.
+        new_methods: New methods to merge in.
+        new_descriptions: New descriptions to merge in.
+    Returns:
+        Merged methods and descriptions.
+    """
+    methods = OrderedDict(**methods)
+    method_descriptions = OrderedDict(**method_descriptions)
+    for k, v in new_methods.items():
+        if overwrite or k not in methods:
+            methods[k] = v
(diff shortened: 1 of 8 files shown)


nerfstudio/configs/annotated_types.py
```diff
@@ -0,0 +1,139 @@
...
+        all_dataparsers,
+        prefix_names=False,  # Omit prefixes in subcommands themselves.
+    )
+
+AnnotatedDataParserUnion = tyro.conf.OmitSubcommandPrefixes[DataParserUnion]  # Omit prefixes of flags in subcommands.
+"""Union over possible dataparser types, annotated with metadata for tyro. This is
+the same as the vanilla union, but results in shorter subcommand names."""
+
+
+def fix_method_dataparser_type(method_config: TrainerConfig):
```
jkulhanek, line 77: @brentyi , can you please take a closer look at this function?

nerfstudio/configs/annotated_types.py
```diff
@@ -0,0 +1,159 @@
...
+from nerfstudio.configs.dataparser_configs import dataparser_configs
+from nerfstudio.configs.external_methods import get_external_methods
+from nerfstudio.configs.method_configs import descriptions as method_descriptions
+from nerfstudio.configs.method_configs import method_configs
+from nerfstudio.data.dataparsers.base_dataparser import DataParserConfig
+from nerfstudio.engine.trainer import TrainerConfig
+from nerfstudio.plugins.registry import discover_dataparsers, discover_methods
+
+
+def merge_methods(methods, method_descriptions, new_methods, new_descriptions, overwrite=True):
```
brentyi, line 30: while we're making changes, can we add annotations here?

nerfstudio/configs/annotated_types.py
```diff
@@ -0,0 +1,159 @@
...
+    Args:
+        methods: Existing methods.
+        method_descriptions: Existing descriptions.
+        new_methods: New methods to merge in.
+        new_descriptions: New descriptions to merge in.
+    Returns:
+        Merged methods and descriptions.
+    """
+    methods = OrderedDict(**methods)
+    method_descriptions = OrderedDict(**method_descriptions)
```
brentyi, line 41: Would `method_descriptions.copy()` work here? (since we require Python >=3.8 `OrderedDict` also seems unnecessary, unless it's just to make the intention clear)

nerfstudio/configs/annotated_types.py
brentyi: This makes sense but feels scary to me, is there no way to keep the parsers annotated as `AnnotatedDataParserUnion`? Is the issue with circular imports? Would breaking this file up into smaller files (like one where the dataparser union is defined and one where the method union is defined) help?

nerfstudio/configs/annotated_types.py
jkulhanek: Sure

nerfstudio/configs/annotated_types.py
jkulhanek: I think having the AnnotatedDataParserUnion in the DataManagerConfig is a bad design which leads to circular imports when you want to define the DataParser and something else (trainerconfig) in the same file. I agree that redefining the types here is not the safest. I think it would be preferable if there was a way to instruct tyro to do it instead: replace base types with the annotated unions? What do you think? If you like this solution I can try to implement it in tyro.

nerfstudio/configs/annotated_types.py
jkulhanek: I agree, no need to keep using OrderedDict since python>=3.8 is enforced.

nerfstudio/configs/annotated_types.py
brentyi: Sorry for missing your reply, I had accidentally left my comment twice and assumed you hadn't replied yet because I was looking at the other one...

> I think it would be preferable if there was a way to instruct tyro to do it instead: replace base types with the annotated unions

Yeah, this makes sense. There's an unfinished PR that would enable this here: https://github.com/brentyi/tyro/pull/30

If you're willing to make a PR to tyro making this kind of thing possible — either building on my PR or starting from scratch — that'd of course be awesome, but I don't have a good sense of how long something like that would take (or how long getting ramped up on the tyro codebase would take).

If it's too much effort just splitting things like the DataParser + trainerconfig that you described into separate files seems like an unideal but workable short-term solution.

Issue #1660, opened 2023-03-24
Connecting the viewer when using Docker


Alpe6825:
I started a docker container with the following command:

`docker run --gpus all -v /home/alex/nerfstudio_workspace:/workspace/ -v /home/alex/.cache/:/home/user/.cache/ -p 7007:7007 --rm -it --shm-size=12gb dromni/nerfstudio`

I trained my first nerf with poster data and use `ns-viewer --load-config outputs/poster/nerfacto/2023-03-24_162756/config.yml`
But visiting https://viewer.nerf.studio/?websocket_url=ws://localhost:7007 shows that renderer is disconnected. During training it was disconneceted too. 

What I have to do in order to connect it?

Liumt1949:
I have the same problem,do u solve it?

Ilyabasharov:
@Liumt1949 @Alpe6825 use 
```bash
ns-train nerfacto --data data/nerfstudio/poster --vis viewer_beta
```

PR #2157, opened 2023-07-02
Feature/run_beta_viewer

Ilyabasharov:
Done:
- add `--vis` option with two cases: `viewer` and `viewer_beta` for `ns-viewer` script
Example:
```bash
ns-viewer --load-config outputs/poster/nerfacto/2023-07-02_233032/config.yml --vis viewer_beta
```
Reason: #1660


tancik (approved):
LGTM, thanks for adding

Ilyabasharov:
Also resolved #2162 by fixing data format for viewer. Default format was torch.float64 after converting from numpy

tancik (approved):

OrangeSodahub:
@tancik Hi, what's the difference between the `viewer` and `viewer_beta`?? Which one should I use?

PR #2161, opened 2023-07-03
fix mesh export and add new sampling function for random cam datamanager

terrancewang:


terrancewang:
fixes https://github.com/nerfstudio-project/nerfstudio/issues/212. 

looks like there is a separate problem in our model implementation that results in a sphere of density surrounding the object. 
pineapplemeshunclipped

for now using the export bounding box helps with this, but I'm planning on taking a look at why this is happening also. 


pineapplemesh

tancik (approved):
LGTM

PR #2165, opened 2023-07-03
Implement nerfacto alpha transparency training

### LINKED ISSUE #1498: Training with masks results in artefacts outside the mask regions

nepfaff:
This PR cleans up and replaces https://github.com/nerfstudio-project/nerfstudio/pull/2025.

### Test data

Mustard data set with alpha masks.
https://drive.google.com/file/d/1XX4ioj9NgaRoMIA00Negp5x9XM8gWxjD/view?usp=sharing

```
ns-train nerfacto --pipeline.model.background_color "random" --pipeline.model.disable-scene-contraction True  instant-ngp-data --data data/mustard_alpha_images/
```

### Results

Accumulation without alpha transparent training:
image

Accumulation with alpha transparent training:
image

Closes #1498
Closes #2025


nepfaff:
@SamDM, @jkulhanek, @f-dy let me know if you have any comments :)

I tried to address all mentioned suggestions in https://github.com/nerfstudio-project/nerfstudio/pull/2025 while keeping the functionality the same.

nerfstudio/model_components/renderers.py
```diff
@@ -101,26 +101,64 @@ def combine_rgb(
...
+        assert isinstance(background_color, Tensor)
+        return background_color
 
-        assert isinstance(background_color, torch.Tensor)
-        comp_rgb = comp_rgb + background_color.to(weights.device) * (1.0 - accumulated_weight)
+    @classmethod
+    def blend_background(
+        cls,
+        image: Tensor,
+        outputs: Dict[str, Union[Tensor, list]],
```
jkulhanek, line 135: Can you please pass the optional background_color directly instead of passing the outputs?

jkulhanek: Also, can you please pass RGB image and "opacity" as two separate arguments?

nerfstudio/data/datasets/base_dataset.py
```diff
@@ -85,10 +85,10 @@ def get_image(self, image_idx: int) -> Float[Tensor, "image_height image_width n
         """
         image = torch.from_numpy(self.get_numpy_image(image_idx).astype("float32") / 255.0)
         if self._dataparser_outputs.alpha_color is not None and image.shape[-1] == 4:
-            assert image.shape[-1] == 4
             image = image[:, :, :3] * image[:, :, -1:] + self._dataparser_outputs.alpha_color * (1.0 - image[:, :, -1:])
-        else:
-            image = image[:, :, :3]
+        elif image.shape[-1] == 3:
```
jkulhanek, line 89: Can you avoid doing this? It will add an overhead for cases where the opacity is not used (most cases). Instead, I would suggest keeping the original number of channels (3 or 4) if alpha_color is not set.

nerfstudio/models/base_surface_model.py
```diff
@@ -292,7 +292,7 @@ def get_loss_dict(self, outputs, batch, metrics_dict=None) -> Dict[str, torch.Te
             metrics_dict: dictionary of metrics, some of which we can use for loss
         """
         loss_dict = {}
-        image = batch["image"].to(self.device)
+        image = batch["image"][..., :3].to(self.device)
```
jkulhanek, line 295: Shouldn't this also call blend_background?

nerfstudio/models/nerfacto.py
```diff
@@ -283,7 +285,14 @@ def get_outputs(self, ray_bundle: RayBundle):
         weights_list.append(weights)
         ray_samples_list.append(ray_samples)
 
-        rgb = self.renderer_rgb(rgb=field_outputs[FieldHeadNames.RGB], weights=weights)
+        background_color = (
+            self.renderer_rgb.get_background_color(rgb=field_outputs[FieldHeadNames.RGB], background_color="random")
```
jkulhanek, line 289: background_color="random" is not correct here, is it? Isn't it part of the config?

nerfstudio/models/nerfacto.py
```diff
@@ -292,6 +301,8 @@ def get_outputs(self, ray_bundle: RayBundle):
             "accumulation": accumulation,
             "depth": depth,
         }
+        if self.config.use_alpha_transparency_training:
```
jkulhanek, line 304: I would suggest making this default and removing the flag from the config. The transparency training will only be used if background_color is "random"

nerfstudio/models/nerfacto.py
```diff
@@ -106,6 +106,8 @@ class NerfactoModelConfig(ModelConfig):
     """Orientation loss multiplier on computed normals."""
     pred_normal_loss_mult: float = 0.001
     """Predicted normal loss multiplier."""
+    use_alpha_transparency_training: bool = False
```
jkulhanek, line 109: Please remove this as it is not needed.

nerfstudio/models/nerfacto.py
```diff
@@ -321,15 +332,15 @@ def get_outputs(self, ray_bundle: RayBundle):
 
     def get_metrics_dict(self, outputs, batch):
         metrics_dict = {}
-        image = batch["image"].to(self.device)
+        image = batch["image"][..., :3].to(self.device)
```
jkulhanek, line 335: If blending is not used here as well, outputs["rgb"] as generated using the random noise background. This would decrease the metrics. Solution? Do not blend neither the RGB nor the target image, but do it inside of loss/metric computation?

nerfstudio/models/base_surface_model.py
nepfaff: This would just add an additional function call to every model that does not use alpha transparent training. Is there a reason why you would prefer using `blend_background` here? I think it would only be more confusing if we call `blend_background(rgb=batch["image"][..., :3].to(self.device), opacity=None, background_color=None)` instead of `batch["image"][..., :3].to(self.device)`. I passed the optional arguments for clarity.

nerfstudio/models/nerfacto.py
SamDM: The current implementation of the `random` setting picks a random color per ray (not matched to the target). 

Because the target pixel does match the random color, the only way to minimize the loss during training is to have an accumulation of 1 along every training ray. Since it is impossible for the radiance field to match the constantly changing randomly chosen color.

I don't know if this has any advantage over the default setting 'last sample', which picks the color generated by the radiance field the furthest along the ray as background color (AFAIK). Maybe the current `random` setting helps in pushing the radiance field to have a more pronounced background model for unbounded scenes. But some small scale testing I did does not confirm that `random` improved results in any way.

nerfstudio/models/nerfacto.py
nepfaff: This wouldn't fix the metric problem as the rendered image still won't contain the original background as this wasn't used during training. For binary alpha masks, it could make sense only to compute the metrics using the pixels with alpha=1. This wouldn't make sense for non-binary alpha masks; however, I don't see a use case for having these.
What do you think about this?

nerfstudio/models/nerfacto.py
nepfaff: That makes sense. Thanks!

nepfaff: Do you guys think it is acceptable to replace that feature with alpha transparent training if an alpha channel is present? In that case, we could get rid of the `use_alpha_transparency_training` argument and make it the default. It might be annoying for people that have images with alpha channels but want to train with the original feature.

nerfstudio/data/datasets/base_dataset.py
nepfaff: done

nerfstudio/model_components/renderers.py
nepfaff: Passed them as separate arguments. Also, it is probably better to not allow for optional arguments but only call this function if we actually want to blend the background. We can resolve this if you agree with removing the optionality of the arguments.

nerfstudio/models/nerfacto.py
nepfaff: I agree if we decide to remove the `use_alpha_transparency_training` argument. Otherwise, it would make more sense to keep it hard coded.

nerfstudio/models/base_surface_model.py
jkulhanek: What I meant was: shouldn't this be implemented for all methods - not just nerfacto?

nepfaff: That would be an option. I didn't do it based on the suggestion [here](https://github.com/nerfstudio-project/nerfstudio/pull/2025#issuecomment-1601654847). But it wouldn't take much time to add it to the other currently existing models once we agree on the `nerfacto` implementation :)

nerfstudio/models/nerfacto.py
jkulhanek: I would remove the use_lapha_transparency_training and make it default. If `background_color=random` is used the target blended background colour should indeed match the randomly generated ray colour. Please correct me if wrong, but this will be compatible with previous behaviour when no transparency in images, but when images are transparent this will achieve the intended behaviour.

nepfaff: All correct :) I will change it

nerfstudio/models/nerfacto.py
jkulhanek: Why wouldn't it fix the problem? Both the prediction and the target would be RGBA images and for metrics computation, a default background colour would be used (e.g. black). Note, that the alpha channel for the prediction would be the accumulation. For loss computation, the matching random colour would be used. However, having the random colour for metrics computation would decrease LPIPS. This will also enable access to the original RGBA images should the user want to store them. What do you think of this proposal?

nepfaff: Just to make sure that I understand, are you suggesting the following?
1. Use accumulation to mask the predicted images
2. Use alpha to mask the input images
3. Insert black background outside the masks
4. Compute the metrics based on these masked images

nerfstudio/models/base_surface_model.py
jkulhanek: I think it would be really cool to have this implemented for all methods.

nerfstudio/models/nerfacto.py
jkulhanek: I suggest the following:
1. predict RGBA images where alpha is the accumulation
2. for loss computation, perform blending with random background (if background_color=random) and compute the loss afterwards
3. for metric computation, if background_color==random perform blending with a non-random (black or configurable) colour. Then the metrics can be computed on the resulting RGB images.

I do not claim this to be the best approach. It is just a suggestion. What are your thoughts?

nerfstudio/model_components/renderers.py
jkulhanek: But shouldn't blending happen always? When background_color=random it would use the provided background_color, otherwise it would use the renderer's background colour? Perhaps I am missing something...

nerfstudio/models/nerfacto.py
nepfaff: Thanks for clarifying! I understand the suggestion a lot better now. 

I think this would be a reasonable approach. I'd say another viable option would be to use the original alpha masks to mask the images before computing the metrics. The first option includes the predicted alpha channel in the metric, while the second one does not. I'm not sure which one is more desirable as I have never used reconstruction metrics before. The second one would not require a configurable parameter. Do you have any preferences/ insights here?

jkulhanek: I would advise against using masks in metrics. 1) the alpha channel may not be binary 2) would be difficult to do for lpips and others

nerfstudio/model_components/renderers.py
nepfaff: It should only happen if the input data includes an alpha channel. If this is the case, we would also have a `background_color` to pass to this method.
So you are right, `if "background_color" in outputs` is not needed.

nepfaff: Actually, I just wanted to change this and realized that using the default `background_color` is non-trivial without saving extra data somewhere. For example, at the moment, we cannot get the RGB color of "last_sample" at this point in the code.

Do you see a specific use case for this functionality (alpha channel for blending without alpha transparent training)? If not, I'd prefer to leave it to keep the code cleaner.

nerfstudio/models/nerfacto.py
nepfaff: Okay, these are good points. I will implement your suggestion then.

nepfaff: Implemented this now. I added another method to the RGB renderer to avoid code duplication, but I'm not a massive fan of it. Let me know if you have suggestions for making this cleaner.

It does significantly improve the metrics and makes them meaningful.

Without the change:
psnr_before

After the change:
psnr_after

nerfstudio/model_components/renderers.py
jkulhanek: Ok, let's keep it cleaner. The reason why I suggested this is that if we always load the alpha channel and background_color!=random, there would still be the alpha channel in the target image, no?

nerfstudio/models/nerfacto.py
jkulhanek: Looks really good! Nice work!

nerfstudio/models/nerfacto.py
```diff
@@ -283,7 +283,16 @@ def get_outputs(self, ray_bundle: RayBundle):
         weights_list.append(weights)
         ray_samples_list.append(ray_samples)
 
-        rgb = self.renderer_rgb(rgb=field_outputs[FieldHeadNames.RGB], weights=weights)
+        background_color = (
+            self.renderer_rgb.get_background_color(
+                rgb=field_outputs[FieldHeadNames.RGB], background_color=self.config.background_color
+            )
+            if self.training
```
jkulhanek, line 290: Doesn't this cause the training metrics to be still computed incorrectly? Would it be possible to defer adding the background colour to the loss computation? (Outputting RGBA instead of RGB)

SamDM: Just my two cents here: 

### For computing gradients:
Here you have to blend with the randomly chosen background color, this is the core principle of this 'carving' principle.

### For computing metrics:

Comparing RGBAs rendered by the model to the input image target RGBAs directly is not a good idea for computing metrics. If the A channel in the target image pixel is low, the RGB values don't matter, and also can't be learned by the model (since it has never seen them during training, it saw target colors blended with the randomly changing background color instead). So you'll get non-justified poor metrics for pixels with low A.

What you could do to counter this is to multiply the metric afterwards with the alpha value per pixel (alpha scaled between 0.0 and 1.0). So in short:
1. Get rendered RGBA and target pixel RGBA, where the A for the rendered pixel is accumulation. No background blending.
2. Compute metric using both RGBA values. (this will yield unjust poor metrics on pixels with low alpha).
3. Multiply the metric from step (2.) with GT alpha value (scaled between 0.0 and 1.0).

But I'm not sure how compatible this is with all metrics. So I'd still just prefer choosing a fixed (i.e. black) but configurable background color for computing metrics.

nerfstudio/model_components/renderers.py
nepfaff: Yes, but in this case, we would want to discard the alpha channel as we don't currently use it for anything other than alpha-transparent training. At least I don't see the benefit of always blending the background color if an alpha channel is present.

nerfstudio/models/nerfacto.py
nepfaff: You are right. The training metrics won't be 100% correct, and the evaluation metrics will be wrong.
Should we reintroduce the `pipeline.model.use-alpha-transparency-training` flag to determine whether to blend at this point (keeping the existing behavior) or later on (alpha-transparent training)?

nepfaff: Thinking about this more carefully, the metrics will be correct once the predicted alpha matches the training alpha (desired behavior for alpha transparent training). So the metrics will still increase as expected and converge toward the correct metrics. So maybe the current implementation isn't bad, especially as computing metrics with uniform color background masking is an approximation anyway?

nerfstudio/model_components/renderers.py
machenmusik: I remember looking at this a good while ago and thinking that a scoreboard of transient values, like this run's background color selection, could be more generally useful.

The other use case I had in my head was some way to prevent single bad iterations from having a catastrophic impact, maybe pause/stop training if metrics plummet instead of saving checkpoint.

nepfaff: What are you thinking in terms of scoreboard? Do you mean having a dedicated object/place to write values like `background_color` to instead of storing them in `outputs`? 

>The other use case I had in my head was some way to prevent single bad iterations from having a catastrophic impact, maybe pause/stop training if metrics plummet instead of saving checkpoint.

I'm sorry, but I'm not quite following how this relates to always blending the background color if an alpha channel is present. Or do you mean as a use case for the scoreboard idea?

machenmusik: > Do you mean having a dedicated object/place to write values like background_color to instead of storing them in outputs?

That was the thought, although not sure whether there is need.

> Or do you mean as a use case for the scoreboard idea?

Yes, this (sorry)

nepfaff: This seems like something that would warrant a separate pull request that then also implements one of your suggested functionalities.
However, I do see one reason for implementing something like this as part of this PR: All the outputs are currently passed through [this resizing operation](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/models/base_model.py#L187). This doesn't make much sense for `background_color`. It does not crash as the background color is a per-image pixel color, but it still isn't great.

Any opinions on this or separate PR?

machenmusik: Separate PR probably - unless it would fit neatly into this one

machenmusik:
Do we think this is in condition to be merged?

yonghoonkwon:
The results come out well when learning with the background data removed by the this pull, but the function for mesh export (e.g., point cloud, mesh) does not seem to work. Exporting the point cloud takes a very long time and does not proceed.

yonghoonkwon:
If set the background to random, rendering works fine, but there seems to be a problem with exporting mesh.
I started learning with the following command and the function for exporting point cloud or mesh (poisson) is not working properly.

`CUDA_VISIBLE_DEVICES=0 ns-train nerfacto --data $DATA_SET --max-num-iterations 300001 --vis viewer+tensorboard --experiment-name $EXP_NAME --pipeline.model.background_color random --pipeline.datamanager.camera-optimizer.mode off --pipeline.model.near-plane 0.5 --pipeline.model.far-plane 10.0 --machine.num-devices 0 --pipeline.datamanager.camera-res-scale-factor 0.5 --pipeline.model.disable-scene-contraction True nerfstudio-data --train-split-fraction 1`

nerfstudio/models/nerfacto.py
jkulhanek: I don't think computing metrics on RGBA images is a viable idea. What I suggested was the following:
1. the get_model_outputs returns RGBA when background colour is random and RGB otherwise
2. if background color is random -> compute_metrics does blending with black colour before the metrics are computed, and loss computation does blending with random colours.
3.

nepfaff:
I didn't get amazing results with Poisson on that dataset. This isn't alpha transparent training dependent though...

TSDF fusion works great and is significantly improved by alpha transparent training. This is actually my sole use case of this feature. Currently on the way to an airport, but will post some mesh export results once I'm there :)

nerfstudio/model_components/renderers.py
```diff
@@ -101,26 +101,96 @@ def combine_rgb(
...
+            predicted_rgb: The predicted RGB values.
+            predicted_opacity: The predicted opacity/ accumulation.
+            background_color: The background color to use for blending. This should be
+                a single color to use for all pixels.
+
+        Returns:
+            A tuple of the predicted and ground truth RGB values.
+        """
+        if gt_image.shape[-1] == 4 and self.background_color == "random":
+            predicted_rgb = self.blend_background(
```
jkulhanek, line 176: This is not correct, right? For predicted RGB the background colour should be just "added" not "blended". This would give different (incorrect results).

jkulhanek: Can we have two methods? "blend_background" and "add_background"? "blend_background" would be called for the GT images and 'add_background" for predictions?

jkulhanek (changes requested):

jkulhanek:
@nepfaff , Can you please add me as a contributor to your fork so I can try implementing some changes? If you are not comfortable with this idea, I can alternatively create a new branch...

yonghoonkwon:
> What does it mean the function is not working properly? Can you please post the error message or something?

If you export pointcloud to a learned model, it will not proceed at 0%, but you will continue to consume hardware resources (GPU). I don't know the exact cause, but I think it's because of the random value in the background. If `pipeline.model.background_color `is black, ghost pixels are included, but the export was successful. Point cloud needs to be extracted for Poisson mesh extraction, but I think there is a problem in this process.

nepfaff:
I did manage to run Poisson but I can try with point cloud. Will report the results soon

nepfaff:
> @nepfaff , Can you please add me as a contributor to your fork so I can try implementing some changes? If you are not comfortable with this idea, I can alternatively create a new branch...

Done :)

yonghoonkwon:
> I didn't get amazing results with Poisson on that dataset. This isn't alpha transparent training dependent though...
> 
> TSDF fusion works great and is significantly improved by alpha transparent training. This is actually my sole use case of this feature. Currently on the way to an airport, but will post some mesh export results once I'm there :)

What I tested is my custom data, but I will learn and test again with mustard data and share the results.

yonghoonkwon:
> I did manage to run Poisson but I can try with point cloud. Will report the results soon

Can you share the script you used to train mustard data?? I thought testing under the same conditions would report the exact results

nepfaff:
> > I did manage to run Poisson but I can try with point cloud. Will report the results soon
> 
> 
> 
> Can you share the script you used to train mustard data?? 

Does the command in the description not work?

nerfstudio/models/nerfacto.py
nepfaff: Isn't this what is happening [here](https://github.com/nerfstudio-project/nerfstudio/pull/2165#discussion_r1252683799) now? `get_model_outputs` already returns accumulation (alpha channel) so no need to have an if/else logic here.

yonghoonkwon:
> > > I did manage to run Poisson but I can try with point cloud. Will report the results soon
> > 
> > 
> > Can you share the script you used to train mustard data??
> 
> Does the command in the description not work?

No, I wanted to check the experimental script because the results of my experiment included strange textures. I share the results of my test in the [Dropbox](https://www.dropbox.com/sh/pc45xgrqtp9dk4x/AAC2AzDGCJ9Pk8uMXmf5iueqa?dl=0). PCL works well on mustard datasets, but there seems to be a problem with rendering because the TSDF mesh texture contains random values. Let me know if I did something wrong :)

nepfaff:
> > > > I did manage to run Poisson but I can try with point cloud. Will report the results soon
> > > 
> > > 
> > > Can you share the script you used to train mustard data??
> > 
> > 
> > Does the command in the description not work?
> 
> No, I wanted to check the experimental script because the results of my experiment included strange textures. I share the results of my test in the [Dropbox](https://www.dropbox.com/sh/pc45xgrqtp9dk4x/AAC2AzDGCJ9Pk8uMXmf5iueqa?dl=0). PCL works well on mustard datasets, but there seems to be a problem with rendering because the TSDF mesh texture contains random values. Let me know if I did something wrong :)

I don't quite understand, sorry! What do you mean with `experimental script`? Do you mean the checkpoints that I got from training?

yonghoonkwon:
> > > > > I did manage to run Poisson but I can try with point cloud. Will report the results soon
> > > > 
> > > > 
> > > > Can you share the script you used to train mustard data??
> > > 
> > > 
> > > Does the command in the description not work?
> > 
> > 
> > No, I wanted to check the experimental script because the results of my experiment included strange textures. I share the results of my test in the [Dropbox](https://www.dropbox.com/sh/pc45xgrqtp9dk4x/AAC2AzDGCJ9Pk8uMXmf5iueqa?dl=0). PCL works well on mustard datasets, but there seems to be a problem with rendering because the TSDF mesh texture contains random values. Let me know if I did something wrong :)
> 
> I don't quite understand, sorry! What do you mean with `experimental script`? Do you mean the checkpoints that I got from training?

I meant a script for training the data. I'm sorry made you confused!
`ns-train nerfacto --pipeline.model.background_color "random" --pipeline.model.disable-scene-contraction True  instant-ngp-data --data data/mustard_alpha_images/`

nepfaff:
> > > > > > I did manage to run Poisson but I can try with point cloud. Will report the results soon
> > > > > 
> > > > > 
> > > > > Can you share the script you used to train mustard data??
> > > > 
> > > > 
> > > > Does the command in the description not work?
> > > 
> > > 
> > > No, I wanted to check the experimental script because the results of my experiment included strange textures. I share the results of my test in the [Dropbox](https://www.dropbox.com/sh/pc45xgrqtp9dk4x/AAC2AzDGCJ9Pk8uMXmf5iueqa?dl=0). PCL works well on mustard datasets, but there seems to be a problem with rendering because the TSDF mesh texture contains random values. Let me know if I did something wrong :)
> > 
> > 
> > I don't quite understand, sorry! What do you mean with `experimental script`? Do you mean the checkpoints that I got from training?
> 
> I meant a script for training the data. I'm sorry made you confused! `ns-train nerfacto --pipeline.model.background_color "random" --pipeline.model.disable-scene-contraction True instant-ngp-data --data data/mustard_alpha_images/`

Sorry, I'm still not quite following. The script is nerfstudio's `ns-train`, called with the command you just posted:
`ns-train nerfacto --pipeline.model.background_color "random" --pipeline.model.disable-scene-contraction True  instant-ngp-data --data data/mustard_alpha_images/`. The data `mustard_alpha_images` should come from the link in the description

nepfaff:
Did some export testing. For me, the PCD export was almost instant.

## PCD results

These are horrible in my opinion.
synthetic_pcd00

## TSDF results

These vary a bit per training due to randomness but the variations are acceptably small. I think this one is pretty great.
additional_cameras_alpha_transparent_mustard_synethic_orig_cameras

The nice thing with this synthetic data is that using additional cameras for TSDF further improves these results (not currently implemented in Nerfstudio). The result below is with a camera sphere of around 600 cameras (training data is only around 200).
additional_cameras_alpha_transparent_mustard_synethic_additional_cameras

nepfaff:
I used reduced but still generous bounds for the mustard export. Keeping the original massive bounding box does increase computation and might slow down exports. Maybe this was your issue?
I used the following bounds: `--use-bounding-box True --bounding-box-min -0.2 -0.3 -0.5 --bounding-box-max 0.2 0.3 0.5`. TSDF export takes around 10-15min on an A5000 Laptop GPU, when using the above bounds.

yonghoonkwon:
It's different from my results. Like the results I shared on Dropbox, the texture is extracted with random values. I hope this was just my problem. I'll test again tomorrow.

nepfaff:
Sorry, I didn't look at your results. Now I did. It does look very familiar to results I got with a previous version. In particular, these are the expected results when not using `if self.training` for getting the background color in `nerfacto`. Let me know if this code reference is confusing :)

If you are using the newest version, then this does look slightly strange. In this case, could you send me your exact command for exporting? And maybe even the checkpoints.

nepfaff:
Nothing about extra options. All in the code 😃. 
I will look into your checkpoints tomorrow (Unfortunately run put of laptop battery)

machenmusik:
w.r.t. non-stellar mustard point cloud export - what are we comparing with?  Pre-carving, mustard export had its own challenges with export unless one made judicious use of cropping, IIRC...

nepfaff:
I agree. The PCD export still seems much better than without alpha-transparent training. Especially as I no longer spent ages finding the smallest export bounding box. My wording was probably a bit too pessimistic 😅

nerfstudio/model_components/renderers.py
jkulhanek: Accumulation is not binary, right?

jkulhanek:
I have implemented some changes here: https://github.com/nepfaff/nerfstudio/compare/alpha_carving_v2...nerfstudio-project:nerfstudio:nepfaff/alpha_carving_v2

@nepfaff  let me know what do you think, please. I still don't like the interface of RGBRenderer though...

yonghoonkwon:
@jkulhanek I left a comment on one part.

nepfaff:
> I have implemented some changes here: [nepfaff/nerfstudio@alpha_carving_v2...nerfstudio-project:nerfstudio:nepfaff/alpha_carving_v2](https://github.com/nepfaff/nerfstudio/compare/alpha_carving_v2...nerfstudio-project:nerfstudio:nepfaff/alpha_carving_v2)
> 
> @nepfaff let me know what do you think, please. I still don't like the interface of RGBRenderer though...

Thanks @jkulhanek! I have mixed feelings about these changes tbh. I see why some of these changes can be beneficial, but I also feel like the API got unnecessarily complex. However, most of this is probably personal style preferences and I'd prefer getting this merged sooner rather than debating about style. So I'd be happy to combine these changes as is to move forward. What do you think?
Nevertheless, wanted to add some comments about stuff that stuck out to me on the first pass. Hopefully they are more helpful than annoying :) 

I can't make comments on that comparison, so some stuff here:
- `add_background_color`: mismatch of argument names and docstring. I think `opacity` is a better name as it is more general. Also, the line `blended_rgb = rgb` could be removed. Probably fine to return the value directly as the function names says it all already.
- `get_rgb_image_pair`: The name of this function confuses me a bit and doesn't seem to go with the docstring. I'm not a fan of this function doing a lot of different things, and seems to mix high-level and low-level logic. I personally preferred some of the higher-level logic staying in the model and the lower-level logic staying in RGBRenderer, even if this leads to some code duplication. Maybe I'm seeing this wrong in the comparison view but isn't `back_color` never used and `elif self.background_color == "random"` can't happen? Also, the docstring seems incomplete.

nepfaff:
@yonghoonkwon 

I looked at the PCD you shared, and I think it looks even better than mine. It does have some random color points, which mine does not have, though. I'd be very happy to use this in an ICP algorithm, etc.
pcd00

For TSDF export, I run your checkpoints, and I'm getting the same random color artifacts...
Not sure what happened there and why there is a mismatch with my results. Let me train it once more and see what happens.

yonghoonkwon:
@nepfaff I tried exporting again using the trained checkpoint, and got improved results (similar to the you shared). I just resized the extraction box to be tighter to the object.

jkulhanek:
@nepfaff yes! I have mixed feelings about my proposal as well! Let’s not accept anything which would pollute the api.

nepfaff:
As a note: I will be traveling without my laptop for around a month from Wednesday and won't be online during that time

nerfstudio/model_components/renderers.py
nepfaff: > Accumulation is not binary, right?

It should be once it converges to the desired alpha. So I don't think it matters for metric computation. Made a comment about this [here](https://github.com/nerfstudio-project/nerfstudio/pull/2165#discussion_r1254305734).

nepfaff: > Also, prediction blending should happen regardless of whether the gt has transparency. In fact, non-transparent bg recovers the previous behaviour.

For your changes, this is correct. The changes currently on this PR still due the previous blending before this function is called.

nepfaff:
> @nepfaff I tried exporting again using the trained checkpoint, and got improved results (similar to the you shared). I just resized the extraction box to be tighter to the object.

Which bounds worked for you?

jkulhanek:
> As a note: I will be traveling without my laptop for around a month from Wednesday and won't be online during that time

It would be cool to merge this before Wednesday.

yonghoonkwon:
> > @nepfaff I tried exporting again using the trained checkpoint, and got improved results (similar to the you shared). I just resized the extraction box to be tighter to the object.
> 
> Which bounds worked for you?

@nepfaff  I tested with `--use-bounding-box True --bounding-box-min -0.2 -0.3 -0.5 --bounding-box-max 0.2 0.3 0.5`. 

@jkulhanek Are you check of the PyTorch error i mentioned? (https://github.com/nepfaff/nerfstudio/commit/8b7e6870c46d87017d96dc408e40e682c4af7ec9#r121605677). Other parts seems to well for the mustard dataset.

nepfaff:
@yonghoonkwon I did a more thorough investigation/ comparison of mesh export quality. I'm getting the best quality with the first commit (0bbaba6f1aa494499711dd1e027a71b0d540e60a). I can't instantly tell why this is, but it is significant because this commit has no random colors showing up, whereas the others do. It is hard to present definite results with multiple trainings per commit due to training times. Let's put recovering this mesh quality as a blocker for merging this

nerfstudio/models/nerfacto.py
```diff
@@ -283,7 +283,16 @@ def get_outputs(self, ray_bundle: RayBundle):
...
+            else None
```
nepfaff, line 291: Using `else "last_sample"` instead of `else None` fixes the mesh export quality. I can't think of a neat fix apart from re-introducing the `use_alpha_transparency_training` flag. Any ideas?

nepfaff: With the change:
with_change01

Without change:
without_change02

jkulhanek:
I think I have found a nice (simpler) solution to:
1. return the RGB colour as if blended if black such that image generation and metrics computation works fine.
2. perform the random-color blending in the loss computation

Let me know what you think. The changes are here: https://github.com/nepfaff/nerfstudio/pull/1

jkulhanek:
Please follow the conversation here: https://github.com/nepfaff/nerfstudio/pull/1

jkulhanek:
Ok, just to summarize here.  Currently, the transparency-based training is fully implemented for all methods and metrics should be computed correctly. The last piece of code needed is to fix the export utils to correctly handle transparency.

jkulhanek:
@nepfaff  Can you please check if the updated export utils produce the desired result?

nerfstudio/exporter/exporter_utils.py
```diff
@@ -155,6 +155,13 @@ def generate_point_cloud(
                 normal = (normal * 2.0) - 1.0
             point = ray_bundle.origins + ray_bundle.directions * depth
 
+            # Filter points with opacity lower than 0.5
```
nepfaff, line 163: It would be nice for this to be configurable

nerfstudio/exporter/exporter_utils.py
```diff
@@ -160,6 +160,13 @@ def generate_point_cloud(
             point = ray_bundle.origins + ray_bundle.directions * depth
             view_direction = ray_bundle.directions
 
+            # Filter points with opacity lower than 0.5
+            mask = rgba[..., -1:] > 0.5
+            point = point[mask]
```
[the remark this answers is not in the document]
nepfaff: Went ahead and fixed this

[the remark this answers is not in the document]
jkulhanek: Great!

nerfstudio/exporter/exporter_utils.py
jkulhanek: Do you think it makes a lot of difference? I assume most opacity will be binary, with only a tiny fraction getting close to 0.5.

nerfstudio/models/base_model.py
```diff
@@ -187,6 +187,32 @@ def get_outputs_for_camera_ray_bundle(self, camera_ray_bundle: RayBundle) -> Dic
...
+        """
+        accumulation_name = output_name.replace("rgb", "accumulation")
+        if (
+            not hasattr(self, "renderer_rgb")
+            or not hasattr(self.renderer_rgb, "background_color")
+            or accumulation_name not in outputs
+        ):
+            raise NotImplementedError(f"get_rgba_image is not implemented for model {self.__class__.__name__}")
+        if self.renderer_rgb.background_color == "random":
+            rgb = outputs[output_name]
```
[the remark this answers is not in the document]
jkulhanek: Will you do it or should I?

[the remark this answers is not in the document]
nepfaff: I can do it :)

nerfstudio/exporter/exporter_utils.py
nepfaff: This is an example histogram:
```
(array([28222,    51,    40,    26,    29,    27,    30,    46,    69,
        4228]), array([0.        , 0.10000004, 0.20000008, 0.3000001 , 0.40000015,
       0.5000002 , 0.6000002 , 0.7000002 , 0.8000003 , 0.90000033,
       1.0000004 ], dtype=float32))
```

opacity is indeed mainly binary. Let's keep the hardcoded threshold :)

jkulhanek: Ok

jkulhanek:
Great! @nepfaff, thank you very much for your work. This has been a pleasure.

jkulhanek (approved):

machenmusik:
Nice!  Just to confirm, this mode is automatic when input images contain alpha, but separate mask images will have previous non-carving behavior - is that right?  Can both be used simultaneously?

nepfaff:
Almost. It is only automatic if the images have an alpha channel and the background color is random (this is not true by default). Hence, alpha images can be used with the previous behavior if the color is not random.

nepfaff:
> Thanks for working on this! Just wondering if this should work with equirectangular images with an alpha channel? The old workflow with separate mask images did not work with equirectangular for some reason. 
> 
> 
> 
> If so, do I just need to add `--pipeline.model.background_color "random"` to my training command?
> 
> 
> 
> Also, do masks in the alpha channel use less VRAM than having separate masks? I was never able to use those with large datasets as I would get CUDA out of memory errors.
> 
> 
> 
> 

It should work. Just add an alpha channel and specify the random background color as you suggested. 
I'd assume that they would use less VRAM but haven't tested this. It would be amazing if you could try this and post the results here :)

nepfaff:
> Hello, Trying to follow up this great work.  Is there any way to transform the image format( nerfstudios' original(.png) &camera position(transforms.jason) )
> 
>  to this nerfacto alpha transparency training model's format?

You need to add an alpha channel to your images. This is a 4th channel with values between 0 and 1. For alpha-transparent training you would use binar values where zero is transparent.
The easiest way to do this is probably numpy, concatenating the alpha channel with the RGB images. 

Then additionally specify a random background color as specified in the PR description.

hnj5247:
> Using   rembg p <input.img.folder> <output.img.folder> 
 I was able to remove the background of real world photo and add alpha. [r,g,b,a] . Thank you!
However it worked when the line '--pipeline.model.disable-scene-contraction True' was removed.
I think it's because my figure wasn't centered. Would my guess be right?

Also if I rather than using  'nerfacto' run 'instant-ngp' would this model use the alpha transparency value?
"ns-train instant-ngp --pipeline.model.background_color "random"  instant-ngp-data --data data/"

Thank you for replying!

f-dy:
@gradeeterna to mask yourself out, you shouldn't use these masks ("alpha transparency trauining" / "alpha carving"), but the "ignore" masks.
Here's how I add a single mask to all my fisheye images. You have to add the mask to the transforms.json, but also downscale it. You can easily customize it to add one mask per image. See also the discussion in https://github.com/nerfstudio-project/nerfstudio/issues/1498 for the "ignore" masking strategy
```python
cat > add_mask_to_transforms_json.py <<EOF
#!/usr/bin/env python
import sys
import json

if len(sys.argv) != 3:
    print(f"Usage: {sys.argv[0]} input_transforms.json output_transforms.json")
    sys.exit(1)
with open(sys.argv[1]) as input_file:
    file_contents = input_file.read()
parsed_json = json.loads(file_contents)
for frame in parsed_json["frames"]:
    frame["mask_path"] = "masks/mask.png"
with open(sys.argv[2], "w") as output_file:
    json.dump(parsed_json, output_file, indent=4)
EOF
```
```python
cat > downsize_mask.py <<EOF
#!/usr/bin/env python
import cv2
import sys
import json
from pathlib import Path

if len(sys.argv) != 2:
    print(f"Usage: {sys.argv[0]} path_to/mask.png")
    print(f"Output is path_to/masks_{downscale}/mask.png")
    sys.exit(1)
mask_path = Path(sys.argv[1])
mask = cv2.imread(str(mask_path), cv2.IMREAD_GRAYSCALE)
height, width = mask.shape[:2]
processed_data_dir = mask_path.parent
downscale_factors = [2, 4, 8]
for downscale in downscale_factors:
    mask_path_i = processed_data_dir / f"masks_{downscale}"
    mask_path_i.mkdir(exist_ok=True)
    mask_path_i = mask_path_i / "mask.png"
    mask_i = cv2.resize(
        mask, (width // downscale, height // downscale), interpolation=cv2.INTER_NEAREST
    )
    cv2.imwrite(str(mask_path_i), mask_i)
    print(f"Wrote {mask_path_i}")
EOF
```

gradeeterna:
@f-dy Hey, using a single ignore mask for all my fisheye images does work, but it slows down training by about 5x. When I use one mask per image with a big dataset, I get out of memory errors on my 3090 24GB. I was trying this method in case it was faster and uses less memory.

The way masks work with NGP is great, and doesn't seem to slow down training or increase memory usage.

Thanks for sharing those scripts, very useful!

machenmusik:
(IIRC there is an option that can be set which puts masks on gpu which significantly speeds up training.)

gradeeterna:
Oh yeah, just found it - `--pipeline.datamanager.masks-on-gpu True`

Training speed is almost the same as without masks, thanks a lot!

hoanhle:
@nepfaff can you re-enable the test dataset again https://drive.google.com/file/d/1XX4ioj9NgaRoMIA00Negp5x9XM8gWxjD/view?usp=sharing so I could debug why training nerfacto with alpha channel doesn't work? tnx a lot!

nepfaff:
@hoanhle I couldn't find it on my drive anymore. I will look for it on my disk and will get back to you in the next few days.

nepfaff:
I think that this should be the original dataset: https://drive.google.com/file/d/13u5PF7OQDLLat1MFMKWJgRMV5jwBuhjm/view?usp=sharing
@hoanhle let me know how it goes!

petertheprocess:
> I think that this should be the original dataset: https://drive.google.com/file/d/13u5PF7OQDLLat1MFMKWJgRMV5jwBuhjm/view?usp=sharing @hoanhle let me know how it goes!

Thanks a lot! Could u also share the `transform.json`?

nepfaff:
[transforms.json](https://github.com/user-attachments/files/16121989/transforms.json)
I found this file but haven't had a chance to test it yet. Let me know if it works. If not, then I can produce a new one by running COLMAP.

nepfaff:
@hoanhle 
It still works for me with my mustard data. The screenshot below is taken after a few seconds of training.
image

It also seems to work when enabling scene contraction:
image

The functionality is independent of the data loader. My data processing scripts were just originally written for instantNGP and I never bothered changing them. However, some data loaders might support slightly different features. I'm not sure about this though.

Did it work for you with my data? If yes, then it could make sense to debug by checking whether the data loader output is the same between the two datasets. If not, then there is probably an issue with your input data or the data loader you're using.

> Misc: how did you visualize the scene box?
I didn't. It just becomes visible when you train without alpha transparent training as the density accumulates throughout the entire scene box.

nepfaff:
For reference, this is what I'm getting after similar training time (few seconds) when I'm disabling alpha transparent training:
image

hoanhle:
@nepfaff if i understand correctly, shouldn't we use `--pipeline.model.background_color "black" ` instead of `--pipeline.model.background_color "random" ` here when using alpha channel? What am I missing?

```
    pred_rgb, image = self.renderer_rgb.blend_background_for_loss_computation(
        pred_image=outputs["rgb"],
        pred_accumulation=outputs["accumulation"],
        gt_image=image,
    )
    rgb_loss = self.rgb_loss(image, pred_rgb)
    loss_dict = {"rgb_loss": rgb_loss}
    return loss_dict    

def blend_background_for_loss_computation(
        self,
        pred_image: Tensor,
        pred_accumulation: Tensor,
        gt_image: Tensor,
    ) -> Tuple[Tensor, Tensor]:
    """Blends a background color into the ground truth and predicted image for
    loss computation.

    Args:
        gt_image: The ground truth image.
        pred_image: The predicted RGB values (without background blending).
        pred_accumulation: The predicted opacity/accumulation.

    Returns:
        A tuple of the predicted and ground truth RGB values.
    """
    background_color = self.background_color
    if background_color == "last_sample":
        background_color = "black"  # No background blending for GT
    elif background_color == "random":
        background_color = torch.rand_like(pred_image)
        pred_image = pred_image + background_color * (1.0 - pred_accumulation)
    gt_image = self.blend_background(gt_image, background_color=background_color)
    return pred_image, gt_image
```

f-dy:
See the top of https://github.com/nerfstudio-project/nerfstudio/pull/2025 for an explanation of how "random" works and why it's better

saleha-fatima:
> I think that this should be the original dataset: https://drive.google.com/file/d/13u5PF7OQDLLat1MFMKWJgRMV5jwBuhjm/view?usp=sharing @hoanhle let me know how it goes!

hi there, how did you prepare this dataset, its like very meticulously generated, would appreciate your help in this. also how were you able to produce  transforms.json file as the `ns-process-data` command does not seem to work with RGBA  images and in turn does not generate anything.

nepfaff:
@saleha-fatima I placed the YCB mustard bottle from [here](https://github.com/RobotLocomotion/models/tree/master/ycb) into a Drake simulator, spawned a camera sphere around it, and rendered with Drake's VTK rendering engine. See [here](https://github.com/nepfaff/sim2sim/blob/main/sim2sim/images/sphere_image_generator.py) for the rendering setup.

I generated the transforms file with a script like [this](https://github.com/nepfaff/scalable-real2sim/blob/main/scalable_real2sim/data_processing/nerfstudio.py). This is from our [Scalable Real2Sim](https://scalable-real2sim.github.io/) work that can use Nerfacto with alphatransparent training to get object-centric reconstructions from real-world robot data.

saleha-fatima:
@nepfaff I made a video, extracted frames from it at certain intervals, then by using sam2 segmented the object and added the alpha mask in the bg of images then did feature extraction, training (using the command mentioned in description) and on tsdf export got this result. i tried this process on several objects to rule out the fact that images might not be good, but the resultant 3d obj does not look any good to me. Attatching the video file here to view obj. if there is something that i am doing wrong or could do better, do let me know. 3d obj looked really good for the ycb mustard bottle, prolly because images are captured using robot and there is little to less motion blurr in it but if i train nerfacto on the data generated myself, results are not really pleasing
[Screencast from 12-22-2025 04:49:04 PM.webm](https://github.com/user-attachments/assets/3bb40c44-6f9d-4edc-b69e-71f70c6ea79e)

nepfaff:
@saleha-fatima those results look like what I'd expect from casual real-world capture and nerfacto.

Segmentation does make a difference. Sam3 has been released and is a big improvement from sam2. Switching might lead to slightly better results.

Another big factor is the pose estimation. How did you obtain the poses? Any errors here will make the results worse.

Next, image resolution makes a big difference. You want to use as high of a resolution as possible and as zoomed in onto the object as possible. I found this to make a huge difference for nerfacto. 

Consistent lighting is also important. A light ring might help a lot.

The ycb mustard bottle uses synthetic data with perfect segmentation and perfect camera poses. Hence, it represents an upper bound on the achievable quality.

However, if you really care about 3D geometry, then NeRF methods are not optimal. 2D Gaussian Splatting or Neuralangelo might be much better here. We added alpha-transparent training to Neuralangelo as part of our Scalable Real2Sim project from above. However, it is very slow. It might be worth looking for the current SOTA geometry reconstruction method (likely something like 2D gaussian splatting) and apply alpha-transparent training to that.

saleha-fatima:
you are right, actually earlier i was using gsplat (cuda accelerated rasterizer) to train 3d gaussian splatting model, it was working quite well and the rendered results were pretty good, only issue was that it produces 3d point cloud checkpoints and offer no obj export support (it would be great if there's any way out for this, already tried SuGaR model but it applies certain license), therefore had to shift to nerfacto but it seems to be very sensitive to image quality.
thanks for the suggestions, will try them out!

Issue #1811, opened 2023-04-25
Downscaling images could use multiple threads


elliotwoods:
**Is your feature request related to a problem? Please describe.**
When downscaling images with `ns-process-data` command, the process takes a very long time and the CPU is mostly idle during this time

**Describe the solution you'd like**
Multi-threaded solution for image resampling (i.e. perform on more than one image file at once)

**Describe alternatives you've considered**
Resampling in advance, but it's not clear in this case if there will be a mismatch with image features not being in the expected place

THE-COB:
The downscale uses ffmpeg so I think it will be faster if you install ffmpeg with GPU? I could be mistaken though

machenmusik:
Quick Google search found this, old but still perhaps [useful](http://www.akpdev.com/articles/2014/11/24/batch-media-processing-ffmpeg-python.html)

SharkWipf:
> The downscale uses ffmpeg so I think it will be faster if you install ffmpeg with GPU? I could be mistaken though

FWIW, no, for the downscaling step, this is not GPU accelerated (and would probably take longer to load up on a GPU for each frame than it'd save time).   

It just runs a single ffmpeg command sequentially for every frame right now, and as these ffmpeg commands aren't dependent on each other at all afaict, it should be fairly simple to just run them in parallel up to presumably the amount of cores available, which would, on my system, effectively result in a 32x speedup on currently one of the slower steps.

Alternatively, it might be possible to process images in batch using a single ffmpeg command and it might handle it automatically, but I don't know what the options here are. If possible this would save a good amount of time in spinning up new processes from scratch (especially on windows where spinning up a new process takes 1-2 extra seconds every time. With the default settings of 300 images and 3 downscales, that should be 900 - 1800 seconds saved before any parallelization!).

machenmusik:
My recollection is that it used to execute a single ffmpeg command, but that command was not reliably processing the correct files.  I agree that spawning a pool of them would likely be faster.

SharkWipf:
I submitted PR #2168 to tackle this, for conversion from video for all but equirects.
Equirects, and image to image downscaling, can be optimized too, but I haven't looked into that yet. Those would be a different path.

PR #2168, opened 2023-07-03
Merge video image extraction and downscaling

SharkWipf:
Disclaimer: Parts of this code were written by ChatGPT-4, albeit with my strict guidance.

This PR merges the thumbnail extraction step and the image downscaling step for `ns-process-data video` into a single ffmpeg pipeline.
This results in the overhead for downscaling almost completely vanishing.
On my machine, on Linux, the difference is very clear: 
Old:
```
(0) sebastiaan@AlmaLinux:nerf $ time ns-process-data video --data /home/sebastiaan/src/nerf/work/kitchen/kitchen_home_long_uwa.mp4 --output-dir /home/sebastiaan/src/nerf/work/out/colmaps/kitchen_home_long_uwa/1500 --camera-type perspective --num-frames-target 1500 --gpu --sfm-tool hloc --matching-method sequential
Number of frames in video: 17942
Number of frames to extract: 1632
[21:07:38] 🎉 Done converting video to images.                                                 process_data_utils.py:182
[21:25:25] 🎉 Done downscaling images.                                                         process_data_utils.py:380

real    19m38.797s
user    28m2.938s
sys     3m2.102s
```

New:
```
(0) sebastiaan@AlmaLinux:nerf $ time ns-process-data video --data /home/sebastiaan/src/nerf/work/kitchen/kitchen_home_long_uwa.mp4 --output-dir /home/sebastiaan/src/nerf/work/out/colmaps/kitchen_home_long_uwa/1500 --camera-type perspective --num-frames-target 1500 --gpu --sfm-tool hloc --matching-method sequential                                       
Number of frames in video: 17942
Number of frames to extract: 1632
[20:55:11] 🎉 Done converting video to images.                                                 process_data_utils.py:197

real    1m53.805s
user    13m3.816s
sys     0m14.801s
```

On Windows, the difference is likely orders of manitude largers, since spawning new processes on Windows can take up to 2 seconds, and this saves, in my example, 4895 executions of ffmpeg.

This should only affect the `ns-process-data video` pipeline, with the exception of equirectangular perspective.  
Equirectangular perspective follows the old flow, as it needed an extra step before downscaling.

It currently drops the possibility to enable nearest neighbor, which is not used by `ns-process-data video` command at all currently. If desirable I can easily add that in as well.

Note: I have not yet been able to fully test this or to compare the resulting images.
The results seemed fine when I prodded at them but I have not yet been able to test every scenario, nor completed a full training yet. Some more testing is recommended before merging.
I will be testing it further myself as well, but I'm less equipped for fully comparing all the edge cases.
Partially fixes issue #1811 for all videos but equirects.
Individual downscaling from images may be further enhanced too, but I haven't looked into this yet.

### DIFF
--- a/nerfstudio/process_data/process_data_utils.py
+++ b/nerfstudio/process_data/process_data_utils.py
@@ -182,13 +182,14 @@ def convert_video_to_images(
 
         if spacing > 1:
             CONSOLE.print("Number of frames to extract:", math.ceil(num_frames / spacing))
+            ffmpeg_cmd += " -vsync vfr -r1"
             select_cmd = f"thumbnail={spacing},setpts=N/TB,"
         else:
             CONSOLE.print("[bold red]Can't satisfy requested number of frames. Extracting all frames.")
             ffmpeg_cmd += " -pix_fmt bgr8"
             select_cmd = ""
 
-        downscale_cmd = f' -filter_complex "{select_cmd}{crop_cmd}{downscale_chain}" -vsync vfr -r 1' + "".join(
+        downscale_cmd = f' -filter_complex "{select_cmd}{crop_cmd}{downscale_chain}"' + "".join(
             [f' -map "[out{i}]" "{downscale_paths[i]}"' for i in range(num_downscales + 1)]
         )
 
--- a/nerfstudio/process_data/process_data_utils.py
+++ b/nerfstudio/process_data/process_data_utils.py
@@ -111,6 +111,7 @@ def convert_video_to_images(
     video_path: Path,
     image_dir: Path,
     num_frames_target: int,
+    num_downscales: int,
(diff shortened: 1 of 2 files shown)


SharkWipf:
So far I have found no issues, though I have not tested all use cases. I have successfully generated NeRFs of seemingly identical quality as before this change.

tancik:
It is indeed much faster, but there seems to be an issue when extracting all frames.

image

The number of images in the `images` folder is incorrect, however the number in `images_[2,4,8...]` are fine.

Works fine when there is a subset:

image

SharkWipf:
Huh, curious, I'll have a look

SharkWipf:
Seems it was indeed a -vsync vfr issue, I've made it so it only applies when thumbnail is used now, and it seems to have fixed the problem (and not have broken the thumbnail path in the process).

tancik:
Closing as this is now superseded by #2213

PR #2175, opened 2023-07-05
Remove pytorch hash table if tcnn implementation is used

georghess:
When selecting tcnn implementation for HashEncoding, a PyTorch version was still initialized and held in memory, but never used. Now, the PyTorch hash table is only created when selecting PyTorch implementation, or if tcnn is not available.


tancik (approved):
Thanks for fixing!

PR #2177, opened 2023-07-05
fix the ZeroDivisionError occured in int(image_width / aspect_ratio) when aspect_ratio equals zero

WebGao:


WebGao:
@colin-de

WebGao:
in the reviewing stage

tancik (approved):
LGTM

PR #2180, opened 2023-07-06
fix index out of bounds exception when using variable resolution with mask

yzslab:
Hi,
When using variable resolution images with masks, an `IndexError` exception will be thrown from [nerfstudio/data/pixel_samplers.py#L151](https://github.com/nerfstudio-project/nerfstudio/blob/c7def8ce8cb42fc2869a31ec886e5797e076489a/nerfstudio/data/pixel_samplers.py#L151).

Because `collate_image_dataset_batch_list()` requires `indices` to be a `nx3` tensor, but calling `self.sample_method()` with `mask=batch["mask"][i]` makes it become a `nx2` one.

This pull request will fix it.


machenmusik:
Thanks - is there an available sample dataset that can be used to verify?

PR #2186, opened 2023-07-07
Fix arguments that are not passed from config to NerfactoField

liu115:
Models like `nerfacto-big` and `nerfacto-huge` are changing `base_res` and `features_per_level` in the config. However, the arguments are not actually passed to NerfactoField currently.


tancik:
Good find! I think it makes sense to reset the nerfacto-big and huge configs so that the performance doesn't change (Im mostly worried about GPU memory). Later we can retune these values now that we know they will actually do something :)

liu115:
@tancik Thanks! Should I then update the configs and add it in this PR?

tancik:
That would be great!

tancik (approved):
Lgtm

PR #2201, opened 2023-07-11
Adding input handling for RFFEncoding's get_out_dim(self)

CardiacMangoes:
Introduced handling in `RFFEncoding` for method `get_out_dim(self)` so that if `self.include_input` is True, the result adds the dimension of `in_dim`.


tancik (approved):
LGTM

PR #2203, opened 2023-07-12
Add saving/loading scheduler state_dict

hyperzy:
- Add saving/loading scheduler state_dict to accommodate the training resumption on some methods with lr schedulers.


tancik (approved):
Thanks for adding

PR #2214, opened 2023-07-13
Try non-strict model loading on RuntimeError. ...

sjuxax:
This allows tcnn models that contained a torch hash_table to load.
In particular, it allows us to load tcnn models prior to PR #2175.

cf. #2175


tancik (approved):
LGTM

PR #2216, opened 2023-07-13
refactor: `get_normalized_directions`

SauravMaheshkar:
The `get_normalized_directions` and `shift_directions_for_tcnn` performed the same function but were placed in two separate files. This PR places them in `base_field` and refactors other usages.


SauravMaheshkar:
> LGTM. Just need to fix the test then it can be merged

Fixed in 6ff033d48d084a33ef6d6ed946e3a15a4360d24c

PR #2220, opened 2023-07-13
Reorient point cloud normals for better surface reconstruction

notagoose:
Normal estimation methods generally are ambiguous between a vector and its negation. Currently, the point cloud extraction picks the normal orientations arbitrarily, while Poisson surface reconstruction expects oriented normals. This leads to incorrect back-faces and extraneous seams in the meshes (below)
image
The view directions can be used to reorient the normals, leading to more consistent meshes.
Can be tested with `ns-export pointcloud --reorient-normals True --normal-method open3d --output-dir ./exports/reoriented --load-config {...}`


tancik (approved):
LGTM

PR #2213, opened 2023-07-13
Fast downscaling for video and image processing (PR2168 plus images)

machenmusik:
This PR extends #2168 by accelerating downscaling for image processing, including polycam upscaled depth.

PR #2229, opened 2023-07-14
Fix colmap video/ffmpeg error

SharkWipf:
I did an oopsie and forgot a space in my final commit for https://github.com/nerfstudio-project/nerfstudio/pull/2168 / https://github.com/nerfstudio-project/nerfstudio/pull/2213 it seems, which breaks ffmpeg.
This fixes that. My apologies.


SharkWipf:
As per #2230, seems like the -r 1 is entirely superfluous and in fact does not work together with -vsync vfr in ffmpeg 6. Which, tbf, makes sense. 
I do remember needing both to fix an issue, but perhaps that was in an older iteration, as this version seems to work fine without it.
I'll close this PR and make a new, clean one with it removed entirely.

PR #2232, opened 2023-07-14
add user UID as build argument

### LINKED ISSUE #2211: Docker default user
**Is your feature request related to a problem? Please describe.**
User UID does not always [equal to 1000](https://github.com/nerfstudio-project/nerfstudio/blob/cebe787bedffd4966fdcb067682da64e168b5565/Dockerfile#L102) (for example if you share a server with your colleagues). This leads to annoying permission denied errors when you mount your folder inside nerfstudio docker container and work with files. 

**Describe the solution you'd like**
Add build arguments for user UID (GID).

**Additional context**
Here is an example of how i [solved](https://gist.github.com/linukc/f793b9aa572c64096b5ef4d211870a84) this problem. It would be nice if we can potentially add this feature to the official docker image.

linukc:
This PR closes #2211.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -100,11 +100,13 @@ RUN git clone --branch 3.8 https://github.com/colmap/colmap.git --single-branch 
 
 # Create non root user that mimic host user and setup environment.
 ARG UID=1000
+ARG GID=1000
 RUN useradd -m -d /home/user -u ${UID} -G sudo user && \
+    groupmod -g ${GID} user && \
     # Set user password
-    && echo "user:user" | chpasswd \
+    echo "user:user" | chpasswd && \
     # Ensure sudo group users are not asked for a password when using sudo command by ammending sudoers file
-    && echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
+    echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
 
 # Switch to new uer and workdir.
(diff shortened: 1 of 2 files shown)


linukc:
@tancik Will my changes affect [dromni](https://hub.docker.com/r/dromni/nerfstudio/tags) docker images?

Ilyabasharov:
@linukc are you sure that adding only UID will solve this problem ? I still run into problems, see this [blog](https://techflare.blog/permission-problems-in-bind-mount-in-docker-volume/)

linukc:
@Ilyabasharov glad to see you and thank you for your comment! It was not smart of me to assume that all potential users would have the same GID as their UID.

linukc:
@Ilyabasharov In spite of this, even without specifying a GID, it should work for you as a user since the UID check is performed first. Tell me more about the problems you encountered. Also, it would be very helpful if you would double check my code.

Ilyabasharov:
I have GID 1002 and `docker build` gives me `groupmod: invalid group ID 'user'` using your code. I think that best practice is to use devcontainer. Where did you get such an error with permission denied? (#2211) Did you share the code between docker container and local machine?

linukc:
@Ilyabasharov I hope groupadd before useradd solves the problem.

tancik:
Is this PR still necissary?

linukc:
@tancik I guess after #2334 merge we can close this PR

PR #2234, opened 2023-07-16
- Load schedulers from ckpt in a backward-compatible manner

hyperzy:
- Allow loading old checkpoints without the `schedulers` key.


brentyi (commented):
Can we leave a comment explaining these conditions? (ideally with some info on what versions of `nerfstudio` this impacts)

hyperzy:
Versions <= 0.3.2 don't support loading schedulers, because the states of schedulers are not saved in Trainer.save_checkpoint(). Specifically, commits older than cebe787bedffd4966fdcb067682da64e168b5565 don't support schedulers loading/saving. 
These changes in the commit will save the states of schedulers by default, i.e. creating a `schedulers` key. Were the checkpoints from previous versions loaded, the loader will check whether the `schedulers` key exist before loading schedulers.

PR #2239, opened 2023-07-17
Add std for eval

Khoa-NT:
Add measure std

### DIFF
--- a/nerfstudio/pipelines/base_pipeline.py
+++ b/nerfstudio/pipelines/base_pipeline.py
@@ -183,12 +183,15 @@ class Pipeline(nn.Module):
 
     @abstractmethod
     @profiler.time_function
-    def get_average_eval_image_metrics(self, step: Optional[int] = None, output_path: Optional[Path] = None):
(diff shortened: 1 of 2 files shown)


nerfstudio/pipelines/base_pipeline.py
```diff
@@ -183,12 +183,15 @@ def get_eval_image_metrics_and_images(self, step: int):
 
     @abstractmethod
     @profiler.time_function
-    def get_average_eval_image_metrics(self, step: Optional[int] = None, output_path: Optional[Path] = None):
+    def get_average_eval_image_metrics(
+        self, step: Optional[int] = None, output_path: Optional[Path] = None, get_std: Optional[bool] = None
```
tancik, line 187: Any reason to use `Optional`?, why not `get_std: bool = False`

Khoa-NT: No reason. I revised the default value to False.

tancik (approved):
LGTM

Issue #1734, opened 2023-04-14
Any plan to add NeRFLix into nerfstudio?


HrsPythonix:
CVPR2023 releases a nerf enhancement paper named NeRFLix, in https://arxiv.org/abs/2303.06919, it seems like an universal module that can be applied to any NeRF method. I'm wondering if there would be any plans to add this into nerfstudio?

HrsPythonix:
Also, see Zip-NeRF which combines mipnerf and instant ngp, this might work also for nerfacto, https://arxiv.org/abs/2304.06706

avclubvids:
I was just coming here to ask if Zip-NeRF was on the roadmap for implementation in NeRF Studio – it is by far the best-looking model so far (which for most uses trumps any other consideration like training cost or speed etc.) So, +1 to this issue!

neekolas:
+1 on zip-nerf. There is even now an unofficial PyTorch implementation of it floating around. https://github.com/SuLvXiangXin/zipnerf-pytorch

huguensjean:
What is zipnerfacto? Is that ZipNERF? (https://jonbarron.info/zipnerf/) Anybody tried this oss zipnerf? (https://github.com/SuLvXiangXin/zipnerf-pytorch) Jon, Ben and I would see ZipNERF in Nerfstudio.

![85G6UmF64cXSEUh](https://github.com/nerfstudio-project/nerfstudio/assets/40362353/f56e39fb-3c52-4b91-b703-1dba2e93b137)

huguensjean:
I tried doing this (https://github.com/nerfstudio-project/nerfstudio/pull/2242) and I ran into an error. ns-train does not recognize zipnerf. Has anyone tried their own open source nerfstudio template? (for some reason, the OSS part is important) . TLDR. What ever your template does, make sure your repo license is there. Thank you LERF (LLLLOOOVVVVVVEEEEE) team!

huguensjean:
![6STfBQgSXCPiSx5](https://github.com/nerfstudio-project/nerfstudio/assets/40362353/4737e17b-31a8-4479-bfe4-bdc8a02e96d3)

With respect to lerf.io, when I try doing that with 8 V100s, the latency gets worse. I believe that bug was fixed? Please help!

huguensjean:
What about something like SMERF? I have plan to add nerfstudio to Google Cloud Vertex AI Model Garden. Can it run in gradio?

https://smerf-3d.github.io/

PR #2242, opened 2023-07-18
Jkulhanek/zipnerf

Ilyabasharov:
Add zipnerf code https://arxiv.org/pdf/2304.06706.pdf based on https://github.com/SuLvXiangXin/zipnerf-pytorch.
Also added:
- hash decay loss to nerfacto
- `LinearizedSceneContraction` from zipnerf paper. `SceneContraction` is not supported for zipnerf die to dimensions of Gaussian `cov`
- torch-scatter lib to compute regularizations
- and others


tancik:
Can you post render results in the description.

tancik:
Using `torch-scatter` is a little annoying, we will need to figure out how to handle installation (for non-docker workflow) since it can't be bundled in the pypi pip package. We could alert users to install instructions if they try to run code that uses it. Anyone have other ideas?

Ilyabasharov:
@tancik I've posted renders and metrics. 

Im using `torch_scatter` twice. I can replace `scale_featurization` via [torch.scatter_reduce](https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_reduce_.html#torch.Tensor.scatter_reduce_) but for `regularize_hash_pyramid` it does not have backward pass.

tancik:
I think we need to think a bit on how this gets merged with the main repo. A few things to consider:
+ torch scatter makes installation more difficult
+ many changes break backwards compatibility
+ PR contains changes that are not related to zipnerf (ie num default samples in nerfacto)
Additionally we are moving in the direction of incorporating external models via the registry instead of directly into the repo. There is a number of reasons for this, but one big one is that maintaining a growing list of methods in the main repo isn't sustainable (ie github issues get polluted and there is no way to pin an external method to a nerfstudio version). My recommendation would be to split this PR up into core nerfstudio changes so that we can consider them each in turn. The actual zip-nerf model can exist in an external github repo (ie [LERF](https://github.com/kerrj/lerf)), this external repo can have the torch scatter dependency

Ilyabasharov:
1) what if I will solve problem with `torch-scatter` library?

2) > many changes break backwards compatibility

yeah, because `nerfstudio` does not fully support gaussians. It is more focused on working with point-based methods. I had to change the structure of the code a bit. I also checked to make sure nothing broke.

3) > PR contains changes that are not related to zipnerf (ie num default samples in nerfacto)

fully agree, I can fix this

4) zip nerf is a great bridge between fast point-based and slow multiscale-based methods. I saw a lot of reactions in [#1734](https://github.com/nerfstudio-project/nerfstudio/issues/1734) and branches from @jkulhanek, @kerrj. Also zipnerf features can be applied to nerfacto and other hash-based methods. Because of this, I think that zipnerf should be included to nerfstudio

Ilyabasharov:
> Hi @Ilyabasharov
> 
> I'm doing some studies and I'm trying to test the ZipNerf in NerfStudio. I pulled the branch, did the installation again, updated the cli, but it gives an error saying that the zipnerf option for the training is invalid. Could you explain how to upgrade to this option, zipnerf, is available?

hello, It is strange, did you solve the problem?

Ilyabasharov:
> > Hi @Ilyabasharov
> > I'm doing some studies and I'm trying to test the ZipNerf in NerfStudio. I pulled the branch, did the installation again, updated the cli, but it gives an error saying that the zipnerf option for the training is invalid. Could you explain how to upgrade to this option, zipnerf, is available?
> 
> hello, It is strange, did you solve the problem? For me it is worked

Ilyabasharov:
@jingyibo123 hello, yes, I set default implementation to `torch`. But it is working also with `tcnn` which is much faster

nerfstudio/model_components/losses.py
```diff
@@ -125,6 +118,70 @@ def interlevel_loss(weights_list, ray_samples_list) -> torch.Tensor:
...
+    return ret
+
+
+def zipnerf_loss(weights_list, ray_samples_list):
+    """Anti-aliased interlevel loss proposed in ZipNeRF paper."""
+    # ground truth s and w (real nerf samples)
+    # This implementation matches ZipNeRF up to the scale.
+    # In the paper the loss is computed as the sum over the ray samples.
+    # Here we take the mean and the multiplier for this loss should be changed accordingly.
+    pulse_widths = [0.003, 0.03]
```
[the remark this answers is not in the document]
Ilyabasharov: proposed loss function for zipnerf just increasing during training. if you implemented it, could you please suggest some fix?
Screenshot 2023-08-10 at 17 16 07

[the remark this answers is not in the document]
Ilyabasharov: interlevel zip loss is increasing. if you already implement it, please help to find bug
Screenshot 2023-08-11 at 10 11 38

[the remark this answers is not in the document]
Ilyabasharov: I have found the bug, `sorted_interp_quad` was written in not correct way

DoriandPetit:
Hello, the implementation seems to work well on real data ! 

However, I can not make it work on Blender data, as the result is an all-white scene whatever the object, which is a bit odd as it is announced to be working on the torch implementation (I did not test it myself though). I am willing to work on it if needed but I don't really know where to begin the debug. 

What's strange is that (normalized) accumulation displayed by the viewer shows the object (its outline at least) (the word "normalized" being the key problem here I guess), and I also tried to display a PCA of the features codes of each samples * weights of each sample after accumulation and we can also see the object's outline... Which means that the model seems to learn something. (I tried lots of hyperparameters combination, including the usual modifications asked for the blender dataset such as near and far plane, etc...)

nerfstudio/field_components/encodings.py
```diff
@@ -381,6 +426,31 @@ def pytorch_fwd(self, in_tensor: Float[Tensor, "*bs input_dim"]) -> Float[Tensor
 
         return torch.flatten(encoded_value, start_dim=-2, end_dim=-1)  # [..., num_levels * features_per_level]
 
+    def regularize_hash_pyramid(
+        self,
+        regularize_fn: Callable[[Tensor], Tensor] = torch.abs,
+    ) -> Float[Tensor, "0"]:
+        """Regularize hash pyramid weights."""
+        hash_decay = segment_coo(
```
georghess, line 434: For avoiding torch_scatter dependency
```suggestion
        hash_decay = torch.zeros((self.num_levels, 2), device=self.hash_table.device)
        hash_decay = hash_decay.scatter_reduce(
            dim=0,
            index=self.level_indexes[..., None].repeat(1, self.features_per_level),
            src=regularize_fn(self.hash_table.view(-1, self.features_per_level)),
            reduce="mean",
        ).mean(dim=-1).sum()
```

Ilyabasharov: Thanks! I kept in mind this solution. but how to solve dependency with `scale_featurization` ?

georghess: I suppose
```
scale_feat = torch.zeros(self.num_levels, device=self.hash_table.device)
scale_feat = scale_feat.scatter_reduce(
            dim=0,
            index=self.level_indexes,
            src=self.hash_table.view(-1, self.features_per_level).pow(2).sum(-1),
            reduce="mean",
)
```
should do the trick?

nerfstudio/fields/density_fields.py
```diff
@@ -115,5 +122,126 @@ def get_density(self, ray_samples: RaySamples) -> Tuple[Tensor, None]:
...
+        assert self.spatial_distortion is not None, "Spatial distortion in ZipNeRF should be defined."
+
+        gaussian_samples = self.spatial_distortion(gaussian_samples)
+        mean = gaussian_samples.mean
+        cov = gaussian_samples.cov
+
+        # shift mean in cube from [-2, 2] to [0, 1]
+        mean = (mean + 2.0) / 4.0
+        # cov in range [0, 1]
+        cov = cov / 2.0
```
georghess, line 215: It seems most of the code uses "cov" as standard deviation. If this is the case here as well, I think it should be: 
```suggestion
        cov = cov / 4.0
```

nerfstudio/field_components/encodings.py
Ilyabasharov: thanks! this works!

Ilyabasharov: @georghess 
Screenshot 2023-10-22 at 19 14 28

> For avoiding torch_scatter dependency

georghess: Sorry, 
`hash_decay = torch.zeros((self.num_levels, 2), device=self.hash_table.device)`
should be
`hash_decay = torch.zeros((self.num_levels, self.features_per_level), device=self.hash_table.device)`

However, for my applications, a simple for loop was way faster.

bobye:
@Ilyabasharov what your plan of this PR? Where can I find the latest version of zip nerf implementation?

I think a high quality nerf method could still have a lot value to be maintained as an external method. I can probably help turn this PR to a separate we’ll maintained repo.

jingyibo123:
@bobye  [Here's one](https://github.com/SuLvXiangXin/zipnerf-pytorch), I thoroughly compared these two, some of the components are fundamentally different, numerically different as well, the best is to test with your own data..

georghess:
FYI, now the official implementation is out at https://github.com/jonbarron/camp_zipnerf

PR #2264, opened 2023-07-25
Visibility Field from Nerfbusters

ethanweber:


Ilyabasharov:
@ethanweber I think now `VisibilityField` name is too common. There are many papers that using `VisibilityField` too, so it would be better to distinguish them

ethanweber:
> @ethanweber I think now `VisibilityField` class name is too common. There are many papers that using `VisibilityField` too, so it would be better to distinguish them in this framework

Oh interesting, thanks for the heads up. Can you point me to some papers that have the same name?

nerfstudio/cameras/cameras.py
```diff
@@ -747,8 +747,6 @@ def _compute_rays_for_vr180(
 
             rotation = c2w[..., :3, :3]
 
-            -torch.pi * ((x - cx) / fx)[0]
```
ethanweber, line 750: Is this a typo?

nerfstudio/model_components/renderers.py
```diff
@@ -335,47 +335,46 @@ def forward(
         ray_samples: RaySamples,
         ray_indices: Optional[Int[Tensor, "num_samples"]] = None,
         num_rays: Optional[int] = None,
-    ) -> Float[Tensor, "*batch 1"]:
+        quantity: Optional[Float[Tensor, "*batch num_samples 1"]] = None,
```
ethanweber, line 338: What are your thoughts on this @FrederikWarburg? I think maybe we should have a separate renderer?

Ilyabasharov:
> > @ethanweber I think now `VisibilityField` class name is too common. There are many papers that using `VisibilityField` too, so it would be better to distinguish them in this framework
> 
> Oh interesting, thanks for the heads up. Can you point me to some papers that have the same name?

[Block-NeRF paper](https://arxiv.org/pdf/2202.05263.pdf)

nerfstudio/cameras/cameras.py
FrederikWarburg: not sure what this line was doing there... the value is not assigned to anything, so I deleted it.

nerfstudio/model_components/renderers.py
FrederikWarburg: So the main differences are:

1. rename steps -> quantity
2. add the argument quantity.

Don't think these changes justify a separate renderer?

nerfstudio/model_components/renderers.py
```diff
@@ -335,47 +335,46 @@ def forward(
...
             num_rays: Number of rays, used when samples are packed.
+            quantity: If specified, then take the median of this quantity instead of the depth.
 
         Returns:
             Outputs of depth values.
         """
 
-        if self.method == "median":
-            steps = (ray_samples.frustums.starts + ray_samples.frustums.ends) / 2
+        if quantity is None:
```
FrederikWarburg, line 353: move median logic to function -> top of file.
Create a new renderer called VisibilityRenderer -> only does median depth.

nerfstudio/models/nerfacto.py
```diff
@@ -221,6 +226,8 @@ def update_schedule(step):
         self.renderer_rgb = RGBRenderer(background_color=self.config.background_color)
         self.renderer_accumulation = AccumulationRenderer()
         self.renderer_depth = DepthRenderer()
+        self.renderer_depth_expected = DepthRenderer(method="expected")
+        self.renderer_depth_median = DepthRenderer(method="median")
```
FrederikWarburg, line 230: replace VisibilityRenderer

nerfstudio/models/nerfacto.py
```diff
@@ -295,13 +305,32 @@ def get_outputs(self, ray_bundle: RayBundle):
             "depth": depth,
         }
 
+        if self.visibility_field is not None:
+            assert isinstance(self.visibility_field, VisibilityField), "self.visibility_field must be a VisibilityField"
+            visibility_samples = self.visibility_field.forward(ray_samples_list[-1])
+            visibility = self.renderer_depth_expected(
```
FrederikWarburg, line 311: remove expected visibility

nerfstudio/pipelines/base_pipeline.py
```diff
@@ -227,6 +222,20 @@ class VanillaPipelineConfig(cfg.InstantiateConfig):
...
+    """how often to apply the visibility loss"""
+    visibility_loss_quantity: Literal["weights", "densities"] = "densities"
+    """whether to apply the visibility loss to weights or densities"""
+    visibility_loss_mult: float = 1.0
+    """multiplier for the visibility loss"""
+    visibility_min_views: int = 1
+    """minimum number of views for the visibility loss"""
+    visibility_num_rays: int = 10
+    """number of rays per batch to use for the visibility loss"""
+    visibility_radius: float = 1.0
```
FrederikWarburg, line 237: 2.0

nerfstudio/pipelines/base_pipeline.py
```diff
@@ -312,6 +327,46 @@ def get_train_loss_dict(self, step: int):
...
+                center=center,
+            )
+            # We only use the first camera index, but it doesn't matter since we don't care about appearance embeddings
+            # when using the visibility loss.
+            camera_indices = torch.tensor([0] * self.config.visibility_num_rays).unsqueeze(-1)
+            ray_bundle_rays = cameras.generate_rays(camera_indices)
+            # ray_bundle is (1, 1, self.config.visibility_num_rays)
+            ray_bundle_rays = ray_bundle_rays.flatten()
+            model_outputs_rays = self.model(ray_bundle_rays)
+            quantity_list = model_outputs_rays[
```
FrederikWarburg, line 356: always use densities -> simply a bit

FrederikWarburg: clean a bit

ethanweber:
Hey @tancik, do you have thoughts on how the visibility loss should be added wrt. the vanilla pipeline? Do you think it should be its own pipeline?

PR #2268, opened 2023-07-26
Use correct pixel area in volumetric sampler

hturki:
The existing volumetric sampler code returns 0 as the pixel areas. This PR returns the correct values instead.


tancik (approved):
LGTM

PR #2277, opened 2023-07-28
More robust install instructions

brentyi:
I just tried an installation on a fresh Ubuntu 22.04 machine, with no system-level CUDA installation.

Unsurprisingly (and as mentioned by @kpertsch), the `tiny-cuda-nn` install step fails without `cuda-toolkit` installed. Running the extra install for `cuda-toolkit` fixes it. I think we should add it to the setup instructions.

A few bits of uncertainty before merging:
- Will a system-level CUDA install break anything here?
- Possibly related: a few people have mentioned needing to fuss with environment variables for all of the linking etc to work. I've never needed to, what's expected here?
- @kpertsch also reported needing to update `gcc` via `conda install -c conda-forge gcc=11.2 gxx=11.2`. We may want to add this to the FAQ.
- Perhaps we should add a note about supported NVIDIA driver versions? The current state of which NVIDIA drivers are compatible which CUDA versions is not entirely clear even to myself... (can a driver be "too new" for CUDA 11.7 etc)

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,17 +92,20 @@ Nerfstudio requires `python >= 3.8`. We recommend using conda to manage dependen
 ```bash
 conda create --name nerfstudio -y python=3.8
 conda activate nerfstudio
-python -m pip install --upgrade pip
+pip install --upgrade pip
 ```
 
 ### Dependencies
 
-Install pytorch with CUDA (this repo has been tested with CUDA 11.7 and CUDA 11.8) and [tiny-cuda-nn](https://github.com/NVlabs/tiny-cuda-nn)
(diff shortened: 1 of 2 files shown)


maturk:
I have found that conda installing cuda-toolkit automatically sets correct Cuda home and other environment variables in a local conda env. I think this is a good addition to the instructions and I have personally seen this issue pop up multiple times in the Discord and conda installing has always fixed it, even if you already have system level cuda installed.

kerrj:
Trying this on a new machine, installation fails on python 3.11 since open3d isn't supported on it yet. Might want to specify >=3.8 and <3.11 in the instructions

kerrj:
3.10 works for me

kerrj (approved):
Changes SGTM as long as they pass doc checks etc, I only tested the instructions themselves not the website layout etc

PR #2282, opened 2023-07-31
Moved Volinga as an external method

Frivas97:
Volinga is now an external method, hosted in https://github.com/Volinga/volinga-model.


tancik (approved):
Lgtm

Issue #1638, opened 2023-03-20
Training can be potentially speed up by 2x?


liruilong940607:
On [this branch](https://github.com/nerfstudio-project/nerfstudio/tree/ethan/benchmarks) training a Mip-NeRF 360 scene (garden) for 20k iterations took 11mins to finish. On wandb it shows 150k training rays/sec and 706k eval rays/sec. 

On the other hand I implemented the nerfacto model in the nerfacc repo and using [this script](https://github.com/KAIR-BAIR/nerfacc/blob/17e28de798dca5d2e2761ff8d514b8b4df26ef6f/examples/train_ngp_nerf_prop.py) to train the garden scene for 20k iterations with the same batch size took only 5mins to finish (see numbers [here](https://www.nerfacc.com/en/latest/examples/unbounded.html)).  That leads 266k training rays/sec and 760k eval rays/sec.  So there should be a ~2x speedup opportunity in the current pipeline. 

 The command I used to reproduce the training on the ethan/benchmarks branch:

```python
command = " ".join(
    (
        f"ns-train {method}",
        "--vis wandb",
        f"--data { data_path / capture_name}",
        "--output-dir outputs/nerfacto-ablations",
        "--steps-per-eval-batch 0 --steps-per-eval-image 0 --steps-per-save 20000",
        "--steps-per-eval-all-images 20000 --max-num-iterations 20001",
        f"--wandb-name {capture_name}_{table_row_name}",
        f"--experiment-name {capture_name}_{table_row_name}",
        # extra_string,
        table_row_command,
    )
)
```

Moreover, I observe the GPU utilization is pretty low during the training (GPU 5 is training with nerfstudio and GPU 6 is training with nerfacc). I suspect it might be the dataloader that is bottlenecking the pipeline. 
<img width="320" alt="Screen Shot 2023-03-20 at 12 22 03 AM" src="https://user-images.githubusercontent.com/10151885/226460647-278e6cd2-f2aa-440b-9743-12c12c1cd877.png">

Not sure if it worth to look into this or it is already a known situation to tradeoff flexibility and GPU memory for speed. Apparently nerfacc preload all images while nerfstudio is loading on the fly.

Happy to look into it once I get my hands freed but think would be worthy to post it here and see if anyone wants to take a bite on it.

machenmusik:
(Maybe another possibility is viewer contention?)

tancik:
[...]

PR #2288, opened 2023-08-01
Option to process images on GPU for speed at the expense of memory

AX-I:
Analogous to #2110. This fix speeds up training by over 3x when using multi-gpu DDP.

The observation that prompted this PR: training with `--num-devices=2` is significantly slower than training with single gpu, with extremely high cpu usage (~2600% on a 28-core cpu).

Profiling with `--logging.profiler=pytorch` reveals that `collate_image_dataset_batch()` in `pixel_samplers.py` takes 70ms with `--num-devices=2` and 30ms with single gpu. The batch image tensors are stored on the cpu, so the indexing operation is cpu-intensive.

Profile

This PR adds an option to remove `"image"` from `exclude_batch_keys_from_device` to keep the image tensor on the gpu. Using the option with suitable hardware reduces cpu usage from 2600% to 200%, and speeds up 2-gpu training **from ~70 Krays/s to ~250 Krays/s**. Single-gpu training is also sped up from 100 to 140 Krays/s. VRAM usage is increased from 2gb to 5gb.

Profiling the 2-gpu training with this fix confirms that the pixel sampler is no longer a bottleneck.

ProfileAfter

---
Training configuration: nerfacto default params, on F2Nerf grass scene
Machine: 2x Xeon Gold 5120 (disabled HT), 2x Titan RTX

### DIFF
--- a/nerfstudio/data/datamanagers/base_datamanager.py
+++ b/nerfstudio/data/datamanagers/base_datamanager.py
@@ -117,6 +117,8 @@ class DataManagerConfig(InstantiateConfig):
     """Specifies the camera pose optimizer used during training. Helpful if poses are noisy."""
     masks_on_gpu: Optional[bool] = None
     """Process masks on GPU for speed at the expense of memory, if True."""
+    images_on_gpu: Optional[bool] = None
+    """Process images on GPU for speed at the expense of memory, if True."""
 
 
 class DataManager(nn.Module):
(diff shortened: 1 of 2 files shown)


AX-I:
Might also help with #1638

AX-I:
> The "expense of memory" is the tricky part. Many users have 8gb cards which `nerfacto` has been designed to fit onto. When the dataset is moved to GPU, the memory is no longer predictable.

Converted to a config option! That should allow for all use cases.

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -117,6 +117,8 @@ class DataManagerConfig(InstantiateConfig):
     """Specifies the camera pose optimizer used during training. Helpful if poses are noisy."""
     masks_on_gpu: Optional[bool] = None
     """Process masks on GPU for speed at the expense of memory, if True."""
+    images_on_gpu: Optional[bool] = None
```
tancik, line 120: Doesn't need to be `Optional`. Use `images_on_gpu: bool = False`.
The same should be true for `masks_on_gpu`

tancik (approved):
LGTM

PR #2289, opened 2023-08-01
Remove system-level CUDA install from install instructions

brentyi:
Shouldn't be necessary when `cuda-toolkit` is installed via `conda`.


tancik:
On a similar note, it is a bit confusing that we have 11.7 and 11.8 docs here - 
image

tancik (approved):

brentyi:
Oops yeah, just un-flipped it!

PR #2292, opened 2023-08-02
Allow same dimensions optimizations to be disabled

### LINKED ISSUE #2293: Fast downscaling breaks for input sets that are not same dimensions
(e.g. mixed portrait and landscape images)

machenmusik:
Use `ns-process-data images --no-same-dimensions` ... to handle image sets with different dimensions properly, by disabling same-dimensions optimizations.  (Fast downscaling assumes image dimensions are the same.)

Fixes #2293 when --no-same-dimensions is specified, which behaves correctly but unfortunately runs slower.


tancik (approved):
LGTM, thanks for fixing

PR #2295, opened 2023-08-02
convert raw to JPG not TIFF, to avoid fast-downscale ffmpeg error

### LINKED ISSUE #2296: Fast downscale breaks raw input images
The fast-downscale path apparently relies on being able to overwrite the original file (eg. with crop etc. applied) which fails for TIFF, and unfortunately that is currently what RAW input images use for conversion.  #2295 should fix it.

machenmusik:
The fast-downscale path apparently relies on being able to overwrite the original file (eg. with crop etc. applied) which fails for TIFF, but works for JPG.  This restores support for raw images as input; fixes #2296.


tancik (approved):
LGTM

Issue #2304, opened 2023-08-04
ns-train depth-nerfacto with mask dataset fails


machenmusik:
Using a dataset with masks, `ns-train depth-nerfacto` fails.
(Even before the NotImplementedError)
```
nerfstudio\nerfstudio\data\pixel_samplers.py:382 in sample_method                        │
│                                                                                                  │
│   379 │   │   mask: Optional[Tensor] = None,                                                     │
│   380 │   │   device: Union[torch.device, str] = "cpu",                                          │
│   381 │   ) -> Int[Tensor, "batch_size 3"]:                                                      │
│ ❱ 382 │   │   if mask:                                                                           │
│   383 │   │   │   # Note: if there is a mask, should switch to the base PixelSampler class       │
│   384 │   │   │                                                                                  │
│   385 │   │   │   raise NotImplementedError()                                                    │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
RuntimeError: Boolean value of Tensor with more than one value is ambiguous
```

ethanweber:
Hey @machenmusik, I'll work on a PR to resolve this issue.

ethanweber:
Hey, I mentioned you in the PR #2305.

machenmusik:
#2305 solves except for Windows which also needs #2303.

machenmusik:
Fixed thx all!

PR #2305, opened 2023-08-04
Adding support for PairPixelSampler and PatchPixelSampler masks

ethanweber:
PR to address Issue #2304. This PR adds mask support for the PairPixelSampler by using an erode function.


machenmusik:
Thanks - even with `--pipeline.datamanager.masks-on-gpu True` it is slow, so may take a while to provide feedback.

machenmusik:
It works with masked datasets (which unfortunately I can't share) -- although hard to say whether depth-nerfacto improves significantly over (non-depth) nerfacto so far, it is obviously better than failing with NonImplementedError.

machenmusik (approved):

machenmusik:
Doing a quick A-B between defaults of depth-nerfacto and nerfacto using generated depth, my quick take is that depth-nerfacto avoids ghosting, but seems less sharp.  Currently I am wondering whether the default resolution for Zoe depth image generation is too low, and/or the default parameters no longer match.

ethanweber:
I also added support for patches.

AdamRashid96 (commented):
This all LGTM! I don't have data with masks to test it though so if @machenmusik could try the patch sampler with masks to see if it works that would be great.

machenmusik:
There is a mustard dataset referenced in a few of the older PRs around masks... anyone recall which methods are currently using patch pixel sampling to try with?

machenmusik:
Not in position to test the patch sampling pieces at the moment, but it can't be worse than NotImplementedError failure, so will merge.

PR #2306, opened 2023-08-04
Cat function for TensorDataclass

ethanweber:
This is helpful to e.g., concatenating cameras. I add a call of this function to `render.py` to start and end at the same camera when rendering with the interpolate mode.

### DIFF
--- a/nerfstudio/scripts/render.py
+++ b/nerfstudio/scripts/render.py
@@ -497,6 +497,8 @@ class RenderInterpolated(BaseRender):
     """Frame rate of the output video."""
     output_format: Literal["images", "video"] = "video"
     """How to save output data."""
+    loop: bool = False
+    """Whether to start and end with the first camera."""
 
     def main(self) -> None:
         """Main function."""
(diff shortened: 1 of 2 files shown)


nerfstudio/utils/tensor_dataclass.py
```diff
@@ -195,6 +195,28 @@ def ndim(self) -> int:
...
+        """
+        if not isinstance(other, TensorDataclass):
+            raise TypeError("Expected 'other' to be a TensorDataclass.")
+
+        def tensor_fn(field_name, x):
+            return torch.cat([x, getattr(other, field_name)], dim=dim)
+
+        dataclass_fn = None
+
+        custom_tensor_dims_fn = None
```
ethanweber, line 216: Could someone more familiar help me complete this? @tancik @jake-austin

I'm not too sure what the functions should look like.

nerfstudio/scripts/render.py
tancik, line 1: Remove

jkulhanek:
Could the function be a classmethod and accept a list of objects? ...to have a similar interface as PyTorch?

PR #2312, opened 2023-08-07
Enable SparseNerf depth ranking loss to be used with GT/Sensor depth

alexsax:
PR #2066 adds a very nice ranking-based loss that can be used with non-metric depth. The ranking loss can also be used with metric depth, and seems to provide smoother results on the replica dataset (image below). 

This PR is a minimal set of changes that allows the ranking loss to be used with depth read from disk (usually GT/sensor depth).

image


ethanweber (commented):
These changes look great! Can approve once it's merged again with main and passing the checks.

AdamRashid96 (approved):
LGTM!

PR #2315, opened 2023-08-08
Monodepth regularization bug fix

AdamRashid96:
- Fixed a bug where we were not switching to true depth loss, when true depth data is available.


Ilyabasharov:
@AdamRashid96 also I have found bug that at different runs, the depth map cache is used, but the indexes of the validation and train samples change. Therefore, either fix the train and validation, or generate the depth for all images, and then read by indexes.

AdamRashid96:
@Ilyabasharov  Thanks for pointing out this bug. I'll figure out whether to move the depth map generation to a different script or not, but I'll fix this indexing bug.

Ilyabasharov:
@AdamRashid96 did you plan to realize spatial continuity regularization from SparseNeRF?

kevin-thankyou-lin:
@AdamRashid96 What's the status on this PR? Seems like generating the depth maps in `ns-process-data` could be a good option? @brentyi @kerrj

PR #2317, opened 2023-08-08
save sfm depth as png to keep 16bit. save depth_file_path as relative…

zhan-xu:
bug fixed:
1. When raw images are in JPG format, the depths will be saved as JPG as well. This losses numerical precision from uint16 to uint8.
2. The depth_file_path in transforms.json is not correct. Apply "relative_to" as has been done to "mask_path".


tancik (approved):
LGTM

Issue #2313, opened 2023-08-07
Monodepth PR broke `PatchPixelSampler`


pwais:
The old `PatchPixelSampler` respect the batch size (e.g. `--pipeline.datamanager.train-num-rays-per-batch` ) and adjusted accordingly:
https://github.com/nerfstudio-project/nerfstudio/blob/d542646c72e2f7c8cd03f6677b7489334c1d92c8/nerfstudio/data/pixel_samplers.py#L247C7-L247C24

But the monodepth PR https://github.com/nerfstudio-project/nerfstudio/pull/2284  made `PatchPixelSampler` no longer respect the parameter `batch_size` https://github.com/nerfstudio-project/nerfstudio/pull/2284/files#diff-b15e0d31bf4d84dbacd4ea2e0c44f597dea42d975f841a5cb47481e876185c4bR375

I tried to fix this but that PR actually appears to change the `PairPixelSampler` logic too, so I just removed `PairPixelSamplerConfig` from the depth-nerfacto config for now.

Nerfstudio might need to remove the separate `--pipeline.datamanager.pixel-sampler.num-rays-per-batch` that conflicts semantically with `--pipeline.datamanager.{train,eva}-num-rays-per-batch` ... that could be why setting the batch size inside the pixelsamplers got spaghetti'ed

Nerfstudio might also need some tests where the batch size params are varied, dunno if those can be added to CI

@ethanweber

elenacliu:
I met a similar problem when export the point cloud result of depth-nerfacto:

> File "/nerfstudio/nerfstudio/data/pixel_samplers.py", line 176, in collate_image_dataset_batch
    assert collated_batch["image"].shape[0] == num_rays_per_batch
AssertionError

alexsax:
Yes it looks like `PairPixelSampler.sample_method()` accepts `batch_size` but [never uses it](https://github.com/nerfstudio-project/nerfstudio/blob/cb2405d3eaa2b848acb82e6a8da780ac40b937ce/nerfstudio/data/pixel_samplers.py#L393-L422).

PR #2327 fixes this :)

pwais:
Thank you so much !!!

tancik:
> err, i don't think `batch_size` is ever Optional (it isn't in the base class) nor is it ever null
> 
[...]

PR #2327, opened 2023-08-10
Update pixel_samplers.py

alexsax:
Now uses the specified batch_size param in sample_method(), fixes the pointcloud exporting error in #2313

### DIFF
--- a/nerfstudio/data/pixel_samplers.py
+++ b/nerfstudio/data/pixel_samplers.py
@@ -392,7 +392,7 @@ class PairPixelSampler(PixelSampler):  # pylint: disable=too-few-public-methods
     # overrides base method
     def sample_method(  # pylint: disable=no-self-use
         self,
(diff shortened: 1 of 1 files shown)


nerfstudio/data/pixel_samplers.py
```diff
@@ -392,7 +392,7 @@ def __init__(self, config: PairPixelSamplerConfig, **kwargs) -> None:
     # overrides base method
     def sample_method(  # pylint: disable=no-self-use
         self,
-        batch_size: int,
+        batch_size: Optional[int],
```
pwais, line 395: tanku for this fix!

is this param really optional tho?  in the superclass it is not optional.  it's a little confusing because the pixel sampler seems to have a side API for configuring batch size even tho notionally the batch size config can be passed in via a different flag

tancik (approved):
LGTM

PR #2332, opened 2023-08-12
Ddp docs

yurkovak:
Looks like `num_gpus` was renamed to `num_devices` in #1936, fixing the remaining usage of the old name


brentyi (approved):
Looks good to me!!

PR #2335, opened 2023-08-13
fix #2318: add train/test/eval split for instant ngp dataparser

### LINKED ISSUE #2318: why are there weird red margins of my loaded images?
**Describe the bug**
When I load the images, they have red margins.

**To Reproduce**
I just install the latest version of nerfstudio (commitid 9b6010ea), and pip install -e . to install it.

**Expected behavior**

**Screenshots**
image

**Additional context**

elenacliu:
To support issue #2318 problems, I choose to add split codes similar to nerfstudio data parser for instant ngp dataparser.

After I add the code, I run `ns-train nerfacto --data ${mypath} --output ${mypath} instant-ngp-data` and open it in the viewer, I can see this, 

image

which shows the split has been made.


tancik (approved):
LGTM.

PR #2341, opened 2023-08-14
Circle crop to images processing pipeline

kerrj:
Percent_radius_crop was inside video processing, but not images. This fixes that


brentyi (approved):
LGTM

PR #2343, opened 2023-08-15
viser 0.1.0, fix type checker errors

brentyi:
The latest release of `viser` renames `GuiHandle` to `GuiInputHandle`, since there are now non-input GUI handles. I added a shim so it shouldn't produce any runtime errors (and/or actually break any `nerfstudio` installations), but it's hidden to the type checker and will produce a `pyright` error.


tancik (approved):
LGTM

PR #2351, opened 2023-08-18
Nerfstudio Feedback Form

ethanweber:
Adding a Nerfstudio feedback form to the readme and docs. The URL will be updated before merging.


tancik (approved):
LGTM once urls are provided

PR #2366, opened 2023-08-23
BARF scheduling for gradients

kerrj:
Some papers say that instead of turning hashgrids on over time, weighting the gradients for the hashgrids results in better performance for NGP BARF implementations. This implements that.
See: https://arxiv.org/abs/2302.01571, https://camp-nerf.github.io/

### DIFF
--- a/nerfstudio/field_components/encodings.py
+++ b/nerfstudio/field_components/encodings.py
@@ -18,7 +18,7 @@ Encoding functions
 
 import itertools
 from abc import abstractmethod
-from typing import Literal, Optional, Sequence
+from typing import Literal, Optional, Sequence, Tuple
 
 import numpy as np
 import torch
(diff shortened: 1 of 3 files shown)


Ilyabasharov:
@kerrj Did you test it? 
individualImage
I've run nerfacto-huge and attached visualization at the late stages of training when the gradient BARF is no longer working

kerrj:
Interesting, does this scene diverge without the scheduling for you as well?

kerrj:
Also how many scheduling steps did you use here?

maturk:
> @kerrj Did you test it? individualImage I've run nerfacto-huge and attached visualization at the late stages of training when the gradient BARF is no longer working

Yes, as with the earlier BARF implementation, it seems to degrade performance compared to No BARF when you have good poses. I've been testing and noticed similar drop in PSNR. However, it is interesting to always test on really bad poses...

Ilyabasharov:
> Interesting, does this scene diverge without the scheduling for you as well?

nono, all good without gradient BARF

Ilyabasharov:
> > @kerrj Did you test it? individualImage I've run nerfacto-huge and attached visualization at the late stages of training when the gradient BARF is no longer working
> 
> Yes, as with the earlier BARF implementation, it seems to degrade performance compared to No BARF when you have good poses. I've been testing and noticed similar drop in PSNR. However, it is interesting to always test on really bad poses...

unfortunately, I can't explain why even when BARF is shut down during network training (fine_iters * max_iters), it can't converge and artifacts still remain

Ilyabasharov:
> Also how many scheduling steps did you use here?

first 8k steps (figure taken at 40k)

kerrj:
@Ilyabasharov  you might want to try a lower number of scheduling steps to be less than the number of proposal annealing steps (which is 5k for nerfacto huge). If the coarse-to-fine scheduling lasts longer than annealing, high frequency details which were missed earlier in training can be impossible to recover since the rays are never sampled in those regions.

PR #2367, opened 2023-08-24
DEPTH_METRIC should be 1 when depth frames are available

padeler:
A small fix, allowing depth-nerfacto to work when depth data are available.


padeler:
Fixed by #2312

PR #2368, opened 2023-08-24
Fixes bug in PairPixelSampler when working with masked dataset

padeler:
PairPixelSampler does not compute the correct indices when there are masks in the dataset. This fixes the issue.


tancik:
Can you run `ns-dev-test`

elenacliu:
sry to bother you, but can you explain what the bug is?

padeler:
> sry to bother you, but can you explain what the bug is?

Basically its an indentation error in the else block. 

The  sample_method is supposed to select ray pairs by first randomly selecting half the rays and then picking the other half from a radius near the first ones. 

In the code when a mask is available it was selecting the indices for the first half but did not select their pairs.

AdamRashid96 (approved):
LGTM

PR #2369, opened 2023-08-24
fix pixel sampling with masks issue when data is list

anc2001:
PixelSampler sample_method expects mask argument to be batched but mask is unbatched when the images / masks are given as list. Batches the mask in this case.


anc2001:
For more context, when training on a dataset with images and masks of different sizes I get the following the following error 
```
  File "/src/external/nerfstudio/nerfstudio/data/pixel_samplers.py", line 222, in collate_image_dataset_batch_list
    all_images.append(batch["image"][i][indices[:, 1], indices[:, 2]])
IndexError: index 2 is out of bounds for dimension 1 with size 2
```
This is because the expected shape from `sample_method` is `[batch_size, 3]` but in this case the returned indices are of shape `[batch_size, 2]`. Batching the mask fixes this error

drewgilliam:
@brentyi & @tancik - hoping you could help here since you've merged some of my recent PRs.  Is there more information or more changes needed to get this PR merged?  Thanks!

brentyi:
Thanks for the ping @drewgilliam and the fix @anc2001; I just checked and this makes sense to me!

brentyi (approved):

drewgilliam:
@brentyi - thanks for the assist, much appreciated!

PR #2371, opened 2023-08-25
Fixes ARG before FROM bug

padeler:
ARG commands before FROM are not visible inside the build unit (after the FROM) unless specified again [without default values](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact).

Dockerfile will not build without this fix.


tancik (approved):
LGTM

PR #2384, opened 2023-08-30
Render nearest training view

AdamRashid96:
- Added logic to render the nearest training view to each camera in the rendered path. 
- The nearest training image is picked through a weighted sum of positional and rotational distance. Cameras that are occluded from the current rendered view are not selected unless there are no other training cameras unoccluded (for cases if the rendered view is in an object)
- Not sure whats the best way to inform people about this. Right now you can select this by adding `"nearest_camera"` to `rendered_output_names`

https://github.com/nerfstudio-project/nerfstudio/assets/71362382/a65c0747-b132-40d8-88a7-d9bb6be8df33


nerfstudio/scripts/render.py
```diff
@@ -47,11 +47,19 @@
 from torch import Tensor
 from typing_extensions import Annotated
 
+from scipy.spatial.transform import Rotation as R
```
ethanweber, line 50: Personally I'm not a fan of using `R` just in case someone uses that letter somewhere in the file. What do you think?

nerfstudio/scripts/render.py
```diff
@@ -129,6 +140,46 @@ def _render_trajectory_video(
...
+                if "nearest_camera" in rendered_output_names:
+                    cam_pos = cameras[camera_idx].camera_to_worlds[:, 3].cpu()
+                    cam_rot = R.from_matrix(cameras[camera_idx].camera_to_worlds[:3, :3].cpu())
+                    cam_quat = cam_rot.as_quat()
+
+                    max_dist, max_idx = None, None
+                    true_max_dist, true_max_idx = None, None
+                    for i in range(len(train_cameras)):
+                        train_cam_pos = train_cameras[i].camera_to_worlds[:, 3].cpu()
+                        # Make sure the line of sight from rendered cam to training cam is not blocked by any object
```
ethanweber, line 152: Is this necessary? I think it would be nice if the result is deterministic, regardless of how the NeRF is. Have you found it to be needed in some cases?

nerfstudio/scripts/render.py
```diff
@@ -129,6 +140,46 @@ def _render_trajectory_video(
...
+                            if true_max_dist is None or dist < true_max_dist:
+                                true_max_dist = dist
+                                true_max_idx = i
+                            continue
+
+                        if max_dist is None or dist < max_dist:
+                            max_dist = dist
+                            max_idx = i
+
+                    if max_idx is None:
```
ethanweber, line 181: I think this will never be done? Because of the lines above `if max_dist is None or dist < max_dist:`

nerfstudio/scripts/render.py
```diff
@@ -158,6 +211,26 @@ def _render_trajectory_video(
...
+
+                # Add closest training image to the right of the rendered image
+                if "nearest_camera" in rendered_output_names:
+                    img = train_dataset.get_image(max_idx)
+                    img = img.permute(2, 0, 1)
+                    img = transforms.ToPILImage()(img)
+                    resized_image = img.resize(
+                        (int(cameras.image_width[0]), int(cameras.image_height[0])), PIL.Image.Resampling.LANCZOS
+                    )
+                    resized_image = transforms.ToTensor()(resized_image).permute(1, 2, 0)
```
ethanweber, line 223: I think this could go down to two lines and you could avoid the extra imports.

```
img = train_dataset.get_image(max_idx)
img = torch.nn.functional.interpolate(
    img.permute(2, 0, 1)[None],
    size=(int(cameras.image_height[0]), int(cameras.image_width[0]))
)[0].permute(1, 2, 0)
```

ethanweber (commented):
Left a few comments. I think we could Tweet about this feature at some point, as the way to let people know about it.

nerfstudio/scripts/render.py
AdamRashid96: Agreed I got rid of the `as R`

nerfstudio/scripts/render.py
AdamRashid96: I added a flag `check-occlusions` so you can toggle whether or not you want the check

nerfstudio/scripts/render.py
AdamRashid96: Actually this will occur in cases where every camera is occluded because of the continue in the if statement above

nerfstudio/scripts/render.py
AdamRashid96: Yep this looks good. Got rid of the PIL import

nerfstudio/scripts/render.py
```diff
@@ -130,6 +139,49 @@ def _render_trajectory_video(
                     aabb_box = SceneBox(torch.stack([bounding_box_min, bounding_box_max]).to(pipeline.device))
                 camera_ray_bundle = cameras.generate_rays(camera_indices=camera_idx, aabb_box=aabb_box)
 
+                max_dist, max_idx = -1, -1
+                true_max_dist, true_max_idx = -1, -1
+
+                if "nearest_camera" in rendered_output_names:
```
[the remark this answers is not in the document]
AdamRashid96: Sounds good, I made it into a flag

nerfstudio/scripts/render.py
```diff
@@ -121,6 +126,10 @@ def _render_trajectory_video(
     with ExitStack() as stack:
         writer = None
 
+        assert pipeline.datamanager.train_dataset is not None
+        train_dataset = pipeline.datamanager.train_dataset
+        train_cameras = train_dataset.cameras.to(pipeline.device)
+
```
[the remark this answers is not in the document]
AdamRashid96: I moved this check inside the `if render_output_names` check so it shouldn't error if other pipelines that don't have train_datasets render.

nerfstudio/scripts/render.py
```diff
@@ -47,13 +47,18 @@
 from torch import Tensor
 from typing_extensions import Annotated
 
+from scipy.spatial.transform import Rotation
```
kerrj, line 50: for consistency might be nice to use the viser transform library

kerrj (approved):
nice

michaelrubloff:
> Left a few comments. I think we could Tweet about this feature at some point, as the way to let people know about it.

Added an article to help inform people! Let me know if I missed anything or if it needs to be corrected: https://neuralradiancefields.io/nerfstudio-introduces-nearest-camera-feature/

AdamRashid96:
> > Left a few comments. I think we could Tweet about this feature at some point, as the way to let people know about it.
> 
> Added an article to help inform people! Let me know if I missed anything or if it needs to be corrected: https://neuralradiancefields.io/nerfstudio-introduces-nearest-camera-feature/

This is awesome! One thing is that I changed it from adding `nearest_camera` to `rendered_output_names` instead to a flag `--render-nearest-camera`. If you could update the article that would be great.

michaelrubloff:
Updated!

PR #2397, opened 2023-09-05
Fix benchmarking: apply get_background_color in renderer and set profiler to none

Mxbonn:
Running `./nerfstudio/scripts/benchmarking/launch_train_blender.sh -m nerfacto` failed because of two errors:
`logging.enable-profiler` is does not exist and the `background_color` can be a string like `white`.

This pull request fixes both issues.
Note: the original code mentioned: `# This case must be before the others or the override is not properly applied`
I moved that case to become the final one, but as they are all still separated by an if - else clause, I don't think it breaks any assumptions.

A further comment related to benchmarking: Would it be possible to add benchmark results somewhere in this repo/documentation along with the last commit/stable release the benchmark was run on? This would let anyone use benchmark results without having to recompute them, as long as they are on the same code base.


Mxbonn:
Benchmark results:
image

nerfstudio/model_components/renderers.py
```diff
@@ -101,19 +101,15 @@ def combine_rgb(
         else:
             comp_rgb = torch.sum(weights * rgb, dim=-2)
             accumulated_weight = torch.sum(weights, dim=-2)
-        if BACKGROUND_COLOR_OVERRIDE is not None:
```
brentyi, line 104: Thanks @Mxbonn!

Can you double-check this logic? I think the override should be applied even when the background color is set to `random` or `last_sample`, which doesn't seem like it would happen if you remove this first `if` statement.

Adding the `else` at the end makes sense to me though.

Mxbonn: Hi @brentyi, you're right. It should indeed still be applied in case of an override.
However, the current way to do it seems a bit awkward to me since get_background_color serves both to override the background color and to convert it to RGB. But you cannot just simply pass background color `random` or `last_sample` because this will fail the assertion in case there is no override. So you need to do the check for override before calling the method. But if you're doing the override check anyway, then you should immediately apply it instead of calling the get_background_color.
Anyways I made a change to the code which should be correct now.

brentyi (approved):
Thanks @Mxbonn!!

PR #2398, opened 2023-09-06
Unify dataparser and method discovery

jkulhanek:


brentyi:
Consistency here makes sense to me. I guess we'd also want to update the docs?

jkulhanek:
I have updated the docs.

tancik (approved):
LGTM

PR #2400, opened 2023-09-06
test_train tmp_path

drewgilliam:
Use `tmp_path` [pytest fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-fixture) during `test_train` to direct the config `output_dir` to temporary pytest subdirectory. Also call to `config.set_timestamp()` to set timestamp on config training directory.

### DIFF
--- a/tests/test_train.py
+++ b/tests/test_train.py
@@ -29,7 +29,7 @@ BLACKLIST = [
 ]
 
 
-def set_reduced_config(config: TrainerConfig):
+def set_reduced_config(config: TrainerConfig, tmp_path):
     """Reducing the config settings to speedup test"""
     config.machine.device_type = "cpu"
     if hasattr(config.pipeline.model, "implementation"):
@@ -56,11 +56,15 @@ def set_reduced_config(config: TrainerConfig):
     # remove viewer
     config.viewer.quit_on_train_completion = True
 
+    # timestamp & output directory
+    config.set_timestamp()
+    config.output_dir = tmp_path / "outputs"
+
     return config
 
 
 @pytest.mark.filterwarnings("ignore::DeprecationWarning")
-def test_train():
+def test_train(tmp_path):
     """test run train script works properly"""
     all_config_names = method_configs.keys()
     for config_name in all_config_names:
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -81,7 +82,6 @@ dev = [
     "pre-commit==3.3.2",
     "pytest==7.1.2",
     "pytest-xdist==2.5.0",
-    "typeguard==2.13.3",
```
brentyi, line 84: > As `jaxtyping` depends on `typeguard` ([link](https://github.com/google/jaxtyping/blob/e347c480d51b2cf0671470b522ba09e38a0e9e25/pyproject.toml#L26)), move versioned `typeguard==2.13.3` from `dev` to base dependencies.

Do we use `typeguard` outside of development + automated tests? If not it makes sense to me to keep this pinned only in the `dev` dependencies.

tests/test_train.py
```diff
@@ -29,7 +29,7 @@
 ]
 
 
-def set_reduced_config(config: TrainerConfig):
+def set_reduced_config(config: TrainerConfig, tmp_path):
```
brentyi, line 32: can we add type annotations?

brentyi (commented):
looks overall reasonable to me!

pyproject.toml
drewgilliam: Some background - I ran into this issue trying to install dev while keeping base packages pinned to already installed versions.  If a user installs base nerfstudio, they get typeguard==4.3.1 (jaxtyping in the base requires typeguard>=2.13.3).  But if a user later installs dev nerfstudio, typeguard must be downgraded to 2.13.3 (unit tests don’t appear to pass with a newer version of typeguard). 

I found it unexpected for dev to introduce new version requirements for already installed base packages. Is that actually expected behavior?

tests/test_train.py
drewgilliam: Done

pyproject.toml
drewgilliam: okie dokie - I removed the `typeguard` change and updated the PR name and description to the limited scope.

brentyi (approved):

pyproject.toml
brentyi: Thanks!!

PR #2408, opened 2023-09-08
Allow colmap parser to load 3D points

jkulhanek:
This PR adds an option to the colmap parser, which allows it to load 3D points. These points can be used in methods like tetra-nerf or Gaussian splatting.

### DIFF
--- a/nerfstudio/data/dataparsers/colmap_dataparser.py
+++ b/nerfstudio/data/dataparsers/colmap_dataparser.py
@@ -71,6 +71,10 @@ class ColmapDataParserConfig(DataParserConfig):
     """Path to depth maps directory. If not set, depths are not loaded."""
     colmap_path: Path = Path("sparse/0")
     """Path to the colmap reconstruction directory relative to the data path."""
+    load_3D_points: bool = True
+    """Whether to load the 3D points from the colmap reconstruction."""
+    max_2D_matches_per_3D_point: int = -1
+    """Maximum number of 2D matches per 3D point. If set to -1, all 2D matches are loaded. If set to 0, no 2D matches are loaded."""
 
 
 class ColmapDataParser(DataParser):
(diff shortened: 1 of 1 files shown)


tancik (approved):
LGTM

maturk:
Thanks @jkulhanek, very useful.

maturk:
@jkulhanek, can you clarify what is the meaning of "points3D_num_points"? I think the naming might be a bit misleading, if my understanding is correct.

jkulhanek:
I agree it is misleading. The meaning is the number of 2D points (number of images) the 3D point is matched to. Should I rename it to something like `points3D_num_images`, or do you have any other suggestions?

maturk:
Thanks, yes. Is there a way to also get a full list of the image IDs that have the corresponding 2d points that correspond to the 3d point in question? From what I see, this p.image_ids just gives the number of point correspondences, which in itself is kind of useful (for example getting rid of poorly corresponding points) but it would be missing the exact image ids that these correspondences can be found in. I do not have much knowledge with colmap, so I am not sure if this is a feature or not. But I think renaming to points3D_num_2d_correspondences or similar would be very clear to me at least for now.

jkulhanek:
Shouldn't p.image_ids contain the actual image IDs, not their number?

maturk:
I am not entirely sure, I read this page: https://colmap.github.io/format.html and they have some information about a list of image_ids and 2d point indexes for a single 3D point. Here is something similar in pycolmap: https://github.com/colmap/colmap/blob/main/scripts/python/read_write_model.py#L327. Again, I do not have much experience with colmap, so I am uncertain what this really means. Let me know if you can clarify my doubts.

jkulhanek:
I believe `p.image_ids` is the list of image IDs, and `p.point2D_idxs` is the list of 2D point IDs within the images. This would be consistent with the pycolmap code you mentioned. What made you think `p.image_ids` was the number of 2D matches?

maturk:
Thanks, the naming just has me confused :D maybe we can rename everything to be consistent with pycolmap code then.

jkulhanek:
Currently, p.image_ids is `points3D_image_ids`; most names are also consistent. However, this is the new `points3D_num_images` field, which does not have a colmap equivalent. We can name it `point3D_num_point2D` or `points3D_num_2D_correspondences`? Finally, there is `points3D_image_xy` which corresponds to the `xys` property of COLMAP's image (but indexed by the point2D_idxs): https://github.com/colmap/colmap/blob/fd84a0d8f77940568aed13c368dbfdeb837b344d/scripts/python/read_write_model.py#L255C17-L255C20
An alternative name could, therefore, be `points3D_point2D_xy`?

PR #2418, opened 2023-09-12
gaussian splatting

jake-austin:
Gaussian splatting close to done in nerfstudio with minimal hacks to have it integrated, anyone else interested in pitching in help clean up some of the residual TODOs I’ve left around for myself and to get some of the bells and whistles like pruning / cloning done. 
I tried to focus on making sure this is clean and super easy to merge into main with good typing and minimal changes / still being consistent with all our existing classes and everything, but its still not perfect and in a day or so I’ll open it up as a PR and start taking some feedback. For now, I didnt create any new abstract classes, I just updated the datamanager to output Union[RayBundle, Camera] instead of RayBundle and that sort of thing (and I actually think its pretty clean this way and might keep it this way tbh).
As of writing this, this still uses the original gaussian splatting rasterizing just to make sure that all this infra works, but it should be a drop in replacement with our new homebrew splatting code. Same goes for a bunch of other places in the code that are currently transplants of their code for testing purposes only, with the intent to fully remove and replace those parts of the code with entirely novel code ASAP.

### DIFF
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -22,35 +22,44 @@ from collections import OrderedDict
 from typing import Dict
 
 import tyro
-from nerfstudio.data.pixel_samplers import PairPixelSamplerConfig
 
 from nerfstudio.cameras.camera_optimizers import CameraOptimizerConfig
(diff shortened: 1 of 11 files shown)


jake-austin:
Removing until we have a novel implementation

PR #2424, opened 2023-09-13
Reintroduce the mipnerf360 dataset (but using the HuggingFace dataset url)

Mxbonn:
With @ethanweber uploading the dataseto to https://huggingface.co/datasets/ethanweber/Mip-NeRF_360_Processed_with_Nerfstudio I would like to see this dataset return to the main branch.

Code borrowed from the https://github.com/nerfstudio-project/nerfstudio/tree/SIGGRAPH-2023-Code branch (+ small adjustments due to changes on master since this branch).

However, one thing to note is that I am currently unable to match the results of nerfacto on this dataset as reported in the siggraph paper.  I have modified the benchmark script and the config file to match what I believe is the setting used in the paper (see https://github.com/Mxbonn/nerfstudio_playground/) but the psnr is not close to the results in the paper. I can take this to a separate issue or PR.

### DIFF
--- a/nerfstudio/configs/dataparser_configs.py
+++ b/nerfstudio/configs/dataparser_configs.py
@@ -20,18 +20,25 @@ from typing import TYPE_CHECKING
 
 import tyro
 
-from nerfstudio.data.dataparsers.arkitscenes_dataparser import ARKitScenesDataParserConfig
+from nerfstudio.data.dataparsers.arkitscenes_dataparser import (
+    ARKitScenesDataParserConfig,
+)
 from nerfstudio.data.dataparsers.base_dataparser import DataParserConfig
 from nerfstudio.data.dataparsers.blender_dataparser import BlenderDataParserConfig
 from nerfstudio.data.dataparsers.colmap_dataparser import ColmapDataParserConfig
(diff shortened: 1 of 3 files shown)


jkulhanek:
Thank you for the PR. I was just wondering - is there a need for the mipnerf360 dataparser? The dataset (original, unprocessed) can be directly loaded by the COLMAP dataparser I believe.

Mxbonn:
> Thank you for the PR. I was just wondering - is there a need for the mipnerf360 dataparser? The dataset (original, unprocessed) can be directly loaded by the COLMAP dataparser I believe.

Good question, I just looked at what they did for the siggraph paper branch. I guess @ethanweber can answer this.

ethanweber:
Hi @Mxbonn, thanks for the request. We'd like to keep mipnerf360 data and parsing out of the main branch for maintainability purposes. We included it only for the SIGGRAPH paper to do benchmarking, but we don't wish to maintain this code. Rather, we'd encourage people to use the COLMAP parsing directly as suggested by @jkulhanek with maybe slight modifications to sample train/eval images in the same way. Thanks for understanding this decision!

PR #2425, opened 2023-09-13
Colmap parser: fix downscale for complex paths

jkulhanek:
This PR makes the following changes to the colmap dataparser:
- Optional paths are relative to config.data
- Fixes downscaling for depth/mask data
- Downscale respects deep image paths (previously it would fail with more then 1 level)


tancik (approved):
LGTM

PR #2429, opened 2023-09-14
Fix import error when tcnn installed but CUDA runtime not available

jkulhanek:
Also, this enables the ns-install-cli to work inside docker build context without requiring cuda-docker-runtime

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -168,6 +168,8 @@ RUN cd nerfstudio && \
 # Change working directory
 WORKDIR /workspace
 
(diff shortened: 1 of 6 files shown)


jkulhanek:
The change to the dockerfile also depends on fixing the import errors in lerf and nerfplayer:
https://github.com/kerrj/lerf/pull/48
https://github.com/lsongx/nerfplayer-nerfstudio/pull/9
Since these are included in the official image.

tancik (approved):
Much nicer

jkulhanek:
@tancik , please, do you know where I find out why the docs are failing?

Issue #2435, opened 2023-09-16
MPS not working properly


Jerry-Master:
**Describe the bug**
The training does not work on M1 Macbook Pro.

**To Reproduce**
1. Process data: ns-process-data images --data trial1/images/ --output-dir trial1/output/ --num-downscales 0 --no-gpu --matching-method exhaustive
2. Run training: ns-train nerfacto --data trial1/output/  --machine.device-type mps --mixed-precision False

**Expected behavior**
It gives this error:

```
Traceback (most recent call last):
  File "/Users/joseperezcano/Desktop/Tenerife/nerf teide/colmap/build/.venv/bin/ns-train", line 10, in <module>
    sys.exit(entrypoint())
  File "/Users/joseperezcano/Desktop/Tenerife/nerf teide/colmap/build/.venv/lib/python3.9/site-packages/nerfstudio/scripts/train.py", line 261, in entrypoint
    main(
  File "/Users/joseperezcano/Desktop/Tenerife/nerf teide/colmap/build/.venv/lib/python3.9/site-packages/nerfstudio/scripts/train.py", line 246, in main
    launch(
  File "/Users/joseperezcano/Desktop/Tenerife/nerf teide/colmap/build/.venv/lib/python3.9/site-packages/nerfstudio/scripts/train.py", line 189, in launch
    main_func(local_rank=0, world_size=world_size, config=config)
  File "/Users/joseperezcano/Desktop/Tenerife/nerf teide/colmap/build/.venv/lib/python3.9/site-packages/nerfstudio/scripts/train.py", line 100, in train_loop
    trainer.train()
  File "/Users/joseperezcano/Desktop/Tenerife/nerf teide/colmap/build/.venv/lib/python3.9/site-packages/nerfstudio/engine/trainer.py", line 255, in train
    loss, loss_dict, metrics_dict = self.train_iteration(step)
  File "/Users/joseperezcano/Desktop/Tenerife/nerf teide/colmap/build/.venv/lib/python3.9/site-packages/nerfstudio/utils/profiler.py", line 127, in inner
    out = func(*args, **kwargs)
  File "/Users/joseperezcano/Desktop/Tenerife/nerf teide/colmap/build/.venv/lib/python3.9/site-packages/nerfstudio/engine/trainer.py", line 467, in train_iteration
    with torch.autocast(device_type=cpu_or_cuda_str, enabled=self.mixed_precision):
[...]

PR #2436, opened 2023-09-16
Mps fix

Jerry-Master:
As expressed in #2435 there is a bug that make nerfstudio fail on apple silicon devices. I make an extra addition to fix it. However, I would like for nerfstudio to have more support on apple devices. I wanted to contribute to the docs on instructions about how to install on Mac and how to process the data to avoid errors but found that couldn't even build the docs locally. For instance, rawpy is not supported on M1 macs. And tyro installation fails. It installs, but later on it says that no module could be found. I also couldn't run tests locally since the developer branch is quite difficult to install on mac and I could not manage to do so. 

When you solve those issues, I will be more than glad to contribute to the docs. For the moment, I'll just leave the instructions to make nerfstudio work here.

First, install pymeshlab like this:

`pip3 install git+https://github.com/cnr-isti-vclab/PyMeshLab`

Then, install functorch with pip, and then install nerfstudio with pip too. That should suffice. When running the processing, I only managed to make it work with this command:

`ns-process-data images --data trial1/images/ --output-dir trial1/output/ --num-downscales 0 --no-gpu --matching-method exhaustive`

You can use images or video, but be sure to use `--no-gpu` and the exhaustive matching type. It should suffice for short sequences. Then, to train the model, run:

`PYTORCH_ENABLE_MPS_FALLBACK=1 ns-train nerfacto --data trial1/output/  --machine.device-type mps --mixed-precision False`

For 100 frames the estimated running time on my device (M1 Max) is 12 hours approximately.


tancik:
Run `ns-dev-test` to fix test errors.

tancik (approved):

PR #2440, opened 2023-09-18
Reimplementing sdfstudio into nerfstudio (Neuralangelo)

AdamRashid96:
- Going to move over and refactor all of the sdfstudio codebase
- Might be good to incrementally merge the code into main
- Starting with sdfstudio's implementation of Neuralangelo
- About 10ms per step lower (Not sure why yet)
- Still want to refactor the model and field

### DIFF
--- a/nerfstudio/data/dataparsers/sdfstudio_dataparser.py
+++ b/nerfstudio/data/dataparsers/sdfstudio_dataparser.py
@@ -33,6 +33,7 @@ from nerfstudio.data.dataparsers.base_dataparser import (
     DataparserOutputs,
 )
 from nerfstudio.data.scene_box import SceneBox
+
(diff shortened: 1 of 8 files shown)


nerfstudio/data/dataparsers/sdfstudio_dataparser.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 the Regents of the University of California, Nerfstudio Team and contributors. All rights reserved.
```
tancik, line 1: Avoid changing the license header

jkulhanek:
How do the methods compare (sdfstudio vs this PR)?

hummat:
Hi, I was planning to do the same thing. Any news here? Would be happy to help out.

Issue #2439, opened 2023-09-18
render image size not equal to train image size


Saixiaoma:
render image size not equal to train image size

maturk:
You can change your desired rendered image width and height and resolution settings in the ns-render or ns-viewer settings.

Saixiaoma:
where is the resolution setting in ns-render？I cannot find in render.py

maturk:
So first you define the render width and height in ns-viewer or ns-train when you make your desired camera path. 
<img width="406" alt="image" src="https://github.com/nerfstudio-project/nerfstudio/assets/30566358/e5db9fef-9276-478a-b062-353b231caef5">
 If you want to also change this after making a camera path, you can open the camera_paths.json file and change the vars "render_width" and "render_height" there before running ns-render command.

maturk:
Are you able to port forward from your server? If you run this command in your local computer `ssh -L 7007:localhost:7007 USER@YOUR_IP`

Saixiaoma:
No, I cannot. permission denied.

wanger-666:
How to change image size when use interpolated render?+1

Issue #2433, opened 2023-09-15
I am running the program on a Ubunut 20.04 server, how can I visualise the training process on my local win11 system?


1gjjuser1:
Firstly the programme works fine on the Ubunut system.
![image](https://github.com/nerfstudio-project/nerfstudio/assets/95596231/eaed4827-6107-488f-b977-a744b6b4df1f)
Then I run the command: ssh -L 6666:localhost:6666 aii@192.168.0.251 (192.168.0.251 is the IP address of the Ubuntu system) in the local cmd window.

After that, I open the url http://localhost:6666/,But the page shows blank, what's going on, I hope to get your help, thanks!

jlightfoot:
Replace localhost in the ssh command with the Ubuntu box's IP.

1gjjuser1:
Firstly, thank you for your prompt reply!
According to you, I used the ssl -L command in the cmd command line on my local win11 system, and after typing it, why I got to the remote Ubuntu system. Does this mean that port forwarding was done successfully?
![image](https://github.com/nerfstudio-project/nerfstudio/assets/95596231/7eda9219-6c24-452d-a2b7-29f080c9e721)
After that I entered the training command ns-train nerfacto --data data/nerfstudio/poster in the cmd command line on win11 and the programme ran successfully but clicking on the url doesn't allow me to view it locally, what's going on?
![image](https://github.com/nerfstudio-project/nerfstudio/assets/95596231/2d3fa00b-3c24-40c1-b35f-b35781b313bc)

Issue #1908, opened 2023-05-11
Can I only install ns-viewer ? 


pansanity666:
Hi, 
I want to download the ckpt from the remote machine to my laptop, and then use ns-viewer to visualize the ckpt. (For some reason, ssh -L is not work for me.) The question is can I only install ns-viewer in my laptop (Macbook M1) instead of the whole nerfstudio?
Best,

brentyi:
The answer is basically no:
- `ns-viewer` depends on having the whole `nerfstudio` package installed
- `ns-viewer` still depends on CUDA for all of the (practical) models, so unfortunately you won't be able to run it directly on a Macbook

pansanity666:
> The answer is basically no:
> 
> * `ns-viewer` depends on having the whole `nerfstudio` package installed
> * `ns-viewer` still depends on CUDA for all of the (practical) models, so unfortunately you won't be able to run it directly on a Macbook

I found a workaround that is to use IFrame in jupyterlab in the remote machine. However, It shows  renderer disconnected:
![image](https://github.com/nerfstudio-project/nerfstudio/assets/47111102/1c85ea36-fc26-4ebd-9e8b-f70392887d07)

Do you have any idea about it ?

brentyi:
Not really, sorry! Just to confirm are you port forward with `-L`? If you're using Safari, it's also worth trying Chrome.

pansanity666:
> Not really, sorry! Just to confirm are you port forward with `-L`? If you're using Safari, it's also worth trying Chrome.

Do you mean ```ssh -L <port>:localhost:<port> USER@REMOTE.SERVER.IP```?  I actually dont know the remote server ip for some reason.  I have decided to use a more naive solution, ie, to save the rendered image each iteration. LOL

brentyi:
Doesn't seem ideal, but glad you found a solution. 😄 

Just wondering, how are you running the `ns-train` command if you don't have the remote IP? Usually what one would do is SSH in using that command to run the training script.

pansanity666:
> Doesn't seem ideal, but glad you found a solution. 😄
> 
> Just wondering, how are you running the `ns-train` command if you don't have the remote IP? Usually what one would do is SSH in using that command to run the training script.

Actually, the ip address does exist, however, it is not revealed to us users considering the safety issue.  I can only use the remote machine via a specified web page.

PR #2445, opened 2023-09-19
Add share link support for beta viewer

brentyi:
This PR adds initial support for shareable links to the beta viewer.

Link generation can be enabled via `--share True` on either `ns-train` or `ns-viewer`. For example,
```
ns-train nerfacto --data ./data/nerfstudio/person --vis viewer_beta --share True
```
wil print:
image

Which, thanks @jonahbedouch's efforts on mobile support, we can open on a phone!

An example URL is https://edge-based-relational-7462.share.viser.studio, which I'll keep up for a few hours if anybody wants to check that they can access it.

Once the final beta viewer features are stable, this can also help with Colab tunnel issues + various port forward-related roadblocks that folks seem to be having (#1908 #2433 #2439).

### DIFF
--- a/nerfstudio/configs/base_config.py
+++ b/nerfstudio/configs/base_config.py
@@ -146,3 +146,5 @@ class ViewerConfig(PrintableConfig):
     """Image format viewer should use; jpeg is lossy compression, while png is lossless."""
     jpeg_quality: int = 90
     """Quality tradeoff to use for jpeg compression."""
+    make_share_url: bool = False
+    """Viewer beta feature: print a shareable URL. `vis` must be set to viewer_beta; this flag is otherwise ignored."""
--- a/nerfstudio/configs/experiment_config.py
+++ b/nerfstudio/configs/experiment_config.py
@@ -22,7 +22,13 @@ from pathlib import Path
 from typing import Any, Dict, Literal, Optional
 
 import yaml
-from nerfstudio.configs.base_config import InstantiateConfig, LoggingConfig, MachineConfig, ViewerConfig
+
+from nerfstudio.configs.base_config import (
+    InstantiateConfig,
+    LoggingConfig,
+    MachineConfig,
+    ViewerConfig,
+)
 from nerfstudio.configs.config_utils import to_immutable_dict
 from nerfstudio.engine.optimizers import OptimizerConfig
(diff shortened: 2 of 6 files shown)


tancik:
This is very cool! Could we add the share link to the rich table instead of as new line?

brentyi:
This is what I had originally; the reason I switched it is that the share link can take a second or two to generate. I can flip it back in `viser`, but if we choose to include it in the table:
- The table printing will be delayed by a bit.
- We would need to decide whether to block until the share URL is ready or return immediately, then do the print later from a background thread. The latter is obviously preferred for saving time, but I'm concerned about how printing a lot of stuff interacts with things like `tqdm` progress bars (probably it's OK).

tancik:
Eventually I kinda want to make the share option opt-out and have it front and center in the table. I wasn't aware that it had a lag, though a couple of seconds is pretty short. Thoughts?

brentyi:
Okay, let's include it in the table! I can add it to the viser to-do list.

I just measured the delay and it's much faster now, ~0.15 seconds, so this is possibly a non-issue. It probably helps that I'm on Ethernet now.

kerrj:
IMO we should rename "share" in the config to something like "public-viewer" or something more descriptive

brentyi:
I'm okay with something like `--create-share-url True`?

kerrj (approved):
nice

smart4654154:
Thank you very much for your work. I use: C:\Users\zzl\erfstudio\outputs\kitchen\erfacto\2024-01-25_131124\config.yml --vis viewer_ Beta 
I click http://0.0.0.0:7007/, and browser display can't reach.

image


When I use ns viewer -- load configuration C:\Users\zzl\erfstudio\outputs\kitchen\erfacto\2024-01-25_131124\config.yml --vis viewer_ Beta  --viewer.make-share-url True. I can obtain new connections and open it
image

Does this mean that --viewer.make-share-url is an effective solution to my problem?
(I am a beginner and English is poor, and my question may be a simple one.Thank you very much for your help.)

jonahbedouch:
Hi! It's currently a known bug that some browsers cannot recognize the 0.0.0.0 url. You can try manually typing in https://localhost: followed by the port that viser spits out into your browser, or you can try to utilize the shared link feature if that doesn't work.

Issue #1606, opened 2023-03-15
Getting scaled exports


Baptiste-AIST:
**Describe the bug**
Hi everyone, I'm not too sure if it's a bug, but I wanted to know how the generated point cloud or meshes are scaled when exporting.
I tried to compare the sizes from instant-ngp, nerfstudio ans COLMAP and they differ between each other (using the same transforms.json).
Is there a place when training or exporting where things get scaled?

Thank you very much!

**Expected behavior**
At least I was expecting nerfstudio to generate a similar scaling as COLMAP or instant-ngp

**Screenshots**
![image](https://user-images.githubusercontent.com/105469463/225216261-ce5a5c61-610f-438f-b5f3-70f18947c4a1.png)

tancik:
The dataprocessing scales and rotates the data. The nerf coordinate frame can be determined using the dataparser_transform.json in the output folder. This can be used to get to the coordinate system of the transforms.json file.

Baptiste-AIST:
I guess you're talking about this function
[colmap_to_json()](https://github.com/nerfstudio-project/nerfstudio/blob/6db1d314cc646598daf0093b3a0513bbb43b718f/nerfstudio/process_data/colmap_utils.py#L388) in colmap_utils.py
But I don't see any scaling there, am I missing something somewhere else?

Baptiste-AIST:
Okay, I found where the scaling happens
https://github.com/nerfstudio-project/nerfstudio/blob/2b8d5a9e5c2dba4332a62434355631c13f6d9c96/nerfstudio/data/dataparsers/nerfstudio_dataparser.py#L70

The centering and scaling are done in the nerfstudio dataparser, contrarily to instant-ngp where it is done in colmap2nerf.py script and are contained in directly in the transforms.json.

It is possible to use the save_dataparser_transform to save the transformation matrix used to center the cameras and the scaling.
I'm not sure where it is possible to call this function, it could be great to have another argument in the command line to generate this file.

 https://github.com/nerfstudio-project/nerfstudio/blob/c0464f8a9113636f877841302a0e48613cb662f8/nerfstudio/data/dataparsers/base_dataparser.py#L77

cdcseacave:
is there a way to export the point-cloud and/or mesh scaled/transformed back to the original coordinates?

LHXhh:
[...]

PR #2447, opened 2023-09-19
Render oriented crops in viewer beta

kerrj:
Adds support for for rendering with oriented boxes in the new viewer.

image

https://github.com/nerfstudio-project/nerfstudio/assets/15806475/7e35c4e4-e433-44c1-b10e-8b5cea752b52

### DIFF
--- a/nerfstudio/viewer_beta/render_panel.py
+++ b/nerfstudio/viewer_beta/render_panel.py
@@ -258,8 +258,11 @@ class CameraPath:
         )
 
 
-def populate_render_tab(server: viser.ViserServer,config_path: Path, datapath: Path, control_panel: ControlPanel) -> None:
+def populate_render_tab(
+    server: viser.ViserServer, config_path: Path, datapath: Path, control_panel: ControlPanel
+) -> None:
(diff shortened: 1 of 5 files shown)


nerfstudio/viewer_beta/render_panel.py
```diff
@@ -539,8 +544,100 @@ def _(_) -> None:
     def _(event: viser.GuiEvent) -> None:
         """TODO: write the render JSON and show the render command."""
```
brentyi, line 545: Let's remove the TODO?

brentyi (commented):
lgtm, let's fix the formatting?

brentyi (approved):

smart4654154:
Thank you very much for your work. My question seems to be related to this job. What is the meaning of the crop scale value in the new version of viewer? I found that the value of the crop scale cannot match the coordinate values of the world coordinate system. For example, when the crop scale is set to 4, 4, 4, it is approximately equivalent to a square with a side length of 40 in the world coordinate system
image
What are the meanings of "crash max" and "crash min" in Viewer_legacy?
image

brentyi:
let's move discussion to #1606!

PR #2450, opened 2023-09-20
[Open for discussion] Rework/remove `density_fn` to support Gaussians.

Mxbonn:
### issue
The `density_fn` in `class Field` takes different arguments than `get_density` (it takes positions instead of RaySamples).
As a result, it is not possible to use the ProposalNetworkSampler with networks that do not take positions as arguments. One example is: MipNerf360, which uses Gaussians in the proposal network.
More generally, I don't think it makes sense to have a function that is nearly identical to `get_density`, but that takes in non-standard arguments.

### Proposed fix
deprecate `density_fn` it can nearly always be replaced with `get_density`. In places where it is not straightforward to replace it with `get_density` (e.g. the occupancy grid) use a custom wrapper function.

#### Problem with the fix
It is a breaking change, but a breaking change will be necessary if you want to support e.g. mipnerf360.
I don't know the policy for breaking changes, but one could always not immediately remove the `density_fn` and put a deprecation warning on it.

### DIFF
--- a/nerfstudio/cameras/camera_utils.py
+++ b/nerfstudio/cameras/camera_utils.py
@@ -25,6 +25,8 @@ from jaxtyping import Float
 from numpy.typing import NDArray
 from torch import Tensor
 
+from nerfstudio.utils import torch_compile
+
 _EPS = np.finfo(float).eps * 4.0
 
 
@@ -405,7 +407,7 @@ def _compute_residual_and_jacobian(
     return fx, fy, fx_x, fx_y, fy_x, fy_y
 
 
-# @torch_compile(dynamic=True, mode="reduce-overhead", backend="eager")
+@torch_compile(dynamic=True, mode="reduce-overhead", backend="eager")
 def radial_and_tangential_undistort(
     coords: torch.Tensor,
     distortion_params: torch.Tensor,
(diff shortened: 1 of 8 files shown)


Ilyabasharov:
@Mxbonn Fully agree! #2242

brentyi:
Does anybody have a sense for how many (public) external repositories would be impacted by this?

jkulhanek:
I believe external repos should fix an NS version. The development breaks lots of stuff between versions so I think its fine. Perhaps keeping a deprecation warning for a couple of versions would be nice.

jkulhanek:
Do you agree to merge this as is, or should we add some deprecation warnings?

brentyi:
> I believe external repos should fix an NS version.

This seems like a reasonable thing to recommend!

> Do you agree to merge this as is, or should we add some deprecation warnings?

I skimmed a few repos, and at the very least it seems like the K-Planes extension will break from this PR:
https://github.com/Giodiro/kplanes_nerfstudio/blob/cc5867b14697732bef4dd1de5ba49fafd419b312/kplanes/kplanes.py#L189-L214

That said, I'm okay with merging without deprecation warnings. If we do this we can follow up with an official recommendation for external methods to pin a nerfstudio version/commit in the docs, and perhaps make PRs for fixing versions of external methods that we know are out there? Does this make sense? cc also @kerrj @tancik

Mxbonn:
So anyone willing to make a decision on this?

jkulhanek:
@brentyi , do you agree with merging?

jkulhanek:
I agree. I will add the error message soon.

PR #2455, opened 2023-09-22
Add `background_color` config to keep consistent with other models

kobejean:
Models like `nerfacto` and `instant-ngp` have the background_color property configurable. Other models like `generfacto`, `tensorf`, `mipnerf` and `vanilla-nerf` have the property fixed to white. I'd like to be able to configure these models in the same way as  in `nerfacto` and `instant-ngp` for running some experiments. I've set the default value to white so as not to change existing behavior.


tancik (approved):
LGTM

kobejean:
Just fixed the linting errors

PR #2457, opened 2023-09-22
Do not materialize mlp_head if no field_heads

Mxbonn:
In case it is used in a proposal network we avoid initializing the head.


tancik (approved):
LGMT

PR #2462, opened 2023-09-23
Change how random background is blended.

Mxbonn:
I don't believe random backgrounds were applied properly.
This PR adds a random background to the predicted rgb during training.
This matches also how it is done in multinerf: https://github.com/google-research/multinerf/blob/main/internal/models.py#L241-L281.

### DIFF
--- a/nerfstudio/model_components/renderers.py
+++ b/nerfstudio/model_components/renderers.py
@@ -66,9 +66,8 @@ class RGBRenderer(nn.Module):
         super().__init__()
         self.background_color: BackgroundColor = background_color
 
-    @classmethod
     def combine_rgb(
-        cls,
+        self,
         rgb: Float[Tensor, "*bs num_samples 3"],
         weights: Float[Tensor, "*bs num_samples 1"],
         background_color: BackgroundColor = "random",
@@ -76,8 +75,6 @@ class RGBRenderer(nn.Module):
         num_rays: Optional[int] = None,
     ) -> Float[Tensor, "*bs 3"]:
         """Composite samples along ray and render color image.
-        If background color is random, no BG color is added - as if the background was black!
-
         Args:
             rgb: RGB for each sample
             weights: Weights for each sample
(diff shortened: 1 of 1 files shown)


nerfstudio/model_components/renderers.py
```diff
@@ -104,13 +101,14 @@ def combine_rgb(
         if BACKGROUND_COLOR_OVERRIDE is not None:
             background_color = BACKGROUND_COLOR_OVERRIDE
         if background_color == "random":
-            # If background color is random, the predicted color is returned without blending,
-            # as if the background color was black.
-            return comp_rgb
+            if self.training:
```
jkulhanek, line 104: This breaks it, I believe. Currently, the blending is done later in `blend_background`. The reason is that we want to use the image with a random background for computing the loss fn, but then blend the image with the correct background for visualization. Everything without computing the accumulated colours twice. If this change is accepted, the computed colour will (for most models) be blended twice. 
Have you tried doing the change to the code and trying to train nerfacto with random bg on some data?

nerfstudio/model_components/renderers.py
```diff
@@ -188,11 +186,8 @@ def blend_background_for_loss_computation(
             A tuple of the predicted and ground truth RGB values.
         """
         background_color = self.background_color
-        if background_color == "last_sample":
-            background_color = "black"  # No background blending for GT
-        elif background_color == "random":
-            background_color = torch.rand_like(pred_image)
-            pred_image = pred_image + background_color * (1.0 - pred_accumulation)
```
jkulhanek, line 195: Why did you remove blending with random bg here?

nerfstudio/model_components/renderers.py
```diff
@@ -66,18 +66,15 @@ def __init__(self, background_color: BackgroundColor = "random") -> None:
         super().__init__()
         self.background_color: BackgroundColor = background_color
 
-    @classmethod
```
jkulhanek, line 69: Changing this to member method will break some methods

jkulhanek (changes requested):
Hi, sorry, but I believe this PR would break most of the current methods. Is there a particular reason why you need this change? What does it resolve?

Mxbonn:
> Hi, sorry, but I believe this PR would break most of the current methods. Is there a particular reason why you need this change? What does it resolve?

I think I got a bit confused with how the background sampling is supposed to work with random sampling but shouldn't it blend with [0.5,0.5,0.5] during testing instead of not blending anything with it?

jkulhanek:
I think for visualisation, any colour is ok. For some scenes, black is preferable for some white. Blending with 0.5 is also ok. Not sure which is better. We could perhaps allow this to be a parameter.

jkulhanek:
I don't think the colours matter much when computing test PSNR as the density will be placed on the random bg if it does not match gt and if there is a mask, the blended colours are the same.

Mxbonn:
Yes, I think I got carried away in the details of a paper I'm reproducing a bit too much. The current implementation has no noticeable difference with my implementation so I'll close the PR.

jkulhanek:
Do you want to import mipnerf360 into the repo?

Mxbonn:
> Do you want to import mipnerf360 into the repo?

Yes. Still working on their interval sampler with dilation but even without that, I am already matching the performance and loss of the multinerf implementation.

jkulhanek:
That is great news! Let me know if I can help.

PR #2464, opened 2023-09-24
Allow custom sampler in ProposalNetworkSampler

Mxbonn:
This allows custom `PDFSamplers` (e.g. mipnerf360 samples intervals instead of points) in distinct repositories or future versions while still using the base `ProposalNetworkSampler`.

### DIFF
--- a/nerfstudio/model_components/ray_samplers.py
+++ b/nerfstudio/model_components/ray_samplers.py
@@ -540,6 +540,7 @@ class ProposalNetworkSampler(Sampler):
         single_jitter: bool = False,
         update_sched: Callable = lambda x: 1,
         initial_sampler: Optional[Sampler] = None,
+        sampler: Optional[Sampler] = None,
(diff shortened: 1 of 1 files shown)


nerfstudio/model_components/ray_samplers.py
```diff
@@ -540,6 +540,7 @@ def __init__(
         single_jitter: bool = False,
         update_sched: Callable = lambda x: 1,
         initial_sampler: Optional[Sampler] = None,
+        pdf_sampler: Optional[PDFSampler] = None,
```
jkulhanek, line 543: Please also add this to "Args" in docstring.

Mxbonn: Done :).

jkulhanek (approved):

jkulhanek:
Thanks for adding the option.

PR #2465, opened 2023-09-24
Make resetting the near_plane in NearFarCollider optional.

Mxbonn:
It is not always desired to reset the near plane to 0 during inference. Added an optional flag.


jkulhanek:
I'm curious why it was there in the first place. Could it be removed for good? @tancik

tancik:
I think it was there because otherwise the scene can get clipped when viewing it in the viewer. iirc this was an issue with the blender synthetic datasets awhile ago.

jkulhanek:
I wonder if we can remove this for good...

kerrj:
It seems like this would only matter if the camera gets super close to objects in the scene, how often do people do this? What situations would you want to keep the near plane the same as training?

Mxbonn:
> situations

Well some methods do not even work when the near plane is zero. E.g. if the sampler is `LinearDisparitySampler` due to the 1/near_plane. This could also be made to work if you replace 0 with a small epsilon but I thought keeping the initial `near_plane` much more logical.

jkulhanek:
> It seems like this would only matter if the camera gets super close to objects in the scene, how often do people do this? What situations would you want to keep the near plane the same as training?

I think changing the near plane skews the sampling in nerf/mipnerf potentially decreasing the eval perf.

jkulhanek:
I would vote for removing resetting to zero. I don’t think having the option here is useful.

jkulhanek:
@brentyi, do you accept this PR?

brentyi:
It looks good to me, I just tweaked the comment to add details on why we reset the clipping plane.

I don't feel strongly, but I'd vote to keep the argument + current reset behavior. I agree that it'd likely hurt eval metrics when evaluation images are a similar distance to the object as the training images (as is the case in the synthetic datasets), but it would dramatically improve them if the evaluation images get much closer (not present in evaluation datasets, but IMO more useful in the real world).

brentyi (approved):

PR #2473, opened 2023-09-27
Update colmap installation instructions with conda

ginazhouhuiwu:
Deprecate `apt` with `conda` for colmap installation to ensure correct version 3.8 with CUDA support.


brentyi (approved):
Thanks @ginazhouhuiwu!

PR #2477, opened 2023-09-28
fix:correction of PairPixelSampler in case of using masks with depth-nerfacto

pierremerriaux-leddartech:
Hi,
When I try to train depth-nerfacto with mask, assert below pop:
File "/home/pierre.merriaux/project/nerfstudio/nerfstudio/data/pixel_samplers.py", line 176, in collate_image_dataset_batch
assert collated_batch["image"].shape[0] == num_rays_per_batch

It is because in PairPixelSampler in case of using masks, indices are returned with shape self.rays_to_sample and self.rays_to_sample = self.config.num_rays_per_batch // 2.
I added pair_indices stacking as without using masks. I will push the PR soon


AdamRashid96 (approved):
This LGTM

pierremerriaux-leddartech:
Hi @AdamRashid96 , could we merge this one?
Thanks

PR #2490, opened 2023-10-05
Add documentation about unknown depth pixel convention

jampekka:
Updated the documentation to explicate that zero value in depth images is treated as unknown depth. This behavior is defined in https://github.com/nerfstudio-project/nerfstudio/blob/a484d255b4f71c55915afcfc52d90ec88963779f/nerfstudio/model_components/losses.py#L242


tancik (approved):
LGTM

PR #2498, opened 2023-10-09
Add scannetpp dataparser

liu115:
Dataparser for the [ScanNet++](https://kaldir.vc.in.tum.de/scannetpp/) dataset adapted from the nerfstudio dataset. The dataset parser follows the dataset file structure and can be used by specifying the dataset root and the scene ID.

Example usage
```shell
DATA_ROOT=SCANNETPP_DATASET_ROOT
SCENE_ID=036bce3393

ns-train nerfacto \
     --max_num_iterations 100000 \
     --pipeline.datamanager.train_num_rays_per_batch 8192 \
     --pipeline.datamanager.train_num_images_to_sample_from 400 \
     --pipeline.datamanager.train_num_times_to_repeat_images 100 \
     --vis viewer+tensorboard \
     scannetpp-data \
     --data ${DATA_ROOT} \
     --scene-id ${SCENE_ID} \
     --scene_scale 1.5
```
The scene contain more than 1k images. Set `train_num_images_to_sample_from` and `train_num_times_to_repeat_images` to avoid cpu out-of-memory.

Training with the viewer. The images with red borders are the predefined testing images by the dataset
image

The result rendering
https://github.com/nerfstudio-project/nerfstudio/assets/8998128/e4c01178-cafa-4e28-837c-6b0bdf5538d5

### DIFF
--- a/nerfstudio/data/dataparsers/scannetpp_dataparser.py
+++ b/nerfstudio/data/dataparsers/scannetpp_dataparser.py
@@ -22,11 +22,8 @@ from typing import Literal, Type
 import numpy as np
 import torch
 from nerfstudio.cameras import camera_utils
-from nerfstudio.cameras.cameras import (CAMERA_MODEL_TO_TYPE, Cameras,
-                                        CameraType)
-from nerfstudio.data.dataparsers.base_dataparser import (DataParser,
-                                                         DataParserConfig,
-                                                         DataparserOutputs)
+from nerfstudio.cameras.cameras import CAMERA_MODEL_TO_TYPE, Cameras, CameraType
+from nerfstudio.data.dataparsers.base_dataparser import DataParser, DataParserConfig, DataparserOutputs
 from nerfstudio.data.scene_box import SceneBox
 from nerfstudio.utils.io import load_from_json
 from nerfstudio.utils.rich_utils import CONSOLE
--- a/nerfstudio/data/dataparsers/scannetpp_dataparser.py
+++ b/nerfstudio/data/dataparsers/scannetpp_dataparser.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Literal, Optional, Type
+from typing import Literal, Type
 
 import numpy as np
 import torch
--- a/nerfstudio/data/dataparsers/scannetpp_dataparser.py
+++ b/nerfstudio/data/dataparsers/scannetpp_dataparser.py
@@ -22,11 +22,8 @@ from typing import Literal, Optional, Type
 import numpy as np
 import torch
 from nerfstudio.cameras import camera_utils
-from nerfstudio.cameras.cameras import (CAMERA_MODEL_TO_TYPE, Cameras,
-                                        CameraType)
-from nerfstudio.data.dataparsers.base_dataparser import (DataParser,
-                                                         DataParserConfig,
-                                                         DataparserOutputs)
+from nerfstudio.cameras.cameras import CAMERA_MODEL_TO_TYPE, Cameras, CameraType
+from nerfstudio.data.dataparsers.base_dataparser import DataParser, DataParserConfig, DataparserOutputs
 from nerfstudio.data.scene_box import SceneBox
 from nerfstudio.utils.io import load_from_json
 from nerfstudio.utils.rich_utils import CONSOLE
--- a/nerfstudio/data/dataparsers/scannetpp_dataparser.py
+++ b/nerfstudio/data/dataparsers/scannetpp_dataparser.py
@@ -1,0 +1,244 @@
+# Copyright 2022 the Regents of the University of California, Nerfstudio Team and contributors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+""" Data parser for ScanNet++ datasets. """
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Literal, Optional, Type
+
+import numpy as np
+import torch
+from nerfstudio.cameras import camera_utils
+from nerfstudio.cameras.cameras import (CAMERA_MODEL_TO_TYPE, Cameras,
+                                        CameraType)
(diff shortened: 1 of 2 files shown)


liu115:
I also noticed that training is much slower with mask (1-2s per iteration for default nerfacto) because the pixel sampler is running nonzero on CPU, especially when the number of images and the image resolution are large.

https://github.com/nerfstudio-project/nerfstudio/blob/81db1808a72b798d0e314d5bdb6abb25bd9c71c8/nerfstudio/data/pixel_samplers.py#L96-L99

One way is to have less image per samples and more frequent sampling like
```
--pipeline.datamanager.train_num_images_to_sample_from 8 \
--pipeline.datamanager.train_num_times_to_repeat_images 1 \
```

Or how the mask being used in the code has to be changed. Since the invalid pixels are usually a small part of the pixels among all the images, we could save all the `mask == 0` pixels as preprocessing and sample pixels for training from them. Or ignore the mask during sampling and only apply mask filtering during loss computation.

nerfstudio/data/dataparsers/scannetpp_dataparser.py
```diff
@@ -0,0 +1,241 @@
...
+from nerfstudio.cameras.cameras import CAMERA_MODEL_TO_TYPE, Cameras, CameraType
+from nerfstudio.data.dataparsers.base_dataparser import DataParser, DataParserConfig, DataparserOutputs
+from nerfstudio.data.scene_box import SceneBox
+from nerfstudio.utils.io import load_from_json
+from nerfstudio.utils.rich_utils import CONSOLE
+
+
+@dataclass
+class ScanNetppDataParserConfig(DataParserConfig):
+    """ScanNet++ dataset config."""
```
tancik, line 34: Can you add a link to dataset along with a description.

tancik (approved):
LGTM, thanks for adding

OrangeSodahub:
@liu115 Hi, I wonder you used nerfacto or nerfacto-huge, and each iteration costs 1-2s?

liu115:
I used nerfacto

One quick solution to avoid the mask sampling bottleneck is not to use mask during pixel sampling. Instead, apply the mask only during the loss computation

For example, in `nerfacto.py`
```python

        loss_dict = {}
        pred_rgb, gt_rgb = self.renderer_rgb.blend_background_for_loss_computation(
            pred_image=outputs["rgb"],
            pred_accumulation=outputs["accumulation"],
            gt_image=image,
        )
        if "mask" in batch:
            mask = batch["mask"].view(-1)
            assert mask.shape[0] == gt_rgb.shape[0]
            loss_dict["rgb_loss"] = self.rgb_loss(gt_rgb[mask], pred_rgb[mask])
        else:
            loss_dict["rgb_loss"] = self.rgb_loss(gt_rgb, pred_rgb)
```

and remove the if statement here
https://github.com/nerfstudio-project/nerfstudio/blob/7edd7682406a03f94e20e18af6534887ca473a1e/nerfstudio/data/pixel_samplers.py#L98-L101

OrangeSodahub:
@liu115 Hi, I have another thing want to confirm with you. That is, there exist two ways that I could get an extrinsic matrix of a camera. One is from the `colmap/images.txt` file and use the `qvec2rotmat` method you provided, and in this case the matrix is in OpenCV format right? Another way is from the `transforms.json`, which is in OpenGL format, is that right?

liu115:
Yes. `transforms.json` follows nerfstudio/OpenGL format, and the colmap files follows the format defined by colmap.

OrangeSodahub:
@liu115 But I found that the translation vectors extracted from the extrinsic matrix from `transform.json` and from colmap files are different? Specifically, mat_A is from `transform.json`, mat_B is from colmap file, and
```
mat_A[..., :3, 3] != mat_B[..., :3, 3]
```

liu115:
Reference here when `keep_original_world_coordinate` is false. 
https://github.com/nerfstudio-project/nerfstudio/blob/3a4ed804d3abe1565ccd521247a2724b68ee28e2/nerfstudio/process_data/colmap_utils.py#L434-L441

OrangeSodahub:
@liu115 Thanks for your explaination. But I'v checked the difference between to version of extrinsic matrix, for example, for scene `56a0ec536c` and image_id = `DSC03534`:

```
# extrinsic matrix obtained from colmap/images.txt
[[-0.74302961  0.32258907 -0.58638137  3.327299  ]
 [ 0.66483297  0.45637146 -0.59137312  2.70114165]
 [ 0.07683719 -0.82925348 -0.55356532  1.35660008]
 [ 0.          0.          0.          1.        ]]

# transform matrix stored in nerfstudio/transforms.json
[[0.66483301, -0.45637144, 0.591373122, 2.70114157],
[-0.74302967,-0.322589057,0.586381441,3.32729927],
[-0.07683724,-0.829253521,-0.55356538,-1.35660021],
[0.0,        0.0,       0.0,        1.0      ]]
```

If do the same operation as the reference on the matrix from colmap it should be:

```
c2w = c2w(np.array([0, 2, 1, 3]), :]
c2w[2, :] *= -1

c2w = 
array([[-0.74302961,  0.32258907, -0.58638137,  3.327299  ],
       [ 0.07683719, -0.82925348, -0.55356532,  1.35660008],
       [-0.66483297, -0.45637146,  0.59137312, -2.70114165],
       [ 0.        ,  0.        ,  0.        ,  1.        ]])
```

I'm not sure where is the problem.

ozgurcelik:
Hi. I can run scannetpp data with nerfacto and instant-ngp but it doesn't work with splatfacto:

`ns-train nerfacto scannetpp-data --data /cluster/scratch/tcelik/scannet++/data/56a0ec536c` works. so does `ns-train instant-ngp scannetpp-data --data /cluster/scratch/tcelik/scannet++/data/56a0ec536c`. But when I do `ns-train splatfacto scannetpp-data --data /cluster/scratch/tcelik/scannet++/data/56a0ec536c` I get


ns-train 8 <module>
sys.exit(entrypoint())

train.py 262 entrypoint
main(

train.py 247 main
launch(

train.py 189 launch
main_func(local_rank=0, world_size=world_size, config=config)

train.py 100 train_loop
trainer.train()

trainer.py 261 train
loss, loss_dict, metrics_dict = self.train_iteration(step)

profiler.py 112 inner
out = func(*args, **kwargs)

trainer.py 496 train_iteration
_, loss_dict, metrics_dict = self.pipeline.get_train_loss_dict(step=step)

profiler.py 112 inner
out = func(*args, **kwargs)

base_pipeline.py 303 get_train_loss_dict
loss_dict = self.model.get_loss_dict(model_outputs, batch, metrics_dict)

splatfacto.py 868 get_loss_dict
mask = self._downscale_if_required(batch["mask"])

splatfacto.py 649 _downscale_if_required
return resize_image(image, d)

splatfacto.py 95 resize_image
return tf.conv2d(image.permute(2, 0, 1)[:, None, ...], weight, stride=d).squeeze(1).permute(1, 2, 0)

RuntimeError:
"slow_conv2d_cpu" not implemented for 'Bool'

Do you have any idea why this might be?

PR #2509, opened 2023-10-11
Integrate scene click capability from original viewer

chungmin99:
In the original viewer, a user could click on a point in the screen, to specify a "ray"-like interaction (akin to a pointer finger stick originating from the client camera, going through the clicked pixel on the image plane).

This replicates the original capability in viewer_beta, with a few differences:
 - Functionality to remove scene click listeners, once instantiated (`unregister_click_cb`)
 - Wraps the scene click events (which take as input nerfstudio's `ViewerClicks`) inside a function that takes as input `ScenePointerEvent`, for viser-nerfstudio compatibility.

(Implementation based on the recently integrated viser PR https://github.com/nerfstudio-project/viser/pull/106)


kerrj (approved):
sgtm

PR #2519, opened 2023-10-12
Fix multiclient jitter with separate render threads for each client

ginazhouhuiwu:
- separate `RenderStateMachine` threads are created for each newly connected client in replacement of the old universal render thread --> clients now able to control their own cameras
- handle disconnected clients by killing respective threads


kerrj (approved):
lgtm

nerfstudio/viewer_beta/viewer.py
```diff
@@ -329,7 +342,7 @@ def update_scene(self, step: int, num_rays_per_batch: Optional[int] = None) -> N
         """
         self.step = step
 
-        if self.camera_state is None:
+        if len(self.render_statemachines) is None:
```
brentyi, line 345: Is this condition still doing something? Will this ever be `None`?

nerfstudio/viewer_beta/viewer.py
```diff
@@ -184,24 +185,30 @@ def nested_folder_install(folder_labels: List[str], element: ViewerElement):
             ]
         for c in self.viewer_controls:
             c._setup(self)
-        self.render_statemachine.start()
+
+    def get_camera_state(self, client):
```
brentyi, line 189: Can we make this: `def get_camera_state(self, client: viser.ClientHandle) -> CameraState:`

brentyi (commented):
Exciting!!

nerfstudio/viewer_beta/viewer.py
ginazhouhuiwu: Oops I changed it now to `if len(self.render_statemachines) == 0`

ginazhouhuiwu:
I accidentally overlooked [`get_camera`](https://github.com/nerfstudio-project/nerfstudio/blob/042c7d1465d3760cd5e256907f8ede8035a7364b/nerfstudio/viewer_beta/viewer_elements.py#L120) in `viewer_elements.py`. 

"Returns the Cameras object representing the current camera for the viewer." What is the desired behavior now that we want multiple clients to be able to control the same viewer?

ginazhouhuiwu:
> "Returns the Cameras object representing the current camera for the viewer." What is the desired behavior now that we want multiple clients to be able to control the same viewer?

I'm changing it to take `client` as a parameter but I don't see where this function is called anywhere else.

PR #2521, opened 2023-10-13
Gaussian splatting

kerrj:
A home-grown re-implementation of splatting! It uses [gsplat](https://github.com/nerfstudio-project/gsplat) as the rasterization backend, which has some nice features like rendering higher dimensional features and more pythonic bindings exposed to the user. Here's a video of a partially trained model in the viewer, and some completed video renders:

https://github.com/nerfstudio-project/nerfstudio/assets/15806475/965c4d6c-cbc6-455b-b588-e29bfe61f9f1

https://github.com/nerfstudio-project/nerfstudio/assets/15806475/e23a7c17-ffce-4125-af23-32cf2eb1ed0c

https://github.com/nerfstudio-project/nerfstudio/assets/15806475/b25cf6d2-7654-493f-9285-b71b1b4a533c

Implementation-wise, there are still some things to be done:
- unify the datamanager pipeline, currently there is a full_images_datamanager which is disjoint from the
- to support rasterization, we had to overload get_outputs to take in a Camera or Raybundle. This is a bit hacky, ideally we would have a cleaner way of supporting this behavior, maybe a subclass of Model called RasterizationModel or something.
- Some parts of the pipeline like exporting point-clouds won't fully work (cropping not implemented yet), and it's unclear how well the depth will work for pointclouds
- Currently cropping is implemented by calling a set_crop function. This works well, but it's a bit divergent from the RayBundle strategy of cropping which crops the near and far plans for each ray before passing it into the model. this could make code messy in rendering/export pipelines since we need to case on whether the model is a raybundle or rasterization model

There's lots of exciting things which _could_ be done too, like floater removal via the viewer, exporting as PLY, or exporting cropped scenes to get object-centric models.


nerfstudio/models/gaussian_splatting.py
```diff
@@ -0,0 +1,653 @@
...
+    b = -t
+    r = znear * math.tan(0.5 * fovx)
+    l = -r
+    n = znear
+    f = zfar
+    return torch.tensor(
+        [
+            [2 * n / (r - l), 0.0, (r + l) / (r - l), 0.0],
+            [0.0, 2 * n / (t - b), (t + b) / (t - b), 0.0],
+            [0.0, 0.0, (f + n) / (f - n), -1.0 * f * n / (f - n)],
```
maturk, line 77: Is there a mistake in the projmat[2,2]? 
refer to reference code: https://github.com/graphdeco-inria/gaussian-splatting/blob/414b553ef1d3032ea634b7b7a1d121173f36592c/utils/graphics_utils.py#L69

maturk: I guess there is some difference to [OpenGL](http://www.songho.ca/opengl/gl_projectionmatrix.html) convention too. Wonder which is correct now.

nerfstudio/viewer/server/render_state_machine.py
```diff
@@ -142,12 +149,21 @@ def _render_img(self, cam_msg: CameraMessage):
                             device=self.viewer.get_model().device,
                         )
                     with background_color_override_context(background_color), torch.no_grad():
-                        outputs = self.viewer.get_model().get_outputs_for_camera_ray_bundle(camera_ray_bundle)
+                        if isGaussianSplattingModel:
+                            outputs = self.viewer.get_model().get_outputs_for_camera_ray_bundle(camera)
```
pwais, line 153:
```suggestion
                            outputs = self.viewer.get_model().get_outputs_for_camera_ray_bundle(camera_ray_bundle, camera=camera)
```

nerfstudio/viewer/server/render_state_machine.py
```diff
@@ -142,12 +149,21 @@ def _render_img(self, cam_msg: CameraMessage):
...
+                        else:
+                            outputs = self.viewer.get_model().get_outputs_for_camera_ray_bundle(camera_ray_bundle)
                 else:
                     with torch.no_grad():
-                        outputs = self.viewer.get_model().get_outputs_for_camera_ray_bundle(camera_ray_bundle)
+                        if isGaussianSplattingModel:
+                            outputs = self.viewer.get_model().get_outputs_for_camera_ray_bundle(camera)
```
pwais, line 159:
```suggestion
                            outputs = self.viewer.get_model().get_outputs_for_camera_ray_bundle(camera_ray_bundle, camera=camera)
```

nerfstudio/viewer/server/render_state_machine.py
```diff
@@ -127,7 +128,13 @@ def _render_img(self, cam_msg: CameraMessage):
         assert camera is not None, "render called before viewer connected"
 
         with self.viewer.train_lock if self.viewer.train_lock is not None else contextlib.nullcontext():
-            camera_ray_bundle = camera.generate_rays(camera_indices=0, aabb_box=self.viewer.get_model().render_aabb)
+            # TODO jake-austin: Make this check whether the model inherits from a camera based model or a ray based model
+            # TODO Zhuoyang: First made some dummy judgements, need to be fixed later
+            isGaussianSplattingModel = isinstance(self.viewer.get_model(), GaussianSplattingModel)
+            if isGaussianSplattingModel:
+                pass
```
pwais, line 135:
```suggestion
                # TODO fix me before ship
                camera_ray_bundle = camera.generate_rays(camera_indices=0, aabb_box=self.viewer.get_model().render_aabb)
```

pwais (commented):
make `ns-viewer` work again (viewer _during train_ works at time of writing).  my guess is the viewer integration might change but at least this patch helps the branch in current state

nerfstudio/models/gaussian_splatting.py
maturk: I think the difference in (f+n) and just (f) is negligible probably. We can leave it as is.

nerfstudio/viewer/server/render_state_machine.py
kerrj: Oops yes I only tested in the viewer_beta, since the old one will be deprecated soon. thanks for these fixes.

nerfstudio/models/gaussian_splatting.py
pwais: https://github.com/nerfstudio-project/nerfstudio/pull/2521/files/40c05be7a60bde25adc28cd281bb9d742c2a77e9#diff-fa59b1578ecffb1d76c0a43bec7a2a12217696ce7cb3632aba6af5731a881459R505   well the znear and zfar right now are hard-coded

( which is a bit confusing because the model inherits collider params ... https://github.com/nerfstudio-project/nerfstudio/blob/8a3e24738ba7b31139418e109896915a69bfb6a4/nerfstudio/models/base_model.py#L48   )

maybe the overall question is worth returning to later to see if the viser webgl viewer https://github.com/nerfstudio-project/viser/pull/110/files  ends up consistent with images rendered thru nerfstudio

seems at the least znear and zfar should be user configurable instead of hardcoded...  those are effectively hyperparameters, if not also the whole P matrix used ... neither gsplat back-ends use OpenGL render-to-texture tho at least.

nerfstudio/viewer/server/render_state_machine.py
pwais: tanku!  sure i mean like maybe in nerfstudio RayBundle should just also include the camera the rays came from haha

pablovela5620:
> make `ns-viewer` work again (viewer _during train_ works at time of writing). my guess is the viewer integration might change but at least this patch helps the branch in the current state

I am currently getting the opposite problem, get the following error when trying to visualize training, but am able to see just fine when using `ns-viewer`
```
(viser) Connection opened (3, 4 total), 2641 persistent messages
connection handler failed
Traceback (most recent call last):
  File "/home/pablo/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/websockets/legacy/server.py", line 240, in handler
    await self.ws_handler(self)
  File "/home/pablo/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/viser/infra/_infra.py", line 260, in serve
    await asyncio.gather(
  File "/home/pablo/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/viser/infra/_infra.py", line 416, in _consumer
    message = message_class.deserialize(raw)
  File "/home/pablo/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/viser/infra/_messages.py", line 99, in deserialize
    message_type = cls._subclass_from_type_string()[cast(str, mapping.pop("type"))]
KeyError: 'CameraMessage'
```

machenmusik:
view-while-train is working for me.

Training seems slow past 4800 steps though, looks like it wants more than 16GB VRAM?

nerfstudio/viewer/server/render_state_machine.py
```diff
@@ -142,12 +150,21 @@ def _render_img(self, cam_msg: CameraMessage):
...
                 else:
                     with torch.no_grad():
-                        outputs = self.viewer.get_model().get_outputs_for_camera_ray_bundle(camera_ray_bundle)
+                        if isGaussianSplattingModel:
+                            outputs = self.viewer.get_model().get_outputs_for_camera_ray_bundle(camera_ray_bundle, camera=camera)
+                        else:
+                            outputs = self.viewer.get_model().get_outputs_for_camera_ray_bundle(camera_ray_bundle)
                 self.viewer.get_model().train()
-        num_rays = len(camera_ray_bundle)
+        if True:
```
machenmusik, line 164: Should this be `if IsGaussianSplattingModel:` instead of True?

machenmusik:
Crashed with CUDA OOM shortly after 8K steps.
(Not sure what expectations are.)

pwais:
@pablovela5620 viewer during train worked for me, using new viser UI; your traceback seems that maybe your environment does not have the right / latest viser release that this branch uses perhaps?  i am installing nerfstudio using `pip install -v -e . ` which should pull in required viser ....

I also see a slowdown at ~5000 iters but perhaps that's known.  The iteration latency goes up and GPU power usage too so mebbe an issue with gsplat kernel moreso than nerfstudio.

pwais:
FWIW it looks like for most models in nerfstudio, "depth" means "ray length", and in particular the viewer may convert this to z-depth silently always: https://github.com/nerfstudio-project/nerfstudio/blob/e8cee3fdadda5068309cb14bdf04aced082cb091/nerfstudio/viewer_beta/render_state_machine.py#L158  

example of "ray length" in nerfstudio: https://github.com/nerfstudio-project/nerfstudio/blob/e8cee3fdadda5068309cb14bdf04aced082cb091/nerfstudio/model_components/renderers.py#L333


however, gsplat appears to render z-depth? https://github.com/nerfstudio-project/gsplat/blob/3a22dd497eaebf57394b7ee84f3a75b2d32d6ce1/gsplat/cuda/csrc/forward.cu#L90    

probably worth some comments / docs for clarification.  There is no explicit remark in nerfstudio data conventions https://docs.nerf.studio/quickstart/data_conventions.html#depth-images  

and the issue can be really important for wide angle cameras / getting very close to objects.

machenmusik:
This branch is getting to be pretty substantial... is there a schedule/timetable planned to consider merging?

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -0,0 +1,412 @@
...
+            else:
+                raise NotImplementedError("Only perspective and fisheye cameras are supported")
+            data["image"] = torch.from_numpy(image)
+
+            if "mask" in data:
+                mask = data["mask"].numpy()
+                if camera.camera_type.item() == CameraType.PERSPECTIVE.value:
+                    mask = cv2.undistort(mask, K, distortion_params, None, None) # type: ignore
+                elif camera.camera_type.item() == CameraType.FISHEYE.value:
+                    mask = cv2.fisheye.undistortImage(mask, K, distortion_params, None, None)
```
nikmo33, line 185: I think this is a mismatch from the images since for the images we compute the new image by using newK - I think either we should reuse the map generated for the images or pass in newK into this call. 
```suggestion
                    mask = cv2.fisheye.undistortImage(mask, K, distortion_params, None, None)
```

nikmo33: Also notice we dont undistort the depth image in this case. Maybe it would be cleaner to move both of this under the main fisheye condition?

kerrj: This was fixed in #2540 !

nerfstudio/pipelines/base_pipeline.py
```diff
@@ -371,26 +373,27 @@ def get_average_eval_image_metrics(
...
                 metrics_dict, images_dict = self.model.get_image_metrics_and_images(outputs, batch)
 
-                if output_path is not None:
-                    camera_indices = camera_ray_bundle.camera_indices
-                    assert camera_indices is not None
-                    for key, val in images_dict.items():
-                        Image.fromarray((val * 255).byte().cpu().numpy()).save(
-                            output_path / "{0:06d}-{1}.jpg".format(int(camera_indices[0, 0, 0]), key)
-                        )
+                # if output_path is not None:
```
tancik, line 384: Delete

tancik (commented):
nits

tancik (approved):
🎉

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -0,0 +1,410 @@
...
+    along with relevant information about camera intrinsics
+    """
+    eval_num_images_to_sample_from: int = -1
+    """Number of images to sample during eval iteration."""
+    eval_num_times_to_repeat_images: int = -1
+    """When not evaluating on all images, number of iterations before picking
+    new images. If -1, never pick new images."""
+    eval_image_indices: Optional[Tuple[int, ...]] = (0,)
+    """Specifies the image indices to use during eval; if None, uses all."""
+    cache_images: Literal["no-cache", "cpu", "gpu"] = "cpu"
```
pwais, line 62: @kerrj  `no-cache` does nothing, it actually implies `cpu`

PR #2522, opened 2023-10-13
Fix datamanagers' get_rays_per_batch functions in case of using DynamicBatchPipeline

sjj118:
This correct the "Train Rays / Sec" when using DynamicBatchPipeline


kerrj (approved):
good catch

PR #2525, opened 2023-10-14
Add camera_type dropdown in new viewer

cvachha:
Added a gui dropdown menu for selecting the camera type in the viser viewer render panel. Currently, the render view doesn't update to preview the equirectangular or fisheye render but the user select the camera type which updates the camera_path json file.

Also updated image url of the nerfstudio logo in the viewer


kerrj (approved):
sgtm

PR #2530, opened 2023-10-16
Added UI to load camera path from existing path for the dataset

THE-COB:
Adds button to load camera path. Button creates modal with dropdown to select existing camera paths


kerrj:
Can you post some pics/videos?

kerrj (approved):
lgtm, thanks!

PR #2535, opened 2023-10-17
Gaussian Splatting export function

akristoffersen:


3a1b2c3:
thank you

akristoffersen:
Looks like there is still some sort of transformation difference between what nerfstudio uses and what the major open-source viewers use. But they are at least renderable there.

nerfstudio/scripts/exporter.py
```diff
@@ -469,13 +464,64 @@ def main(self) -> None:
...
+            map_to_tensors["normals"] = o3d.core.Tensor(np.zeros_like(positions), o3d.core.float32)
+
+            colors = model.colors.data.cpu().numpy()
+            for i in range(colors.shape[1]):
+                map_to_tensors[f"f_dc_{i}"] = colors[:, i]
+
+            shs = model.shs_rest.data.cpu().numpy()
+            shs = shs.reshape((colors.shape[0], -1, 1))
+            for i in range(shs.shape[-1]):
+                map_to_tensors[f"f_rest_{i - 3}"] = shs[:, i]
```
alexander-kristoffersen-skydio, line 499: should be just i

machenmusik:
Just tried this, using same output as https://github.com/nerfstudio-project/nerfstudio/pull/2537#issuecomment-1769500466

Since there are no spherical harmonics in the output to export, this PR throws an error:
```
│ C:\code\nerfstudio\nerfstudio\scripts\exporter.py:499 in main                                    │
│                                                                                                  │
│   496 │   │   │   shs = model.shs_rest.data.cpu().numpy()                                        │
│   497 │   │   │   shs = shs.reshape((colors.shape[0], -1, 1))                                    │
│   498 │   │   │   for i in range(shs.shape[-1]):                                                 │
│ ❱ 499 │   │   │   │   map_to_tensors[f"f_rest_{i}"] = shs[:, i]                                  │
│   500 │   │   │                                                                                  │
│   501 │   │   │   map_to_tensors["opacity"] = model.opacities.data.cpu().numpy()                 │
│   502                                                                                            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
IndexError: index 0 is out of bounds for axis 1 with size 0
```

machenmusik:
When spherical harmonics are present, this PR works fine.
image

nerfstudio/scripts/exporter.py
machenmusik: done

machenmusik:
This branch now works for exporting outputs with or without spherical harmonics.

machenmusik:
> Looks like there is still some sort of transformation difference between what nerfstudio uses and what the major open-source viewers use.

From https://github.com/aras-p/UnityGaussianSplatting/blob/main/readme.md#usage
```
The rendering takes game object transformation matrix into account; the official gaussian splat models seem to be all rotated by about -160 degrees around X axis, and mirrored around Z axis, so in the sample scene the object has such a transform set up.
```

kerrj:
Is the transformation an issue? or can this be merged. also how does this relate to the other PR on the same topic? @alexander-kristoffersen-skydio @machenmusik

machenmusik:
IMO could merge, and if needed do another PR for rotation later; I did not try the other PR using spherical harmonics, and don't recall whether it had same rotation issue, sorry

akristoffersen:
Cool, I'll merge. I think the other PR is essentially equivalent. If there are additional fixes we can add them as other PRs.

PR #2537, opened 2023-10-17
export gaussian splats reconstruction as PLY

cdcseacave:
Export a checkpoint trained with gaussian splatting in nerfstudio to a PLY file compatible with the original format. For example it can be visualized at `https://antimatter15.com/splat/` by simply drag&drop the ply on the browser window.

[Video.webm](https://github.com/nerfstudio-project/nerfstudio/assets/2802670/cfabc940-f1ee-453d-980d-d40e98e20a83)

### DIFF
--- a/nerfstudio/scripts/exporter.py
+++ b/nerfstudio/scripts/exporter.py
@@ -46,6 +46,7 @@ from nerfstudio.exporter.marching_cubes import (
     generate_mesh_with_multires_marching_cubes,
 )
 from nerfstudio.fields.sdf_field import SDFField
+from nerfstudio.models.gaussian_splatting import GaussianSplattingModel
 from nerfstudio.pipelines.base_pipeline import Pipeline, VanillaPipeline
 from nerfstudio.utils.eval_utils import eval_setup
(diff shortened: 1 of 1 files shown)


maturk:
There are now two PRs for this, ref #2535. Can these be combined?

cdcseacave:
That one does not work, i had to reimplement

machenmusik:
I can confirm this works.  Although initial orientation from drag-n-drop onto this page is not as expected; this is after rotations etc.  
Will also try the other one...
image

machenmusik:
> That one does not work, i had to reimplement
> […](#)
> On Tue, Oct 17, 2023 at 15:47 Matias Turkulainen ***@***.***> wrote: There are now two PRs for this, ref #2535 <#2535>. Can these be combined?

See discussion on https://github.com/nerfstudio-project/nerfstudio/pull/2535; it does work if spherical harmonics are present in the output to be exported, but fails if they are not.

kerrj:
Given the other PR was merged and it seems to support both RGB only and SH, can we close this one? @machenmusik @cdcseacave

PR #2540, opened 2023-10-18
Fix the bugs of mask input in gaussian-splatting branch

Zhuoyang-Pan:
When I try to use the mask images in 3dgs branch of nerfstudio, I encounter the error 
> cv2.error: OpenCV(4.6.0) error: (-5:Bad argument) in function 'undistort'
> Overload resolution failed:
>  - src data type = 0 is not supported
>  - Expected Ptr<cv::UMat> for argument 'src'

I guess this is because we save the masks as bool somewhere in data_util and here we need to convert it back to uint8 before undistorting it.


kerrj:
this is actually a bit out of date now that the undistorting code was fixed, could you change it to match the other code?

kerrj (approved):

PR #2550, opened 2023-10-20
make refine intrinsics during ns-process-data optional

machenmusik:
ns-process-data will still refine intrinsics by default for compatibility.

Specify --no-refine-intrinsics to turn off.


machenmusik:
Motivation: at least at present, there seems to be significant improvement using non-refined versions of some datasets with #2521.

kerrj (approved):
sgtm

PR #2553, opened 2023-10-23
Few viewer changes

cvachha:
Small changes for the new viewer including changing the FPS slider to a text entry and fixing the incorrect resolution text box description.

Added a check if render_time is non zero in the render_state_machine.py in the new viewer due to a divide by zero error occasionally causing the viewer to stop running. I noticed this usually happened when using nerfacto-huge.


tancik (approved):
LGTM

PR #2555, opened 2023-10-24
Add gradient scaling option to more methods

kobejean:
Here is a proposal to add gradient scaling as an option to other methods like instant-ngp, mipnerf, tensorf, vanilla-nerf.  It is disabled by default so it won't change existing behaviour. The code was copied from the nerfacto method.


brentyi (approved):
Looks good to me!

PR #2556, opened 2023-10-24
Expose camera frustums size and occlusions to CLI

ethanweber:
I added two parameters to the Viewer config. Now we can change the size of the training & eval camera frustums, and we can enable/disable rendering the scene with depth compositing. This is useful for (1) when the scene is small and frustums are too tiny or (2) when there are floaters or artifacts in the way and you still want to see the camera frustums.

**Default, trained for a little bit.**
```bash
ns-train nerfacto --data data/nerfstudio/plane --vis viewer_beta
```
Screenshot 2023-10-24 at 12 55 01 PM

**Modifying the camera frustum scale from the default 0.1 to 1.0 (10x) and turning off depth compositing.**
```bash
ns-train nerfacto --data data/nerfstudio/plane --vis viewer_beta --viewer.camera_frustum_scale 1.0 --viewer.default_composite_depth False
```
Screenshot 2023-10-24 at 12 53 22 PM


kerrj (approved):
sgtm

PR #2557, opened 2023-10-24
Ply dataparser

CardiacMangoes:
IN PROGRESS do not merge yet

Loads a ply file as initial point cloud if specified

Specify `transforms.json` with "ply_file_path" in header, i.e

{
    "camera_model": "OPENCV",
    "cx": 320,
    "cy": 240,
    "w": 640,
    "h": 480,
    "ply_file_path": "smith.ply",
    ...


ethanweber:
Nice, this is useful. I think we can merge the dataparser with its own PR, and then a separate PR should add code to ns-process-data to generate a PLY from colmap when it finishes.

kerrj:
I think this is useful enough to merge into main, since it includes generally loading .ply files in the dataparser and also metric export option

kerrj:
Need to pull out just the relevant changes though, changing the base branch includes all of splatting too

kerrj (approved):
sgtm

PR #2558, opened 2023-10-25
ns-process-data images fix: multinerf uses equals sign

machenmusik:
(Bringing over the rest of #2554 to main.)


kerrj (approved):
nice catch

PR #2575, opened 2023-11-01
`process_data` support multi-camera intrinsics

Legend94rz:
Allow multi-camera intrinsics (per-frame) when using `ns-process-data`.

After changes, when there are multiple camera founded in COLMAP database, it will export the `transforms.json` in [per-frame format](https://docs.nerf.studio/quickstart/data_conventions.html#camera-intrinsics).

It still doesn't support per-frame camera model for now.

### DIFF
--- a/nerfstudio/process_data/colmap_utils.py
+++ b/nerfstudio/process_data/colmap_utils.py
@@ -451,16 +451,16 @@ def colmap_to_json(
 
     if set(cam_id_to_camera.keys()) != {1}:
         cameras = {k: parse_colmap_camera_params(v) for k, v in cam_id_to_camera.items()}
-        model = list(set(x['camera_model'] for x in cameras.values()))
+        model = list(set(x["camera_model"] for x in cameras.values()))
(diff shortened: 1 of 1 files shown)


nerfstudio/process_data/colmap_utils.py
```diff
@@ -449,9 +450,17 @@ def colmap_to_json(
         frames.append(frame)
 
     if set(cam_id_to_camera.keys()) != {1}:
-        raise RuntimeError("Only single camera shared for all images is supported.")
-    out = parse_colmap_camera_params(cam_id_to_camera[1])
-    out["frames"] = frames
+        cameras = {k: parse_colmap_camera_params(v) for k, v in cam_id_to_camera.items()}
+        model = list(set(x["camera_model"] for x in cameras.values()))
```
jb-ye, line 454: What if we have multiple camera models? Shall we also raise an exception here?

Maybe we should assert this assumption since colmap only supports one camera model at a time.

Issue #2438, opened 2023-09-17
Mac Installation Instructions?


sudoku-lord:
Hi! I saw on [this link](https://neuralradiancefields.io/nerfstudio-v0-3-2-released/) that NeRFStudio is now compatible with MacBooks which have M1/M2 chips. I'm wondering if the instructions for installation are the same as the PC ones, or whether there's a separate set of instructions for Mac installation available somewhere else. Additionally, are there Docker images for Mac?

Thank you so much!

tancik:
For mac, just run `pip install nerfstudio`, skip the `tiny-cuda-nn` and `pytorch` dependency steps.

LuyuanPeng:
> For mac, just run `pip install nerfstudio`, skip the `tiny-cuda-nn` and `pytorch` dependency steps.

I have a M2 Mac on Ventura 13.4.1. I could not install nerfstudio using `pip3 install nerfstudio` in a clean conda environment. The output for this command is as below. 
```
Collecting nerfstudio
  Obtaining dependency information for nerfstudio from https://files.pythonhosted.org/packages/2d/db/2f6697579b7813b5413118726f30cb6e3e6427580385943c797d9cbd02cb/nerfstudio-0.3.3-py3-none-any.whl.metadata
  Using cached nerfstudio-0.3.3-py3-none-any.whl.metadata (23 kB)
Collecting appdirs>=1.4 (from nerfstudio)
  Using cached appdirs-1.4.4-py2.py3-none-any.whl (9.6 kB)
Collecting av>=9.2.0 (from nerfstudio)
  Using cached av-10.0.0-cp311-cp311-macosx_11_0_arm64.whl (19.6 MB)
Collecting cryptography>=38 (from nerfstudio)
  Obtaining dependency information for cryptography>=38 from https://files.pythonhosted.org/packages/91/68/5c33bb0115b3413a974dd4d23625b99ed22522582b513f82e93ce00f954c/cryptography-41.0.3-cp37-abi3-macosx_10_12_universal2.whl.metadata
  Using cached cryptography-41.0.3-cp37-abi3-macosx_10_12_universal2.whl.metadata (5.2 kB)
Collecting tyro>=0.5.3 (from nerfstudio)
[...]

PR #2577, opened 2023-11-01
Fix torch.compile for torch 2.1

jkulhanek:
Eager backend support in torch.compile seems not to support the reduce-overhead mode anymore. This PR switches to the default inductor backend. Some data points: https://github.com/nerfstudio-project/nerfstudio/pull/1922

@kerrj 
Adresses #2438.


tancik (approved):
LGTM

PR #2584, opened 2023-11-05
Remove compile line in camera_utils

ginazhouhuiwu:
Re commit 85677d310fcb8700b284225de646c63bb02b83de, which caused the following when training nerfacto:

Screenshot 2023-11-05 at 8 37 42 AM
Screenshot 2023-11-05 at 8 37 58 AM


tancik:
@jkulhanek FYI

tancik (approved):

ginazhouhuiwu:
Removing the compile line instead of reverting changes since reverting it will break torch 2.1

ginazhouhuiwu:
I believe this is just a temporary solution for now since main was having some issues. Will try to patch this

PR #2585, opened 2023-11-06
Speed up mask sampling with rejection sampling 

anc2001:
Speeds up mask sampling and avoids OOM errors for large images / masks by avoiding `torch.nonzero` with rejection sampling. Found that this solves many of the issues related to inefficient sampling with masks. 

Ex. 
rays / sec without masks 
no_masks

rays / sec with current mask sampling 
masks_old

rays / sec with new mask sampling
masks_ne


blacksino:
you could save nonzero index  as a member either.

akristoffersen:
I recently relied on this PR to dramatically speed up training with masks. @anc2001 would you be interested in getting this PR over the finish line? Perhaps with a optional flag as @KevinXu02 suggests, though I think it may make more sense for this to be the default behavior, and revert to the slower behavior with an optional flag.

anc2001:
Yes @akristoffersen I'd love to push this PR over the finish line! Just been a little busy recently and have had some issues setting up the dev environment.

machenmusik:
Do we have any usage data on how often this is used with sparse data such as lidar?  My impression is that sparse data usage is less common, and so enabling this option by default would be better for the majority.

anc2001:
A better solution than the optional flag might be some kind of adaptive thresholding where if the number of valid pixels in the mask is below a certain threshold percentage revert functionality back to original and if above rejection sample. This might be better for datasets with a mix of sparse and dense masks, but I'm not sure if that's necessary so will leave to any future PRs if people want that kind functionality.

nerfstudio/data/pixel_samplers.py
```diff
@@ -95,15 +99,39 @@ def sample_method(
...
+                    if num_valid == batch_size:
+                        break
+                    else:
+                        replacement_indices = (
+                            torch.rand((batch_size - num_valid, 3), device=device)
+                            * torch.tensor([num_images, image_height, image_width], device=device)
+                        ).long()
+                        indices[~chosen_indices_validity] = replacement_indices
+
+                if num_valid != batch_size:
```
akristoffersen, line 123: Instead of just raising warning, I think it would make sense to default back to the slow non-rejection sampling if this occurs. I would still issue a warning, but it would suck for training to fail on the off chance that not enough valid indices are generated in time.

akristoffersen (approved):
LGTM, small comment but otherwise looks good to merge :D

akristoffersen (commented):
small nit: With the most recent change, if the first iter fails with rejection sampling, a warning will be generated but the indices that will be returned will not necessarily only contained valid locations within the mask. Every other sampling will be fine, but the first one will still have invalid indices.

I think instead, we should throw away the current generated indices and start from scratch with the non-rejection sampling method, as well as change the config flag to use the non-rejection sampling method for all future iterations.

akristoffersen:
LGTM, thanks again @anc2001! feel free to merge :D

anc2001:
Hey @akristoffersen, I don't have write access to merge this PR! Do you know who we would need to @ to get this merged?

akristoffersen:
That would be me! Should merge once all checks are done

nepfaff:
@anc2001, did you get these timing results with `--pipeline.datamanager.masks-on-gpu True --pipeline.datamanager.images-on-gpu True` or without these options? Not including these still seem to be slow for me but maybe I have a bug somewhere

Issue #2542, opened 2023-10-19
[gsplat] more than 512 images cause ns-train error in viewer_beta?


machenmusik:
I have a dataset which I can't share, that has 600 images (480x480).
`ns-process-data images ...` works fine, but
`ns-train gaussian-splatting --pipeline.model.sh-degree 0 --data ...` fails with this error
```
IndexError: index 512 is out of bounds for axis 0 with size 512
```

Adding `--viewer.max-num-display-images 1024` to the training command avoids the error.

PR #2586, opened 2023-11-06
Fix large dataset (> 512 images) training in viewer_beta

ginazhouhuiwu:
Re #2542. Fix index bug by creating two different iterators in `update_camera_poses`, since `self.original_c2w`(dict) and `c2ws_delta`(np array) should be indexed into differently.


tancik (approved):
Thanks for fixing!

PR #2597, opened 2023-11-11
Update LICENSE

cpheinrich:
Filling in the license to what I think it should be. We're adding nerfstudio to our list of third party libraries :)


brentyi (approved):
Thanks @cpheinrich 🙂

PR #2613, opened 2023-11-16
Added support for .tif and .tiff files in equirect utils

THE-COB:
https://github.com/nerfstudio-project/nerfstudio/pull/1446#issuecomment-1812115064


tancik (approved):
LGTM

PR #2617, opened 2023-11-16
Add support for Project Aria datasets.

sweeneychris:
1) An export script for processing Project Aria datasets
2) Add support for Fisheye624 cameras
3) Add support for center-crop masking and sampling

Example commands for running nerfstudio on Project Aria data:

```
python scripts/datasets/process_project_aria.py --vrs_file=<path_to_vrs> --mps_data_dir=<path to MPS directory> --output_dir=<nerfstudio_data_dir>

ns-train nerfacto --data <nerfstudio_data_dir>  nerfstudio-data --orientation-method none

### DIFF
--- a/nerfstudio/scripts/datasets/process_project_aria.py
+++ b/nerfstudio/scripts/datasets/process_project_aria.py
@@ -23,13 +23,12 @@ import numpy as np
 import tyro
 from PIL import Image
 
-try:
-    from projectaria_tools.core import mps
-    from projectaria_tools.core.data_provider import VrsDataProvider, create_vrs_data_provider
-    from projectaria_tools.core.sophus import SE3
-except ImportError:
-    print("projectaria_tools not found, please install with pip3 install projectaria-tools'[all]'")
-    sys.exit(1)
+from projectaria_tools.core import mps
+from projectaria_tools.core.data_provider import (
+    VrsDataProvider,
+    create_vrs_data_provider,
+)
+from projectaria_tools.core.sophus import SE3
 
 ARIA_CAMERA_MODEL = "FISHEYE624"
 
@@ -226,11 +225,6 @@ class ProcessProjectAria:
             transform_file.write_text(json.dumps(nerfstudio_frames))
 
 
-def entrypoint():
-    """Entrypoint for use with pyproject scripts."""
+if __name__ == "__main__":
     tyro.extras.set_accent_color("bright_yellow")
     tyro.cli(ProcessProjectAria).main()
-
-
-if __name__ == "__main__":
-    entrypoint()
--- a/nerfstudio/scripts/datasets/process_project_aria.py
+++ b/nerfstudio/scripts/datasets/process_project_aria.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+import sys
 import threading
 from dataclasses import dataclass
 from pathlib import Path
@@ -21,9 +22,14 @@ from typing import Dict, List
 import numpy as np
 import tyro
 from PIL import Image
-from projectaria_tools.core import mps
-from projectaria_tools.core.data_provider import VrsDataProvider, create_vrs_data_provider
-from projectaria_tools.core.sophus import SE3
+
+try:
+    from projectaria_tools.core import mps
+    from projectaria_tools.core.data_provider import VrsDataProvider, create_vrs_data_provider
+    from projectaria_tools.core.sophus import SE3
+except ImportError:
+    print("projectaria_tools not found, please install with pip3 install projectaria-tools'[all]'")
+    sys.exit(1)
 
 ARIA_CAMERA_MODEL = "FISHEYE624"
 
@@ -180,6 +186,7 @@ class ProcessProjectAria:
     def main(self) -> None:
         """Generate a nerfstudio dataset from ProjectAria data (VRS) and MPS attachments."""
         # Create output directory if it doesn't exist.
+        self.output_dir = self.output_dir.absolute()
         self.output_dir.mkdir(parents=True, exist_ok=True)
 
         provider = create_vrs_data_provider(self.vrs_file.absolute().as_posix())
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by brentyi** [comment]
Had to fiddle with the `transforms.json` image paths, but I just gave this a try and it worked for me!

From Aria captures in Berkeley:

image

image

**Comment by sweeneychris** [comment]
Awesome! @brentyi could you expand a bit on what changes you needed to make in your transforms.json paths? Is it something I can update in the code?

File: nerfstudio/scripts/datasets/process_project_aria.py [hunk 1]
```diff
@@ -0,0 +1,227 @@
...
+
+        # create the NerfStudio frames from the AriaImageFrames.
+        print("Creating NerfStudio frames...")
+        RGB_VALID_RADIUS = 707.5
+        nerfstudio_frames = {
+            "camera_model": ARIA_CAMERA_MODEL,
+            "frames": [to_nerfstudio_frame(frame) for frame in aria_frames],
+            "fisheye_crop_radius": RGB_VALID_RADIUS,
+        }
+
```
brentyi, line 212: It could also be helpful to save the point cloud here, so #2521 can use it for initialization (once it's merged):

```python
        # import pandas as pd
        # import open3d as o3d
        points_path = self.mps_data_dir / "global_points.csv.gz"
        if points_path.exists():
            print("Found global points, saving to CSV...")
            df = pd.read_csv(points_path, compression="gzip")
            points = df[["px_world", "py_world", "pz_world"]].values
            pcd = o3d.geometry.PointCloud()
            pcd.points = o3d.utility.Vector3dVector(points)
            ply_file_path = self.output_dir / "global_points.ply"
            o3d.io.write_point_cloud(str(ply_file_path), pcd)

            nerfstudio_frames["ply_file_path"] = "global_points.ply"
        else:
            print("No global points found!")
```

We would also need to add support for Fisheye624 in the splatting pipeline... ideally we would integrate the camera model directly into gsplat/the rasterization kernels, but maybe just undistorting the label images is fine too? Seems like that's how the OpenCV fisheye model is being handled right now.

cc @kerrj, @vye16

vye16:
I'm hesitant to write custom kernels for each camera model, especially
because the affine approximation for Gaussians doesn't hold as well with
more distortion. In the meantime we can undistort the full images in the
gsplat PR.

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -0,0 +1,227 @@
...
+        # create an AriaImageFrame for each image in the VRS.
+        print("Creating Aria frames...")
+        aria_frames = [
+            to_aria_image_frame(provider, index, name_to_camera, t_world_devices, self.output_dir)
+            for index in range(0, provider.get_num_data(stream_id))
+        ]
+
+        # create the NerfStudio frames from the AriaImageFrames.
+        print("Creating NerfStudio frames...")
+        RGB_VALID_RADIUS = 707.5
```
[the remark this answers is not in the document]
sweeneychris: > (1) Is this necessary to hardcode?
 Right now the project aria tools don't provide a way to obtain the valid mask directly as a mask image. So I think that leaves us with two options 1) stick with hardcoding this value in the script and sample pixels uniformly in a circle given this value (current solution) 2) generate the valid mask image for all exported images and then use the pixel sampler with masks. Happy to go for either solution, the only main advantage of option 1) I see is saving a lot of memory by not having to load/store/process masks for each image.

2) Indeed, this value needs to be scaled based on the recording profile. This value works for profiles with 1408x1408 RGB so I will update the code to detect what size RGB images are present in the VRS to accomodate different profiles.

sweeneychris:
@vye16 Agreed -- it would be a nightmare to write custom kernels for each projection function. It would be nice if you could use jax or functorch to compute the jacobian of the projection functions dynamically (which would allow 3d -> 2d gaussian projection with correct covariance).

Undistorting Aria images for now is the right approach. I'd prefer to leave that for a followup PR if possible

brentyi:
Ok, makes sense to me! Undistortion in a separate PR seems fine. 🙂

nerfstudio/scripts/datasets/process_project_aria.py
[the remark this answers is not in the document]
SeaOtocinclus: @sweeneychris 1415 would be the value for full image size. Since we have only two RGB images size, the best could be to check the ImageSize of the RGB stream and just set 707.5 or 1408 accordingly

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -0,0 +1,227 @@
...
+def to_nerfstudio_frame(frame: AriaImageFrame) -> Dict:
+    return {
+        "fl_x": frame.camera.fx,
+        "fl_y": frame.camera.fy,
+        "cx": frame.camera.cx,
+        "cy": frame.camera.cy,
+        "distortion_params": frame.camera.distortion_params.tolist(),
+        "w": frame.camera.width,
+        "h": frame.camera.height,
+        "file_path": frame.file_path,
```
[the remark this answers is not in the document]
sweeneychris: I added this to the main logic:

        self.output_dir = self.output_dir.absolute()

nerfstudio/scripts/datasets/process_project_aria.py
sweeneychris: Mind if I leave this for a future commit? Since the main nerf pipeline won't use the points (right?)

brentyi:
Left a last question above but things look ready to me! I made some minor tweaks: a docs update, dependencies, and exposed the Aria executable via `ns-process-data aria`.

Also checked the whole pipeline on my machine; here's a render from the data I collected earlier:

https://github.com/nerfstudio-project/nerfstudio/assets/6992947/7c28e9ee-fa9d-4c4c-812f-cf38fa148de7

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -0,0 +1,229 @@
...
+    output_dir: Path
+    """Path to the output directory."""
+
+    def main(self) -> None:
+        """Generate a nerfstudio dataset from ProjectAria data (VRS) and MPS attachments."""
+        # Create output directory if it doesn't exist.
+        self.output_dir = self.output_dir.absolute()
+        self.output_dir.mkdir(parents=True, exist_ok=True)
+
+        provider = create_vrs_data_provider(self.vrs_file.absolute().as_posix())
```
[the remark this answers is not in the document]
sweeneychris: I also don't have a windows machine to test this either! I don't think project aria tools supports windows but I could be wrong about that...

brentyi (approved):

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -0,0 +1,231 @@
...
+        aria_frames = [
+            to_aria_image_frame(provider, index, name_to_camera, t_world_devices, self.output_dir)
+            for index in range(0, provider.get_num_data(stream_id))
+        ]
+
+        # create the NerfStudio frames from the AriaImageFrames.
+        print("Creating NerfStudio frames...")
+        CANONICAL_RGB_VALID_RADIUS = 707.5
+        CANONICAL_RGB_WIDTH = 1408
+        rgb_valid_radius = CANONICAL_RGB_VALID_RADIUS * (aria_frames[0].camera.width / CANONICAL_RGB_WIDTH)
```
AntonioMacaronio, line 215: How did you get these numbers? On the project aria github linked [here](https://github.com/facebookresearch/projectaria_tools/blob/4aee633cb667ab927825dc10477cad0df8393a34/core/calibration/loader/SensorCalibrationJson.cpp#L97C1-L99C19), I see different numbers; Is there a reason why?

brentyi: Do you mean the width? This is just from different recording profiles; the RGB camera can be configured to record at either 1408x1408 (2MP) or 2880x2880 (8MP).

AntonioMacaronio: Gotcha! that makes sense why the comment says "(full res during calibration)"

PR #2623, opened 2023-11-19
Fix dycheck depth and rgb rescale

### LINKED ISSUE #2619: Dycheck Dataparser rescale shape error
**Describe the bug**
When using downscaling with the dycheck dataparser, the automatic downscaling has a shape error between `depth_raw` and `cosa` [here](https://github.com/nerfstudio-project/nerfstudio/blob/9528a3f717fe28a6c6500199b5b37d4f2c6e10ca/nerfstudio/data/dataparsers/dycheck_dataparser.py#L176). Could possibly only exist for record3d-captured scenes?

**To Reproduce**
Use dycheck dataparser  with downscale_factor > 1 for a scaling that does not already exist. I used the backpack scene from [here](https://drive.google.com/drive/folders/1cBw3CUKu2sWQfc_1LbFZGbpdQyTFzDEX) with a downscale factor of 2 and received a dimension mismatch error:
`ValueError: operands could not be broadcast together with shapes (480,360,1) (480,360)`

**Expected behavior**
There should be no shape error.

**Additional context**
Adding an extra dimension to `depth_raw` if it is not 3-dimensional fixes the issue. Does anyone know what the extra dimension is for?

origamiman72:
Not sure if depth is correct since I'm not sure what the 3rd dim is for, the code runs properly though. Resolves #2619

### DIFF
--- a/nerfstudio/data/dataparsers/dycheck_dataparser.py
+++ b/nerfstudio/data/dataparsers/dycheck_dataparser.py
@@ -173,7 +173,10 @@ def _rescale_depth(depth_raw: np.ndarray, cam: Dict) -> np.ndarray:
     viewdirs /= np.linalg.norm(viewdirs, axis=-1, keepdims=True)
     viewdirs = viewdirs.reshape((*batch_shape, 3))
     cosa = viewdirs @ (cam["camera_to_worlds"][:, 2])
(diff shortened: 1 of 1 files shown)


lsongx (approved):
LGTM

Issue #2625, opened 2023-11-21
Metashape import: camera groups not handled


f-dy:
**Describe the bug**
If the Metashape project contains one or several camera groups, ns-process-data doesn't find the cameras in these groups.

**To Reproduce**
Use the file `cameras_pose.xml` from the following comment:
https://github.com/nerfstudio-project/nerfstudio/issues/1634#issuecomment-1820800783

Workaround:
remove `<group ...>` and `</group>` tags from the xml before processing

**Expected behavior**
Groups should be ignored

**Additional context**
I will submit a PR to fix this

PR #2626, opened 2023-11-22
metashape_utils: work with camera groups

f-dy:
Metashape XML parsing was broken when it contained camera groups.
I changed the way it iterates over XML elements to iterate over elements of a given type (camera) rather than on all children of the selected node.
 
Fixes https://github.com/nerfstudio-project/nerfstudio/issues/2625


jb-ye (approved):
simple fix, lgtm.

PR #2627, opened 2023-11-22
Fix tangential camera distortion (p1,p2 in opencv camera)

jkulhanek:
Tangential camera distortion parameters have flipped y in the current implementation. This is because OpenCV camera model assumes opencv coords, but the coords are passed as opengl. This PR moves the switch from opencv to opengl after the distortion.


tancik (approved):
Nice find!

PR #2630, opened 2023-11-25
Converted many @dataclass assignments for python 3.11 compatibility

jaggzh:
Converted many @dataclass assignments to field()s for python 3.11 compatibility
There are probably more to go, as I only tested this with ns-train nerfacto.


brentyi (approved):
Thanks!!

PR #2635, opened 2023-11-29
Make nuScenes dataparser compatible with ParallelDataManager

dmholtz:
The ParallelDataManager (see #2092) makes pytorch crash if the cameras instance's fx, fy, cx or cy tensors are loaded from a common shared tensor.

This PR fixes the issue by cloning the respective tensors before passing them to the `Cameras(...)` constructor.

### DIFF
--- a/nerfstudio/data/dataparsers/nuscenes_dataparser.py
+++ b/nerfstudio/data/dataparsers/nuscenes_dataparser.py
@@ -200,10 +200,10 @@ class NuScenes(DataParser):
         )
 
         cameras = Cameras(
-            fx=intrinsics[:, 0, 0],
-            fy=intrinsics[:, 1, 1],
-            cx=intrinsics[:, 0, 2],
-            cy=intrinsics[:, 1, 2],
+            fx=intrinsics[:, 0, 0].detach().clone(),
+            fy=intrinsics[:, 1, 1].detach().clone(),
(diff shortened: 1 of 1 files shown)


nerfstudio/data/dataparsers/nuscenes_dataparser.py
```diff
@@ -200,10 +200,10 @@ def _generate_dataparser_outputs(self, split="train"):
...
-            fy=intrinsics[:, 1, 1],
-            cx=intrinsics[:, 0, 2],
-            cy=intrinsics[:, 1, 2],
+            fx=intrinsics[:, 0, 0].detach().clone(),
+            fy=intrinsics[:, 1, 1].detach().clone(),
+            cx=intrinsics[:, 0, 2].detach().clone(),
+            cy=intrinsics[:, 1, 2].detach().clone(),
             height=900,
             width=1600,
             camera_to_worlds=poses[:, :3, :4],
```
kerrj, line 209: should we also do this on the transform matrices?

dmholtz: No, this is not necessary, because the data in `poses` is not shared across multiple parameters as with fx, fy, cx, cy of the tensor `intrinsics`.

With my fix applied, `ns-train nerfacto nuscenes-data --data ... --data_dir ...` runs as expected.

PR #2638, opened 2023-11-29
Keyframe timing for beta viewer

brentyi:
I cleaned up the spline rendering a little bit[^1] and added spherical [^2] handles for transitions, which you can click on to adjust timing between keyframes.

Here's an example workflow, where we make the transition between a pair of keyframes faster:

https://github.com/nerfstudio-project/nerfstudio/assets/6992947/c7d06cc5-ae2b-4b20-9c2e-6ae5d7d67372

As an aside: the render panel code is adapted from research code and really chaotic. I spent a little bit of time trying to refactor and couldn't come up with something I was super happy with; I'd like to revisit this from viser in a "recommended usage patterns" kind of way.

[^1]: via https://github.com/nerfstudio-project/viser/pull/77, thanks @origamiman72!
[^2]: via https://github.com/nerfstudio-project/viser/pull/132, thanks @pritzza!

### DIFF
--- a/nerfstudio/viewer_beta/render_panel.py
+++ b/nerfstudio/viewer_beta/render_panel.py
@@ -21,7 +21,7 @@ import json
 import threading
 import time
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


kerrj (approved):
nice

PR #2642, opened 2023-11-30
fix include_input for NeRFEncoding

blacksino:
NeRFEncoding won't include input if using tcnn,got fixed


brentyi (approved):
Thanks, looks good to me!

PR #2647, opened 2023-12-04
Mask flexibility for methods than need to sample in the masked region

ethanweber:
I'm making minor changes to make the masking feature more flexible. I have a method where I want to sample everywhere in the image, including the masked out regions, as specified in the `transforms.json` file. My custom model will compute losses differently depending on where it is masked out.

Furthermore, I sometimes want to give the masked out region a specific color. A minor modification to the dataloading and dataset enables this.

Lastly, I expose the trainer to the callback because my pipeline needs access to this, specifically to read off the `base_dir` so I can avoid wandb.


kerrj (approved):
sgtm

PR #2651, opened 2023-12-07
Ns-process-data bug with dev version of colmap

AdamRashid96:
When you build colmap from source you get the '3.9-dev' version and the old code would try to turn it to a float to figure out the version number and crash. I changed it to only parse digits and decimals.


tancik (approved):
LGTM

PR #2652, opened 2023-12-07
added Intel GPU support for training and inference

Jing1Ling:
Added 'xpu' device type, which represent Intel GPU.

Using Intel GPU for training and inference requires configuring oneAPI and IPEX.
oneAPI: https://www.intel.com/content/www/us/en/developer/tools/oneapi/base-toolkit-download.html?operatingsystem=linux&distributions=offline
IPEX: https://github.com/intel/intel-extension-for-pytorch
 
Example commands for running nerfstudio on XPU device:
```
# train
ns-train nerfacto --data data/nerfstudio/poster --machine.device-type xpu
# eval
ns-eval  --load-config=outputs/poster/nerfacto/exp_folder/config.yml
# render
ns-render dataset --load-config outputs/poster/nerfacto/exp_folder/config.yml
```

### DIFF
--- a/nerfstudio/configs/base_config.py
+++ b/nerfstudio/configs/base_config.py
@@ -69,7 +69,7 @@ class MachineConfig(PrintableConfig):
     """current machine's rank (for DDP)"""
     dist_url: str = "auto"
     """distributed connection point (for DDP)"""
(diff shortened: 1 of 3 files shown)


jdeinlein:
Thanks! Works perfectly with my A770, but I had to downgrade numpy to <2.0 to make it work.

PR #1008, opened 2022-11-23
[WIP] dreamfusion implementation 

terrancewang:

PR #2653, opened 2023-12-08
Trainer can now always save dataparser transform

pwais:
The base_dataparser now has a base `save_dataparser_transform()` method https://github.com/nerfstudio-project/nerfstudio/blob/64f0b2547ba02b1e2aeef719de95b4e7b1aa188a/nerfstudio/data/dataparsers/base_dataparser.py#L77

I think this might have just been a clean-up item that either didn't get addressed in the earlier PR https://github.com/nerfstudio-project/nerfstudio/pull/1008/files#r1207121018

or likely something else got refactored and this got missed.   The current guard is too restrictive now.  If we want to maintain a guard, probably better to do a`hasattr` on `self.pipeline.datamanager.train_dataparser_outputs` for duck-typing.  

Not sure if there are unit tests that can look for this file?

### DIFF
--- a/nerfstudio/engine/trainer.py
+++ b/nerfstudio/engine/trainer.py
@@ -225,11 +225,9 @@ class Trainer:
         """Train the model."""
         assert self.pipeline.datamanager.train_dataset is not None, "Missing DatsetInputs"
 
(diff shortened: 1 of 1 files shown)


brentyi (approved):
This fix looks correct to me, thanks!!

PR #2531, opened 2023-10-16
Update hloc and add new matcher types

Michael-Spleenlab:

PR #2658, opened 2023-12-08
Update hloc and add new matcher types

alexandrukis:
This PR is just https://github.com/nerfstudio-project/nerfstudio/pull/2531/files, so all credits go to @Michael-Spleenlab, but it also includes the fixed Dockerfile comment, so we can merge this faster


tancik (approved):
LGTM

PR #2661, opened 2023-12-09
Update cameras.py - removed duplicate lines

IamMohitM:
The self.times was being computed twice in cameras.py in the same function which was unnecessary


akristoffersen (approved):
LGTM

PR #2667, opened 2023-12-11
Improved render panel for beta viewer

brentyi:
**(1) Continuos velocity trajectories**

Previously, velocities in the beta viewer were piecewise linear. I think this was because I misinterpreted some of the original viewer code.

This new implementation produces nicer splines with continuous velocities:
image

**(2) Better render preview**

We now support previewing fisheye + equirectangular renders, hide threejs elements during previews, and respect the render aspect ratio.

https://github.com/nerfstudio-project/nerfstudio/assets/6992947/178ff7ab-8f7f-48f4-b980-a6afc48fd3eb


tancik (approved):
LGTM

PR #2678, opened 2023-12-14
[Fix] Pixel Sampler setup in VanillaDataManager for Cameras incl. metadata

tobiasfshr:
This PR fixes the pixel sampler setup in the vanilla datamanager for cameras with metadata other than `fisheye_crop_radius`, i.e. it is not correct to assume that if cameras have metadata, the key `fisheye_crop_radius` will be available in all cases. Therefore, I added a check to the if condition.


tancik (approved):
Thanks for fixing!

maturk (approved):
LGTM

PR #2683, opened 2023-12-16
[Full images datamanager] fix image indexing with cv.undistort

maturk:
From my understanding in `newK, roi = cv2.getOptimalNewCameraMatrix(K, distortion_params, (image.shape[1], image.shape[0]), 0)` the roi that is returned is in pixel notations `(px_start, py_start, px_end, py_end)`

When slicing an image we should instead do `image[y : y + h + 1, x : x + w + 1]` to preserve the same shape as the input image. Otherwise we crop the last pixel and output_image.shape != input_image.shape :D

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -22,6 +22,7 @@ paradigm
 from __future__ import annotations
 
 import random
+from copy import deepcopy
 from dataclasses import dataclass, field
 from functools import cached_property
 from pathlib import Path
@@ -30,7 +31,6 @@ from typing import Dict, ForwardRef, Generic, List, Literal, Optional, Tuple, Ty
 import cv2
 import numpy as np
 import torch
-from copy import deepcopy
 from torch.nn import Parameter
 from tqdm import tqdm
 
@@ -150,19 +150,19 @@ class FullImageDatamanager(DataManager, Generic[TDataset]):
                 image = cv2.undistort(image, K, distortion_params, None, newK)  # type: ignore
                 # crop the image and update the intrinsics accordingly
                 x, y, w, h = roi
-                image = image[y : y + h, x : x + w]
+                image = image[y : y + h + 1, x : x + w + 1]
                 if "mask" in data:
-                    data["mask"] = data["mask"][y : y + h, x : x + w]
+                    data["mask"] = data["mask"][y : y + h + 1, x : x + w + 1]
                 if "depth_image" in data:
-                    data["depth_image"] = data["depth_image"][y : y + h, x : x + w]
+                    data["depth_image"] = data["depth_image"][y : y + h + 1, x : x + w + 1]
                 # update the width, height
-                self.train_dataset.cameras.width[i] = w
-                self.train_dataset.cameras.height[i] = h
+                self.train_dataset.cameras.width[i] = w + 1
+                self.train_dataset.cameras.height[i] = h + 1
                 if "mask" in data:
                     mask = data["mask"].numpy()
                     mask = mask.astype(np.uint8) * 255
                     mask = cv2.undistort(mask, K, distortion_params, None, newK)  # type: ignore
-                    mask = mask[y : y + h, x : x + w]
+                    mask = mask[y : y + h + 1, x : x + w + 1]
                     data["mask"] = torch.from_numpy(mask).bool()
                 K = newK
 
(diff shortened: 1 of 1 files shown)


nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -150,19 +164,19 @@ def cache_images(self, cache_images_option):
                 image = cv2.undistort(image, K, distortion_params, None, newK)  # type: ignore
                 # crop the image and update the intrinsics accordingly
                 x, y, w, h = roi
-                image = image[y : y + h, x : x + w]
+                image = image[y : y + h + 1, x : x + w + 1]
```
maturk, line 167: this should preserve 1) input.shape == output.shape

maturk:
also added option for mipnerf360 style evaluation in colmap dataparser to hold out every n'th image in train set for eval dataset. Useful for comparing with inria eval scripts

maturk:
Reverted sampling strategies in colmap dataparser in favour of the new PR #2709 which also implements them. This PR fixes 1) the output image size bug after cv.undistort and 2) implements the logic for "eval_image_indices" and "eval_num_images_to_sample_from" in the datamanager config.

kerrj:
How does this PR interact with #2726 ?

liorfritz:
isn't it OK that after the undistortion the shape of the image changes?
Maybe the fact that in Mip-NeRF 360 the images are already undistorted, and the second undistortion caused a loss of pixel (prior to PR #2726), lead to this change?

jb-ye:
@maturk For the first issue: do you have sources indicating the roi is in (left, upper, right, lower) convention? I think it is in x, y, w, h convention.

f-dy:
The fix should be done upstream: w and h returned by the latest version of icvGetRectangles are off by 1 pixel, see https://github.com/opencv/opencv/issues/24831

maturk:
@f-dy okay wow this is actually pretty crazy XD Never would I have thought that I would find a bug in OpenCV! Thank you for making the issue and even tracking down the problem in the source code.

f-dy:
I would recommend adding a unit test to test_datamanager.py, so that we're sure to get the thing right in the end.

f-dy:
And btw we should probably not use cv.undistort, which uses bilinear interpolation, and write our own function that calls cv2.remap with INTER_CUBIC instead. bilinear interpolation creates uneven blur in the image, especially for small distortion amounts, i.e with a small k1 you would see rings of blur in the undistorted images. I'll add some code that demonstrates that

kerrj:
So with opencv 4.8+ this is no longer an issue?

liorfritz:
Still an issue, even with 4.9.
Hopefully will be patched in the future, following the issue that @f-dy opened.

maturk:
Any suggestions what we should do with this issue with the cropped pixel?

jb-ye:
@maturk Let's wait opencv fix this bug, I believe they will fix it very soon. Once they fix the issue, we can bump up opencv version dependency again.

f-dy:
> Any suggestions what we should do with this issue with the cropped pixel?

gate the fix you did with if 4.8<=opencv_version<4.10, since the issue is marked for OpenCV 4.10, and add a unit test, so that we can adjust the gate if they don't fix it in 4.10

SharkWipf:
OpenCV 4.10 is out now, and with that I'd like to add a small request: Can we loosen up the OpenCV dependency ever so slightly? 
Specifically, can we loosen up at least the 4th digit of the version requirement?
The reason being, Conda cannot select package versions down to the 4th digit and if you install COLMAP via Conda/-forge, it will pull in opencv-python automatically.
You can restrict the version via Conda, but not down to the 4th digit.
Pip, when installing Nerfstudio, cannot modify the opencv-python version installed by Conda, and because of how tight the version check is, it will fail to install.
The only way around this right now is to manually modify the pyproject.toml file to loosen the version check, but this of course only works for installations from git, not from pypi.
So basically: If we could loosen up the version constraints for opencv-python at least to the point where it doesn't care for the 4th version digit (though ideally even looser), this would reduce conda woes quite a bit and simplify installation requirements.

maturk:
> OpenCV 4.10 is out now, and with that I'd like to add a small request: Can we loosen up the OpenCV dependency ever so slightly? Specifically, can we loosen up at least the 4th digit of the version requirement? The reason being, Conda cannot select package versions down to the 4th digit and if you install COLMAP via Conda/-forge, it will pull in opencv-python automatically. You can restrict the version via Conda, but not down to the 4th digit. Pip, when installing Nerfstudio, cannot modify the opencv-python version installed by Conda, and because of how tight the version check is, it will fail to install. The only way around this right now is to manually modify the pyproject.toml file to loosen the version check, but this of course only works for installations from git, not from pypi. So basically: If we could loosen up the version constraints for opencv-python at least to the point where it doesn't care for the 4th version digit (though ideally even looser), this would reduce conda woes quite a bit and simplify installation requirements.

Yeah sounds good to me @SharkWipf , can you do a PR?

f-dy:
> > And btw we should probably not use cv.undistort, which uses bilinear interpolation, and write our own function that calls cv2.remap with INTER_CUBIC instead. bilinear interpolation creates uneven blur in the image, especially for small distortion amounts, i.e with a small k1 you would see rings of blur in the undistorted images. I'll add some code that demonstrates that
> 
> I add my point of view to this, this undistorts, RGBA images with cv2.undistort, for masks the interpolation should be nearest. Kindly add these functionalities

The effect of bilinear vs bicubic should be barely visible on high-resolution images, because the PSF of most cameras is larger than one pixel anyway. I agree that it would be a nice to have. Be aware that `cv2.initUndistortRectifyMap` followed by `cv2.remap` uses more memory than `cv2.undistort` (which works line-by-line).

One should make a PR to OpenCV to add an "interpolation" parameter to cv::undistort that would be passed on to cv::remap

PR #2687, opened 2023-12-18
Fix a bug of random quaternions

Zhuoyang-Pan:
I found a bug of random quaternions generation in gaussian-splatting model. This is used when initializing the gaussians.


tancik (approved):
Good catch

PR #2692, opened 2023-12-19
fixed not cache images for full image datamanager

hardikdava:
- skipping cache image for `no-cache` argument


maturk:
@hardikdava, actually, right now the "no-cache" option should not exist and the next_train crashes if "no-cache" is enabled. I think it is better to remove it completely and only have cpu and gpu options.

kerrj (approved):
lgtm

PR #2700, opened 2023-12-23
Setting save_world_frame as False by default and Splat export panel in viewer

cvachha:
- Setting save_world_frame as False by default for the exporter as well as in the viewer, since it is exports the point cloud in a different coordinate system from the Blender add-on and there is not a similar checkbox in the original viewer, so by default the exported point cloud would not be compatible. This flag can still be set true in the cmd interface and in the new viewer.

- Updating the blender docs to mention support for gs scenes.

- Adding Splat ply export panel in the viewer
splat_panel
splat_modal


nerfstudio/viewer/export_panel.py
```diff
@@ -34,9 +34,11 @@ def _(_) -> None:
         populate_point_cloud_tab(server, control_panel, config_path)
     with server.add_gui_folder("Mesh"):
         populate_mesh_tab(server, control_panel, config_path)
+    with server.add_gui_folder("Splat"):
```
brentyi, line 37: @cvachha can we make this only appear when a splat model is loaded?

since this operation is fairly common it could also be moved above the point cloud / mesh exports I think!

cvachha:
Export panel when running gaussian-splatting models: 
gsplat_export_panel

Export panel when running nerfacto/other models:
nerfacto_export_panel

brentyi (approved):

PR #2704, opened 2023-12-25
fix the bugs associated with blender datas

KevinXu02:
fix the bug when camera.distortion_params is None, which will lead to no data when using blender data.


KevinXu02:
I tested it with lego and the camera convention is not correct now......

nerfstudio/model_components/renderers.py
```diff
@@ -50,8 +50,7 @@ def background_color_override_context(mode: Float[Tensor, "3"]) -> Generator[Non
     old_background_color = BACKGROUND_COLOR_OVERRIDE
     try:
         BACKGROUND_COLOR_OVERRIDE = mode
-        yield
-    finally:
```
KevinXu02, line 54: this finally will lead to the overrided colour be overrided.

nerfstudio/data/dataparsers/blender_dataparser.py
```diff
@@ -61,13 +62,13 @@ def __init__(self, config: BlenderDataParserConfig):
         self.data: Path = config.data
         self.scale_factor: float = config.scale_factor
         self.alpha_color = config.alpha_color
-
-    def _generate_dataparser_outputs(self, split="train"):
         if self.alpha_color is not None:
-            alpha_color_tensor = get_color(self.alpha_color)
+            self.alpha_color_tensor = get_color(self.alpha_color)
+            background_color_override_context(self.alpha_color_tensor)
```
KevinXu02, line 67: override bg colour

nerfstudio/models/gaussian_splatting.py
```diff
@@ -533,14 +533,14 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
         if self.training:
             # currently relies on the branch vickie/camera-grads
             self.camera_optimizer.apply_to_camera(camera)
-        if self.training:
+        # get the background color
```
KevinXu02, line 536: modified logic to make override on top level

brentyi (approved):

KevinXu02:
It seems that my ruff formatter keeps messing everything up, but the background should work now. When using blender data the background_color should be changed to white to avoid cloulds. In the viser the rendered bg color is always overwritten to grey, it will not affect training, but I wonder if there is some way to modify it without using overwrite.

danperazzo:
Hi everyone! How can I change the background color to white? I tried doing this, but I kept getting some strange visualizations

PR #2713, opened 2023-12-30
Fix periodically stuck in parallel datamanager

liruilong940607:
Fix the issue of training periodically stuck at https://github.com/nerfstudio-project/nerfstudio/blob/a8e6f8fa3fd6c0ad2f3e681dcf1519e74ad2230f/nerfstudio/data/datamanagers/parallel_datamanager.py#L280

```
ns-train nerfacto --data data/nerfstudio/poster --max-num-iterations 200
```
Screenshot 2023-12-30 at 1 39 04 PM

**Reason**: `wait()` will be triggered in `.result()` due to the incomplete async put: `self.out_queue.put_nowait((ray_bundle, batch))`.

**Fix**: Change the put from async to sync. Don't need `ThreadPoolExecutor` anymore to acquire data from the queue.

### DIFF
--- a/nerfstudio/data/datamanagers/parallel_datamanager.py
+++ b/nerfstudio/data/datamanagers/parallel_datamanager.py
@@ -120,7 +120,7 @@ class DataProcessor(mp.Process):
                 ray_bundle = ray_bundle.pin_memory()
             while True:
                 try:
-                    self.out_queue.put_nowait((ray_bundle, batch))
+                    self.out_queue.put((ray_bundle, batch))
                     break
                 except queue.Full:
                     time.sleep(0.0001)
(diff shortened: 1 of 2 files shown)


maturk:
Interesting, nice catch

liruilong940607:
Seems only solve partial of the issue ... WIP

liruilong940607:
Identify data queue `.get()` can also sometimes get stuck when unpickle the data sent from the subprocess.

Switch from `import torch.multiprocessing as mp` to `from pathos.helpers import mp` completely gets rid  of the issue, which comes with enhanced serialization.

Screenshot 2023-12-30 at 4 03 52 PM

kerrj (approved):
Looks good! Thanks

PR #2723, opened 2024-01-04
Support multiple camera modes

alexandrukis:
Mostly the work from https://github.com/nerfstudio-project/nerfstudio/pull/2575, but made `ns-process-data` configurable too.

I am mostly trying to achieve two things:

* Check if the quality of the final model can be improved when a camera with auto-focus is used (see https://discord.com/channels/1025504971606724712/1192374369062760479)
* Check if a combination of both a wide and an ultra-wide camera can be used to improve the final model (SfM yields better results with ultra-wide cameras, but these images are usually of lower quality)

For both these points, a starting point would be to support more than one camera when reconstructing, thus this PR.

### DIFF
--- a/nerfstudio/process_data/colmap_converter_to_nerfstudio_dataset.py
+++ b/nerfstudio/process_data/colmap_converter_to_nerfstudio_dataset.py
@@ -31,6 +31,8 @@ class ColmapConverterToNerfstudioDataset(BaseConverterToNerfstudioDataset):
 
     camera_type: Literal["perspective", "fisheye", "equirectangular"] = "perspective"
     """Camera model to use."""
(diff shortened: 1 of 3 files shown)


jb-ye:
Look forward to merging this feature. It would be nice if you can share some results demonstrating the benefits on using dataset captured by two or more cameras, as a sanity check for validating the change.

alexandrukis:
> It would be nice if you can share some results demonstrating the benefits on using dataset captured by two or more cameras

What I am trying to figure out is actually if an auto-focusing camera should be treated as multiple cameras (one per image, depending on the focal distance), so this is what I have used this change for. I don't really have a dataset captured using multiple cameras.

jb-ye:
close given #3081

PR #2709, opened 2023-12-29
Improvement to Gaussian Splatting workflow

jb-ye:
Hi @kerrj 

As @f-dy discussed with you earlier, this PR is an improvement to the gaussian splatting workflow implemented in nerfstudio. It has two major changes as detailed below. Please feel free to comment. 

## Enhancement to the gaussian splatting training. 

This PR has more aggressive gaussian growth strategy. Based on our internal evaluation, this PR makes the nerfstudio-gs implementation outperforms the official Inria implementation by a little bit on average. In general, we observe the current PR improves more on indoor datasets.

I also experimented three public datasets from mipnerf360 dataset (bicycle, garden, kitchen) and reported the results below:

Nerfstudio-gs Before / After applying this PR
  | psnr | ssim | lpips
-- | -- | -- | --
bicycle | 25.84/26.15 | 0.771/0.790 | 0.179/0.147
garden | 28.38/28.63 | 0.875/0.887 | 0.090/0.070
kitchen | 31.35/32.24 | 0.929/0.941 | 0.096/0.076

For reference, I also cited results from mipnerf360 / gaussian splatting papers:

  | psnr | ssim | lpips
-- | -- | -- | --
bicycle | 24.37/25.25 | 0.685/0.771 | 0.301/0.205
garden | 26.98/27.41 | 0.813/0.868 | 0.170/0.103
kitchen | 32.23/30.317 | 0.920/0.922 | 0.127/0.129

## PLY exporter 

I made changes to allow the PLY exporter produce PLY formatted data compatible with the official Inria version. After export the PLY file from saved checkpoint, one can use the exported file in a web viewer that was originally designed for Inria code.

(_DELETED IN LATEST REVISION_:
As part of the implementation, I also implemented an independent PLY based image renderer without reading nerfstudio checkpoint. This is an effort to simplify potential workflow that are more separate from nerfstudio interface.)

PR #2726, opened 2024-01-04
Undistortion should not occur for already-undistorted images

liorfritz:
_This mostly concerns public datasets, like Mip-NeRF 360, which already comes undistorted._

The current logic undistorts even pinhole cameras. To avoid clashing with other logic in the code, we can just skip the undistortion if all distortion parameters are zero.
Unnecessary undistortion smoothes the images, and may incorrectly create the impression that image comparison metrics (PSNR, SSIM, LPIPS) are better. The PR in #2709 shows improvements in the implementation through multiple modifications, that are complementary to this change.

Here's a comparison of what it does to an image from the training set of 'garden',
image

On the garden dataset, without the undistortion the metrics are affected as follows (keep in mind that the change applies to the evaluation images as well, hence the metrics are lower),

images (training and evaluation) | PSNR | SSIM | LPIPS
------ | ---| ---| ---
unnecessarily undistorted | 28.58 | 0.885 | 0.073
original | 27.29 | 0.858 | 0.084

While the metrics are **lower** after the change, they are **aligned** with other repositories (because the evaluation images are unnecessarily undistorted as well). Due to the added details, there are also more gaussians (5.1M gaussians without undistortion, compared to 3.8M with the unnecesary undistortion).

----

This change creates misalignment between how images are downscaled in _nerfstudio_ compared to other repositories (and in Mip-NeRF 360). Hence, the rounding of the downscale factor in `cameras.py`. Without it, in 'garden', the original resolution of `5187x3361` downscaled to 4, results in `1296x840`. While in the supplied images in the data, the resolution is `1297x840`. This occurs because `5187/4=1296.75`. The change in `cameras.py` makes it necessary to apply the same logic to all downscaling mechanisms in the code. This shouldn't change much for most users (e.g., HD resolution).


jb-ye:
Wow, this is an interesting finding.

nerfstudio/process_data/process_data_utils.py
```diff
@@ -462,7 +462,7 @@ def downscale_images(
                 nn_flag = "" if not nearest_neighbor else ":flags=neighbor"
                 ffmpeg_cmd = [
                     f'ffmpeg -y -noautorotate -i "{image_dir / filename}" ',
-                    f"-q:v 2 -vf scale=iw/{downscale_factor}:ih/{downscale_factor}{nn_flag} ",
+                    f"-q:v 2 -vf scale=round(iw/{downscale_factor}):round(ih/{downscale_factor}){nn_flag} ",
```
jb-ye, line 465: What's the reason to have `round` here? I believe the right approach is to ensure aspect ratio of the scaling, by having `setsar=1` append to the last. 

```
ffmpeg -y -noautorotate -i "{image_dir / filename}"  \
 -q:v 2 -vf scale=iw/{downscale_factor}:ih/{downscale_factor}{nn_flag},setsar=1
```

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -132,20 +132,20 @@ def cache_images(self, cache_images_option):
...
+                    distortion_params[4],
+                    distortion_params[5],
+                    distortion_params[2],
+                    distortion_params[3],
+                    0,
+                    0,
+                ]
+            )
+
+            if camera.camera_type.item() == CameraType.PERSPECTIVE.value and np.any(distortion_params):
```
jb-ye, line 148: This change wouldn't be compatible with other features (mask, depth_image). I would suggest to check `np.any(distortion_params)` and if False, return the original K and image without triggering cv2 functions. Something like below should work

```
if np.any(distortion_params):
    newK, roi = cv2.getOptimalNewCameraMatrix(K, distortion_params, (image.shape[1], image.shape[0]), 0)
    image = cv2.undistort(image, K, distortion_params, None, newK)  # type: ignore
else:
    newK = K
    roi = 0, 0, image.width, image.height
```

jb-ye: BTW, I found the following lines are problematic, roi was calculated for undistorted images, not the original image/mask. Should be deleted maybe. 

```
                if "mask" in data:
                    data["mask"] = data["mask"][y : y + h, x : x + w]
                if "depth_image" in data:
                    data["depth_image"] = data["depth_image"][y : y + h, x : x + w]
```

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -206,20 +203,20 @@ def cache_images(self, cache_images_option):
...
```
jb-ye, line 219: See my comments before.

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -254,8 +251,7 @@ def cache_images(self, cache_images_option):
                     mask = cv2.fisheye.undistortImage(mask, K, distortion_params, None, newK)
                     data["mask"] = torch.from_numpy(mask).bool()
                 K = newK
-            else:
-                raise NotImplementedError("Only perspective and fisheye cameras are supported")
```
jb-ye, line 258: We may still want to keep this behavior per earlier comment.

liorfritz: Thank you! Indeed, I removed it for some testing and haven't noticed. Put it back.

nerfstudio/data/datamanagers/full_images_datamanager.py
liorfritz: Great catch. Fixed.
Per the other comment - I removed the lines for the mask, because indeed it occurs twice in the code. I left the depth as is.

nerfstudio/data/dataparsers/colmap_dataparser.py
```diff
@@ -434,7 +434,7 @@ def _downscale_images(self, paths, get_fname, downscale_factor: int, nearest_nei
                 path_out.parent.mkdir(parents=True, exist_ok=True)
                 ffmpeg_cmd = [
                     f'ffmpeg -y -noautorotate -i "{path}" ',
-                    f"-q:v 2 -vf scale=iw/{downscale_factor}:ih/{downscale_factor}{nn_flag} ",
+                    f"-q:v 2 -vf scale=round(iw/{downscale_factor}):round(ih/{downscale_factor}){nn_flag} ",
```
jb-ye, line 437: ffmpeg seems doesn't support `round` function here. I am getting errors by running this command.

The bigger concern here is whether the rescaled image by ffmpeg has the exact same resolution as specified by `camera.rescale_output_resolution()`

I did some experiment locally and found that it could be quite tricky.

nerfstudio/process_data/process_data_utils.py
liorfritz: Reverted change. The `cameras.py` rescaling now accepts the height and width, so no need for this entire hassle.

liorfritz:
Per the great comments from @jb-ye, we removed the problematic logic of image downscaling. Now, `cameras.py` receives the width/height of the image in either case, so that it works even with the odd resolution of Mip-NeRF 360.

In addition, just to be clear, this PR leads to better results. Here are the results on the 'garden' sample, evaluated on the original evaluation set (i.e., without duplicated undistortion), comparing training with the original way to this PR.

Training Code | PSNR | SSIM | LPIPS
-- | -- | -- | --
Original | 27.08 | 0.833 | 0.142
This PR | 27.29 | 0.858 | 0.084

image

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -132,27 +132,29 @@ def cache_images(self, cache_images_option):
...
+                    distortion_params[5],
+                    distortion_params[2],
+                    distortion_params[3],
+                    0,
+                    0,
+                ]
+            )
 
             if camera.camera_type.item() == CameraType.PERSPECTIVE.value:
-                distortion_params = np.array(
```
jb-ye, line 137: I think we should keep this distortion_params update inside PERSPECTIVE type if-clause, otherwise, it may affect the FISHEYE to read the correct order of parameters.

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -206,22 +209,26 @@ def cache_images(self, cache_images_option):
...
```
jb-ye, line 211: See my comments earlier.

nerfstudio/data/datamanagers/full_images_datamanager.py
kerrj: +1

elialshan: Good call. Thanks

devernay:
what causes the blurriness exactly? is newK returned by getOptimalNewCameraMatrix different than K?
In the following test (undistorting with zero distortion params), the undistorted image is the same as the input image:
```
import cv2
import numpy as np
from matplotlib import pyplot as plt

# zero distortion
distortion_params = np.array(
    [
        0,#k1
        0,#k2
        0,#p1
        0,#p2
        0,#k3
        0,#k4,
        0,#k5,
        0,#k6
    ]
)

# https://docs.opencv.org/4.x/d9/d0c/group__calib3d.html#ga7a6c4e032c97f03ba747966e6ad862b1
width=1920
height=1080
ffactor=1.6
# intrinsics with a non-centered principal point
K = np.array([[width*ffactor, 0, (width-1)*0.5+0.05],
              [0,width*ffactor, (height-1)*0.5-0.04],
              [0,0,1]])

newK, roi = cv2.getOptimalNewCameraMatrix(K, distortion_params, (width, height), 0)

print(np.linalg.norm(K-newK))

# create a checkerboard image
img = np.fromfunction(lambda i, j,k: 255*((i//64+j//64)%2), (height,width,3), dtype=np.uint8)

plt.imshow(img)
plt.title('checkerboard')
plt.show()

imgu = cv2.undistort(img, K, distortion_params, None, newK)

print(np.linalg.norm(img-imgu))
```
Outputs:
```
4.547473508864641e-13
0.0
```
What is different between this code snippet and what happens in nerfstudio?

devernay:
OK so maybe there are two problems.
Are the distortion parameters 0? Can you provide a standalon test code like the one above?
I submitted an OpenCV issue for the wrong ROI (https://github.com/opencv/opencv/issues/24831)

devernay:
Ok, so the bug newK != K is present in OpenCV 4.6 and 4.7, bug is fixed in 4.8 (which is the version I used above).
Here's the OpenCV fix: https://github.com/opencv/opencv/pull/23305

I suggest we:
- skip the undistortion if dist params are 0 (this PR) to avoid issues with past, present (roi bug) and future versions of OpenCV
- think about upgrading OpenCV in nerfstudio (4.6 is 1.5yr old, latest release is 4.9.0)

kerrj:
Thanks for tracking that down! Bumping opencv to 4.9 sounds good to me, would this also fix #2683 or is that a separate problem?

jb-ye:
> Thanks for tracking that down! Bumping opencv to 4.9 sounds good to me, would this also fix #2683 or is that a separate problem?

I think the ROI bug is not yet fixed even in the latest version, sadly. But as long as the input images are undistorted, the colmap loader should work as expected after this PR.

f-dy:
> would this also fix https://github.com/nerfstudio-project/nerfstudio/pull/2683 or is that a separate problem?
https://github.com/nerfstudio-project/nerfstudio/pull/2683 is not the right fix. This is an upstream opencv issue (https://github.com/opencv/opencv/issues/24831).

liorfritz:
Right. If the ROI issue with opencv is fixed, then #2683 will break.
I'm thinking as follows,
1. This PR makes the code compatible with older versions of opencv, and the one used in nerfstudio currently (4.6)
2. Update opencv to 4.8
3. Keep a close eye on [opencv/opencv#24831](https://github.com/opencv/opencv/issues/24831) and update opencv when the ROI bug is fixed

kerrj (approved):
thanks for the catch!

PR #2730, opened 2024-01-05
Add `visible` as kwarg to ViewerElements

chungmin99:
Allow `ViewerElement` visibility to be set upon initialization. 

Currently, *all* elements are always set to visible. 

However, it might be good to intentionally make elements invisible from the get-go, like the dropdown example below. This is not possible in the current code, because `ViewerElement`'s `gui_handle` is initialized *after* the pipeline. 

Corresponding code is also attached. 

```
self.viewer_0_selection = ViewerDropdown(name="Options", default_value="1", options=["1", "2"])
self.viewer_1_button_1 = ViewerButton(name="Button 1", visible=True,  cb_hook=lambda _: None)
self.viewer_1_button_2 = ViewerButton(name="Button 2", visible=False, cb_hook=lambda _: None)

# Change visibility of buttons based on dropdown value
def swap_visible_button(dropdown: ViewerDropdown):
    if dropdown.value == "1":
        self.viewer_1_button_1.set_visible(True)
        self.viewer_1_button_2.set_visible(False)
    else:
        self.viewer_1_button_1.set_visible(False)
        self.viewer_1_button_2.set_visible(True)
self.viewer_0_selection.cb_hook = swap_visible_button
```

https://github.com/nerfstudio-project/nerfstudio/assets/10284938/a4455999-489b-45ae-af02-ef955e95e21c

### DIFF
--- a/nerfstudio/viewer_beta/viewer_elements.py
+++ b/nerfstudio/viewer_beta/viewer_elements.py
@@ -368,7 +368,14 @@ class ViewerSlider(ViewerParameter[IntOrFloat], Generic[IntOrFloat]):
     def _create_gui_handle(self, viser_server: ViserServer) -> None:
         assert self.gui_handle is None, "gui_handle should be initialized once"
         self.gui_handle = viser_server.add_gui_slider(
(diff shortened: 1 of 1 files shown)


kerrj (approved):
lgtm

PR #2734, opened 2024-01-08
Add PyNeRF external method

hturki:
I wanted to add PyNeRF to the documentation as an external method - let me know if this PR works!

https://haithemturki.com/pynerf/

### DIFF
--- a/nerfstudio/configs/external_methods.py
+++ b/nerfstudio/configs/external_methods.py
@@ -152,6 +152,23 @@ To enable Instruct-GS2GS, you must install it first by running:
     )
 )
 
+# K-Planes
+external_methods.append(
+    ExternalMethod(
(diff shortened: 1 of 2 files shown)


nerfstudio/configs/external_methods.py
```diff
@@ -152,6 +152,29 @@ class ExternalMethod:
     )
 )
 
+# K-Planes
```
brentyi, line 155: currently it says K-Panes here and in the instruction below! otherwise lgtm :)

hturki: whoops :)

hturki: fixed now!

brentyi (approved):

Issue #2735, opened 2024-01-09
Cache images in byte instead of float32 in GS training/eval


jb-ye:
**Is your feature request related to a problem? Please describe.**
When training GS with large number of images (>1000), caching images consumes a lot memory even for CPU mem. It is because the current cached images are preprocessed as float32 tensors. See `InputDataset.get_image()` function. If those images are cached as bytes, we would be 4x more memory efficient and allow to run GS on lower end machines. 

**Describe the solution you'd like**
Add a feature to allow caching images in bytes and do `image / 255.` only fly during training.

**Describe alternatives you've considered**
No.

**Additional context**
for an image with 1600x1600x3, caching 1000 images in bytes consumes about 7.7GB, while caching them in float32 consumes 4 times more.

kerrj:
Sounds good to me, my only worry would be a slowdown having to cast every image back to float for training. It shouldn't be that expensive but would be good to measure before merging

ichsan2895:
> Sounds good to me, my only worry would be a slowdown having to cast every image back to float for training. It shouldn't be that expensive but would be good to measure before merging

Maybe we can trigger this function if the dataset is too Big?
When data is small, its using default float32. But, if they are potentially OOM CUDA caused by large dataset, it automatically cache the image by bytes.

IMO, Its better slowdown training (in large dataset) instead of OOM CUDA.

PR #2741, opened 2024-01-09
Add option of caching image in bytes rather than float32

jb-ye:
See also https://github.com/nerfstudio-project/nerfstudio/issues/2735

This PR also changes the default behavior of GS training to cache images in bytes. Other models in nerfstudio are not impacted (still caching in float32). There is no sensible slow-down in the training.

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -61,6 +61,8 @@ class FullImageDatamanagerConfig(DataManagerConfig):
     """Specifies the image indices to use during eval; if None, uses all."""
     cache_images: Literal["cpu", "gpu"] = "cpu"
     """Whether to cache images in memory. If "cpu", caches on cpu. If "gpu", caches on device."""
+    cache_images_type: Literal["uint8", "float32"] = "float32"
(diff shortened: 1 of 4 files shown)


jb-ye:
@kerrj Could you help review it?

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -61,6 +61,8 @@ class FullImageDatamanagerConfig(DataManagerConfig):
     """Specifies the image indices to use during eval; if None, uses all."""
     cache_images: Literal["cpu", "gpu"] = "cpu"
     """Whether to cache images in memory. If "cpu", caches on cpu. If "gpu", caches on device."""
+    cache_images_type: Literal["uint8", "float32"] = "uint8"
```
pwais, line 64: maybe keep the old `float32` so no surprises / keep existing behavior?  that said, nerfstudio currently does not even support bigger bit depth https://github.com/nerfstudio-project/nerfstudio/blob/55b7a2c8ef3d2cc8a7eb7e66c67ec0278057df72/nerfstudio/data/datasets/base_dataset.py#L72

jb-ye: You made a good point. I can update the PR to reflect the original behavior.

PR #2742, opened 2024-01-09
Add Mill 19 dataset

hturki:
This adds the [Mill 19](https://meganerf.cmusatyalab.org/#data) dataset to the ns-download-data script and converts it to the format that the nerfstudio data parser expects:

image

image

### DIFF
--- a/nerfstudio/scripts/downloads/download_data.py
+++ b/nerfstudio/scripts/downloads/download_data.py
@@ -15,6 +15,7 @@
 """Download datasets and specific captures from the datasets."""
 from __future__ import annotations
 
+import json
 import os
 import shutil
 import tarfile
@@ -24,7 +25,9 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Union
 
 import gdown
+import torch
 import tyro
+from nerfstudio.process_data import process_data_utils
 from typing_extensions import Annotated
 
 from nerfstudio.configs.base_config import PrintableConfig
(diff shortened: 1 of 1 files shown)


nerfstudio/scripts/downloads/download_data.py
```diff
@@ -453,6 +456,96 @@ def download(self, save_dir: Path):
...
+    Mill19CaptureName = tyro.extras.literal_type_from_choices(mill19_downloads.keys())
+
+
+@dataclass
+class Mill19Download(DatasetDownload):
+    """Download the Mill 19 dataset."""
+
+    capture_name: Mill19CaptureName = "building"
+
+    def download(self, save_dir: Path):
```
brentyi, line 477: (nit) this could use a return annotation `-> None`

nerfstudio/scripts/downloads/download_data.py
```diff
@@ -453,6 +456,96 @@ def download(self, save_dir: Path):
...
+                if capture_name != "all":
+                    Mill19Download(capture_name=capture_name).download(save_dir)
+            return
+
+        assert (
+            self.capture_name in mill19_downloads
+        ), f"Capture name {self.capture_name} not found in {mill19_downloads.keys()}"
+        url = mill19_downloads[self.capture_name]
+        target_path = str(save_dir / f"mill19/{self.capture_name}")
+        os.makedirs(target_path, exist_ok=True)
```
brentyi, line 492: can we use the `pathlib` API for these operations instead? like:
```
target_path = save_dir / ...
target_path.mkdir(exist_ok=True)
```

same for replacing `os.listdir()`, `os.path.join()`, etc below

nerfstudio/scripts/downloads/download_data.py
```diff
@@ -453,6 +456,96 @@ def download(self, save_dir: Path):
...
+                        "cx": metadata["intrinsics"][2].item(),
+                        "cy": metadata["intrinsics"][3].item(),
+                        "w": metadata["W"],
+                        "h": metadata["H"],
+                        "transform_matrix": c2w.tolist(),
+                    }
+                )
+
+        with (target_path / "transforms.json").open("w") as f:
+            json.dump({"frames": frames}, f, indent=4)
```
brentyi, line 543: instead of a context maybe just `(target_path / "transforms.json").write_text(json.dumps(...))`?

brentyi (commented):
just nits 🙂

brentyi (approved):

PR #2745, opened 2024-01-10
Have Mill19 dataset use official splits by default

hturki:
This is a follow-up to https://github.com/nerfstudio-project/nerfstudio/pull/2742 - I wasn't aware of this option: https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/dataparsers/nerfstudio_dataparser.py#L195


brentyi (approved):
sorry for the delay here; lgtm!

PR #2749, opened 2024-01-11
Remove unused make_share_url arg in ns-viewer

brentyi:
Looks like this argument is both redundant (there's both `--make-share-url` and `--viewer.make-share-url`) and ignored (immediately overwritten)!


tancik (approved):
Lgtm

PR #2750, opened 2024-01-12
Update dependencies for aarch64

ymtoo:
Installing nerfstudio on NVIDIA Jetson throws an error:
```
ERROR: Could not find a version that satisfies the requirement pymeshlab>=2022.2.post2; 
platform_machine != "arm64" (from nerfstudio) (from versions: none)
```

This PR fixes the issue, enabling installation of nerfstudio on AArch64 machines.


brentyi (approved):
lgtm, thanks!

PR #2753, opened 2024-01-12
Remove deprecated dependency_links config

ringohoffman:
~~Move `dependency_links` to `setup.cfg`~~
    
`[options]` is not a valid table in `pyproject.toml`: https://packaging.python.org/en/latest/specifications/pyproject-toml/

but it is in `setup.cfg`: https://setuptools.pypa.io/en/latest/deprecated/dependency_links.html

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,12 +130,6 @@ ns-export = "nerfstudio.scripts.exporter:entrypoint"
 ns-dev-test = "nerfstudio.scripts.github.run_actions:entrypoint"
 ns-dev-sync-viser-message-defs = "nerfstudio.scripts.viewer.sync_viser_message_defs:entrypoint"
 
-[options]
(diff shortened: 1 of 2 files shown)


ringohoffman:
@brentyi my TOML extension ([Even Better TOML](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml)) really doesn't like this config in `pyproject.toml` and it makes this file hard to read:
Screenshot 2024-01-15 at 21 44 22

The documentation I linked also indicates that this isn't the correct place for this config. Can we consider moving it to `setup.cfg`?

brentyi:
Thanks for the bump, maybe we can just remove these lines without adding them to `setup.cfg`?

From the doc page you linked: `Dependency links support has been dropped by pip starting with version 19.0 (released 2019-01-22).`

brentyi (approved):

PR #2754, opened 2024-01-12
Add isort

ringohoffman:
`isort` is a linter that complements black and helps keeps imports nicely formatted: https://pycqa.github.io/isort/

Mainly, it splits imports into 3 groups: builtins, third party (e.g. `torch`), first party (i.e. `nerfstudio`). It enforces proper spacing between these groups, combines multiple imports from the same module into one line, alphabetizes them, and respects line-length by splitting long import statements into multiple lines.


brentyi:
Can we use ruff's import sorting instead? I think this is the more future-proof option.

(similarly, we could drop `black` in favor of `ruff format`)

PR #2755, opened 2024-01-12
Fixed bug that prevented the Blender dataset from being loaded in gaussian splatting.

pa-la:
Images from the [Blender dataset](https://drive.google.com/drive/folders/1JDdLGDruGNXWnM1eqY1FNL9PlStjaKWi) do not have any distortion. The current code skips images for which distortion parameters are absent, which causes no image to be loaded at all.
Instead, when the distortion parameters are absent, we load the image and we skip the un-distortion part.


kerrj (approved):
lgtm thanks!

KevinXu02:
#2704 could you please check that this PR also works? I think I fixed the same thing but with fewer changes, thanks!

kerrj:
Closing this for now since I think #2704 fixes this, it also adds background color support in splatfacto. Feel free to re-open if there's still issues

PR #2760, opened 2024-01-13
Clean up `ns-train {method} --help` for not-yet-installed external methods

brentyi:
`ExternalMethodTrainerConfig` previously inherited from `TrainerConfig`, which resulted in a bunch of arguments.


kerrj (approved):
nice

PR #2761, opened 2024-01-13
Improve viewer banner when server is bound to 0.0.0.0

### LINKED ISSUE #2740: Impossible to connect to webviewer
**Describe the bug**
When training a model or loading a checkpoint, the given viewer address do not lead to any page.

**To Reproduce**
1. Train a model
2. Ctrl + Click on the address of the viewer
3. A blank page with the error message : "This webpage in unreachable" opens

**Expected behavior**
The webviewer should open.

**Additional context**
The given address seems to be wrong. It is by default : "http://0.0.0.0:7007".
This only happens when installing from source. It works fine with pip install.

brentyi:
We now print `localhost:port` instead of `0.0.0.0:port` when the viewer is bound to `0.0.0.0`. The latter was confusing people.

I also removed the ANSI code that makes the viewer URL blink, but can re-add it if anybody prefers to have that feature retained!

closes #2740

### DIFF
--- a/nerfstudio/engine/trainer.py
+++ b/nerfstudio/engine/trainer.py
@@ -21,11 +21,12 @@ import dataclasses
 import functools
 import os
 import time
+from collections import defaultdict
 from dataclasses import dataclass, field
 from pathlib import Path
 from threading import Lock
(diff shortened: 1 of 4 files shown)


tancik (approved):
Lgtm once the pyright error is fixed

PR #2763, opened 2024-01-14
Reduce startup overhead

brentyi:
Some changes to reduce startup overhead!

For just printing the helptext, from `$ time ns-train --help`,
- Before: `10.00s user 5.89s system 166% cpu 9.519 total`
- After: `2.75s user 1.62s system 156% cpu 2.797 total`

The biggest chunk of this was from making some less-used imports lazy. Some examples of things that mattered:
- `torchvision` took 0.5~0.6 seconds to import. (most of which was extensions for the ConvNeXt model...)
- `comet_ml` required 0.3~0.4 seconds.
- `diffusers` required 0.4~0.5 seconds.
- `open3d` required 0.4~0.5 seconds.

(we should ideally do some of this profiling in the CI)

I also bumped to the faster `viser` version (from https://github.com/nerfstudio-project/viser/pull/159) and cleaned up some CLI things, where we were creating unnecessary subparsers for deprecated config fields.


kerrj (approved):
nice!

PR #2772, opened 2024-01-16
sample pixels from all data in variable-resolution batches

decrispell:
This PR bring the behavior of [collate_image_dataset_batch_list](https://github.com/nerfstudio-project/nerfstudio/blob/32939e0d4198e23e9cef2fb64718cc1d263cab61/nerfstudio/data/pixel_samplers.py#L218) in line with that of [collate_image_dataset_batch](https://github.com/nerfstudio-project/nerfstudio/blob/32939e0d4198e23e9cef2fb64718cc1d263cab61/nerfstudio/data/pixel_samplers.py#L163C9-L163C36) by sampling pixels from all data in the batch, not just "image".


decrispell:
@brentyi  I believe this (old) MR is still needed to fix the `collate_image_dataset_batch_list()` function for batches that contain any per-pixel data other than `"image"` and `"depth_image"`.  If the batch contains any other image-like data, it will crash [here](https://github.com/nerfstudio-project/nerfstudio/blob/2091a0d9273f87a01d48af730999b17806926094/nerfstudio/data/pixel_samplers.py#L374) since `value` will be a list, not a tensor.

Any chance this could get merged so we don't have to continue keeping it up to date with main in parallel?

brentyi (approved):
Thanks @decrispell for the bump! I did a pass and this looks reasonable to me.

PR #2777, opened 2024-01-17
camtoworld is already a tensor in sdfstudio_data

maturk:
The dataparser (sdfstudio dataparser) loads camera_to_worlds in a torch tensor just like most other dataparsers do. No need to try to convert from np.array.


kerrj (approved):
sgtm

PR #2778, opened 2024-01-17
Migrate to python functions for gsplat

kerrj:
Update the interface of gsplat to reflect the more pythonic function-based API introduced in gsplat==0.1.2


nerfstudio/models/gaussian_splatting.py
```diff
@@ -691,10 +691,10 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
         camera.rescale_output_resolution(camera_downscale)
 
         # avoid empty rasterization
-        num_intersects, _ = compute_cumulative_intersects(self.xys.size(0), num_tiles_hit)
+        num_intersects, _ = compute_cumulative_intersects(num_tiles_hit)
```
liruilong940607, line 695: I think this assert can be replaced with `(num_tiles_hit > 0).any()` which avoid cumsum so should be faster. Ideally we should gracefully handle this case in gsplat rasterization in the next version.

liruilong940607 (approved):
LGTM with minor suggestion.

nerfstudio/models/gaussian_splatting.py
kerrj: ok updated!

PR #2780, opened 2024-01-17
Update documentation for 3DGS

cvachha:
Adding 3rd party viewers for splats and other edits to gs documentation.


kerrj (approved):
thanks!

PR #2783, opened 2024-01-18
Don't throw away non-tensor objects in TensorDataclass

brentyi:
I noticed after these lines:
https://github.com/nerfstudio-project/nerfstudio/blob/368c9ecd1272c17de32bea6d8c33284936396287/nerfstudio/data/dataparsers/nerfstudio_dataparser.py#L291-L303
that `cameras.metadata` was `{}` even if the `metadata` variable was set to `{'fisheye_crop_radius': 707.5}`.

cc @sweeneychris maybe relevant to you?


liuyuex97:
Hi @brentyi would it work if I have custom fisheye data? I tried setting the fisheye_crop_radius manually but the radius still looks like half of the shorter dimension. Any tips would be great!

brentyi:
@liuyuex97 I think it should work for custom data!

I saw you closed your issue, did you figure this out?

PR #2785, opened 2024-01-18
Gaussian splatting support for Aria

brentyi:
Added support for the Fisheye624 model in the full image dataloader. @kerrj can you check if my changes make sense?

`ns-process-data aria` now also saves the Aria MPS points, which seems to improve results slightly.

Without points:
aria_without_points

With points:
aria_with_points

(this is the same capture as in https://github.com/nerfstudio-project/nerfstudio/pull/2617)

FYI @sweeneychris

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -34,6 +34,7 @@ import torch
 from torch.nn import Parameter
 from tqdm import tqdm
 
+from nerfstudio.cameras.camera_utils import fisheye624_project
 from nerfstudio.cameras.cameras import Cameras, CameraType
 from nerfstudio.configs.dataparser_configs import AnnotatedDataParserUnion
 from nerfstudio.data.datamanagers.base_datamanager import DataManager, DataManagerConfig, TDataset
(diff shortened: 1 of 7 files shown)


nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -414,3 +317,137 @@ def next_eval_image(self, step: int) -> Tuple[Cameras, Dict]:
...
+                ),
+                params=torch.from_numpy(fisheye624_params[None, :]),
+            )
+            .reshape((undist_w, undist_h, 2))
+            .numpy()
+        )
+        map1 = unproj[..., 1]
+        map2 = unproj[..., 0]
+
+        image = cv2.remap(image, map1, map2, interpolation=cv2.INTER_LINEAR)
```
sweeneychris, line 422: nit: might be better to use grid_interpolate?

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -219,6 +221,21 @@ def main(self) -> None:
             "fisheye_crop_radius": rgb_valid_radius,
         }
 
+        # save global point cloud, which is useful for Gaussian Splatting.
+        points_path = self.mps_data_dir / "global_points.csv.gz"
+        if points_path.exists():
```
sweeneychris, line 226: You should always filter the points from the aria point cloud --> here is a simple way to do it: https://facebookresearch.github.io/projectaria_tools/docs/data_utilities/core_code_snippets/mps#point-cloud

sweeneychris (commented):
overall LGTM and glad it was not too much surgery to get it working! Only real Q is about the choice for undistortion parameters (hardcoded vs programmatic). Note that not all Aria profiles have the same resolution for the RGB camera

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -414,3 +317,137 @@ def next_eval_image(self, step: int) -> Tuple[Cameras, Dict]:
...
+
+        # Desired parameters of the undistorted image.
+        import warnings
+
+        warnings.warn("Fisheye624 support in the full images datamanager currently assumes data from Project Aria.")
+        undist_h = 500
+        undist_w = 500
+        undist_K = torch.eye(3)
+        undist_K[0, 0] = 150.0  # fx
+        undist_K[1, 1] = 150.0  # fy
```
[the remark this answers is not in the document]
SeaOtocinclus: Should not come from the camera calibration itself so it would work for various Aria recording data profile

nerfstudio/scripts/datasets/process_project_aria.py
SeaOtocinclus: We also have LOAD interface for any MPS data, no need to include pandas

i.e
```
points_data = mps.read_global_point_cloud(points_path) if points_path else None
# Filter out low confidence points
points_data = filter_points_from_confidence(points_data)
# Down sample points (optional)
# points_data = filter_points_from_count(points_data, 500_000)
# Retrieve point position
point_positions = [it.position_world for it in points_data]
```

nerfstudio/data/datamanagers/full_images_datamanager.py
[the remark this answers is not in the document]
kerrj: agree, isn't it weird for these to be hardcoded?

nerfstudio/scripts/datasets/process_project_aria.py
brentyi: done!!

filtered points (red) look reasonable:
image

nerfstudio/data/datamanagers/full_images_datamanager.py
brentyi: - `grid_interpolate`: you mean switching to a PyTorch-based remapping? I think doing that would probably be better but we can leave for a future PR; the `cv2.remap()` implementation seems more consistent with the other camera models and we could refactor all of them together
- for using `projectaria_tools`: this seems doable but since we don't have access to the VRS files in the preprocessed dataset I think avoiding the dependency here makes sense!

brentyi:
Thanks for the feedback everyone! I added logic for computing the undistorted camera parameters, and fixed some issues with the masks.

Some more results:
image

The results look cleaner now:
image

With initialization points (filtered from MPS) overlayed:
image

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -416,3 +319,156 @@ def next_eval_image(self, step: int) -> Tuple[Cameras, Dict]:
...
+        )
+
+        # Heuristics to determine parameters of an undistorted image.
+        undist_h = int(fisheye_crop_radius * 2)
+        undist_w = int(fisheye_crop_radius * 2)
+        undistort_focal = undist_h / (2 * torch.tan(fov_radians / 2.0))
+        undist_K = torch.eye(3)
+        undist_K[0, 0] = undistort_focal  # fx
+        undist_K[1, 1] = undistort_focal  # fy
+        undist_K[0, 2] = (undist_w - 1) / 2.0  # cx; for a 1x1 image, center should be at (0, 0).
```
sweeneychris, line 417: I think it would make more sense to set the principle point of the rectified pinhole image with the principle point from the fisheye calibration. In other words, I'm not sure the principle point should move during rectification

brentyi: Do you mean to replace `(undist_w - 1) / 2.0` directly with `camera.cx` (or equivalently, `fisheye624_params[3]`)?

To me that seems less compatible with how we're currently computing the undistorted image dimensions. The principal point would need to change if you decreased `fisheye_crop_radius` such that the the undistorted image was much smaller than the original distorted one, right?

f-dy:
Thank you for fixing the masks in GS! https://github.com/nerfstudio-project/nerfstudio/pull/2785/commits/d1783dfb5baf880395fae11f25a8f73517bc09df LGTM!

kerrj (approved):
lgtm!

PR #2788, opened 2024-01-18
MAX_AUTO_RESOLUTION is ensured in auto downscale

jb-ye:
Simple change, this is to ensure 1600 pixels is the maximal (inclusive) dimension in auto-downscaling. This behavior is consistent with several other repos such as Inria Gaussian Splatting ( https://github.com/graphdeco-inria/gaussian-splatting/blob/main/utils/camera_utils.py#L26 )


kerrj (approved):
sgtm

PR #2791, opened 2024-01-18
Switching the gaussian-splatting dataparser from ColmapDataParser to  NerfStudioDataparser and updating ns-process-data 

CardiacMangoes:
Hey guys!

The goal of this PR is to switch the ColmapDataParserConfig that's currently default in the gaussian-splatting method to NerfstudioDataParserConfig

In order to do so, I update ns-process-data so that it outputs a .ply file in its outputs along with a path to the file named `"ply_file_path"`  in transforms.json. The default name of this file is "sparse_pc.ply"

I also added some backwards compatibility so that this commit does work with older datasets. If the dataset includes colmap outputs but a .ply point cloud, we ask the user if we can create one for them.


nerfstudio/process_data/colmap_utils.py
```diff
@@ -638,3 +643,42 @@ def get_matching_summary(num_initial_frames: int, num_matched_frames: int) -> st
         result += " or large exposure changes."
         return result
     return f"[bold green]COLMAP found poses for {num_matched_frames / num_initial_frames * 100:.2f}% of the images."
+
+
+def create_ply_from_colmap(recon_dir, output_dir):
```
ethanweber, line 648: Annotate this with Path's

nerfstudio/process_data/colmap_utils.py
```diff
@@ -638,3 +643,42 @@ def get_matching_summary(num_initial_frames: int, num_matched_frames: int) -> st
...
+    else:
+        raise ValueError(f"Could not find points3D.txt or points3D.bin in {recon_dir}")
+
+    # Load point Positions
+    points3D = torch.from_numpy(np.array([p.xyz for p in colmap_points.values()], dtype=np.float32))
+    # Load point colours
+    points3D_rgb = torch.from_numpy(np.array([p.rgb for p in colmap_points.values()], dtype=np.uint8))
+
+    # write ply
+    with open(output_dir / "sparse_pc.ply", "w") as f:
```
ethanweber, line 668: You can make `sparse_pc.ply` an input parameter to this function. Assert that the filename is a .ply.

nerfstudio/process_data/colmap_utils.py
```diff
@@ -460,6 +461,10 @@ def colmap_to_json(
     applied_transform[2, :] *= -1
     out["applied_transform"] = applied_transform.tolist()
 
+    # create ply from colmap
+    create_ply_from_colmap(recon_dir, output_dir)
+    out["ply_file_path"] = "sparse_pc.ply"
```
ethanweber, line 466: This will change a bit based on my comment to make "sparse_pc.ply" an input param to `create_ply_from_colmap`.

nerfstudio/process_data/colmap_utils.py
```diff
@@ -460,6 +462,11 @@ def colmap_to_json(
     applied_transform[2, :] *= -1
     out["applied_transform"] = applied_transform.tolist()
 
+    # create ply from colmap
+    assert ply_filename.endswith(".ply"), f"ply_filename: {ply_filename} does not end with '.ply'"
+    create_ply_from_colmap(ply_filename, recon_dir, output_dir)
```
jb-ye, line 467: There is a bug here, when creating ply from colmap, you need to apply `applied_transform` to the point cloud similar like here: https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/dataparsers/colmap_dataparser.py#L354-L366

Otherwise, the camera poses will be in different world coordinate as point clouds. And the quality of training GS will be severely affected.

jb-ye:
I am actually against changing the default parser to nerfstudio because majority of community users are using colmap directly now. If this is decided by the nerfstudio team, I would prefer I participate in validating this PR, since converting colmap to nerfstudio format and ensuring all things are well arranged requires a bit more careful checks. @tancik

nerfstudio/process_data/colmap_utils.py
```diff
@@ -664,6 +664,9 @@ def create_ply_from_colmap(filename: str, recon_dir: Path, output_dir: Path):
 
     # Load point Positions
     points3D = torch.from_numpy(np.array([p.xyz for p in colmap_points.values()], dtype=np.float32))
+    # convert to nerfstudio coordinate system
+    points3D = points3D[..., torch.tensor([1, 0, 2])]
+    points3D[..., 2] *= -1
```
jb-ye, line 669: Again this change is still not the right change, even though the training can go through. See https://github.com/nerfstudio-project/nerfstudio/pull/2793 for more discussions.

brentyi:
> I am actually against changing the default parser to nerfstudio because majority of community users are using colmap directly now. If this is decided by the nerfstudio team, I would prefer I participate in validating this PR, since converting colmap project to nerfstudio format and ensuring all things are well arranged requires a bit more careful checks. @tancik

I think we should prioritize consistency here; since all of the other methods default to `nerfstudio-data`, it's unideal if the splatting model is the only one that defaults to `colmap`.

Maybe there's some middle ground, like having the nerfstudio dataparser support loading raw colmap datasets?

jb-ye:
> > I am actually against changing the default parser to nerfstudio because majority of community users are using colmap directly now. If this is decided by the nerfstudio team, I would prefer I participate in validating this PR, since converting colmap project to nerfstudio format and ensuring all things are well arranged requires a bit more careful checks. @tancik
> 
> I think we should prioritize consistency here; since all of the other methods default to `nerfstudio-data`, it's unideal if the splatting model is the only one that defaults to `colmap`.
> 
> Maybe there's some middle ground, like having the nerfstudio dataparser support loading raw colmap datasets?

I am fine with the consistency rationale. Some comments:
(1) I could help validate the PR to ensure the change is compatible with https://github.com/nerfstudio-project/nerfstudio/pull/2793
(2) Add some docs to describe how to use GS from a colmap project either from a colmap parser or converting colmap to nerfstudio and use nerfstudio parser. 
(3) Ideally both path (colmap->gs and colmap->nerfstudio->gs) should create the same artifacts post training (and this requires a bit more careful checks)

brentyi (approved):
I adjusted the coordinate frame logic + left some comments with my understanding of each of them -- things look good when I tested but if you could double-check @CardiacMangoes @jb-ye that would be appreciated!

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -305,28 +307,98 @@ def _generate_dataparser_outputs(self, split="train"):
...
-                [applied_transform, torch.tensor([[0, 0, 0, 1]], dtype=transform_matrix.dtype)], 0
-            )
+        # The naming is somewhat confusing, but:
+        # - transform_matrix contains the transformation to dataparser output coordinates from saved coordinates.
+        # - dataparser_transform_matrix contains the transformation to dataparser output coordinates from original data coordinates.
+        # - applied_transform contains the transformation to saved coordinates from original data coordinates.
+        if "applied_transform" not in meta:
+            # For converting from colmap, this was the effective value of applied_transform that was being
+            # used before we added the applied_transform field to the output dataformat.
+            meta["applied_transform"] = [[0, 1, 0, 0], [1, 0, 0, 0], [0, 0, -1, 0]]
```
jb-ye, line 317: @brentyi For my understanding, is this to backward support old converted nerfstudio format from colmap?

jb-ye:
lgtm, I will test out later this week. I left one comment for my own understanding.

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
brentyi: yeah, that's right!

I ended up tweaking this logic slightly to prevent it from breaking converted datasets that didn't come from colmap, although the core logic is still the same:
https://github.com/nerfstudio-project/nerfstudio/commit/f55bd04ca6f16a85d7f12a479b94850228e3c7d5

Issue #2784, opened 2024-01-18
Colmap to Nerfstudio


frasiolas:
I extract cameras.bin and images.bin from colmap. I use from colmap_utils colmap_to_json to get the transforms.json file and then i run the ns train command. The problem is that the images are not in the same position and orientation that they are in the colmap GUI. Do you know why this is happening?

kerrj:
This might be because of frame conventions, see https://docs.nerf.studio/quickstart/data_conventions.html
You can also use `ns-process-data` directly to call colmap and have it generate the transforms for you

jb-ye:
You can try to set this variable to True
https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/process_data/colmap_utils.py#L394

frasiolas:
I think i found it if you want to keep the same orientation you call colmap dataparser instead of nerfstudio dataparser in ns-train and put the arguments [--method none --center_method none]

jb-ye:
I think you need to call `--orientation_method=none --center_method=none --assume_colmap_world_coordinate_convention=False` in colmap parser after this change ( https://github.com/nerfstudio-project/nerfstudio/pull/2793 )

Nas-Azzam:
> I extract cameras.bin and images.bin from colmap. I use from colmap_utils colmap_to_json to get the transforms.json file and then i run the ns train command. The problem is that the images are not in the same position and orientation that they are in the colmap GUI. Do you know why this is happening?

May be you can try this file 
[colmap2nerf.zip](https://github.com/nerfstudio-project/nerfstudio/files/14120546/colmap2nerf.zip)

LuckyOne09:
> I think you need to call `--orientation_method=none --center_method=none --assume_colmap_world_coordinate_convention=False` in colmap parser after this change ( #2793 ). This is to keep the exactly same world coordinate as input colmap project up to a scale.

@jb-ye Thank you very much for your guidance. But when I follow like this `ns-train splatfacto --data {data_dir} --viewer.quit-on-train-completion False --max-num-iterations 10000 --pipeline.model.sh-degree 0 colmap --colmap_path sparse/0 --downscale_factor=1 --orientation_method=none --center_method=none --assume_colmap_world_coordinate_convention=False --auto-scale-poses=False`  and after I run `ns-export`, I get a splat.py whose many vertices are NaN. I think it is weird since I have nice results in web viewer, could you give me some hints for debugging?

[...]

Issue #1504, opened 2023-02-25
Why is the world coord system rotated in colmap_utils.py?


jkulhanek:
What is the reason the world coord system is rotated here?
https://github.com/nerfstudio-project/nerfstudio/blob/079f419e544b8d2aa6aad6f1a31decf0e06cb88c/nerfstudio/process_data/colmap_utils.py#L620-L621

Could we keep the coordinates as they are in COLMAP?

pwais:
COLMAP uses OpenCV camera conventions, but nerfstudio (and many other nerf impls, maybe all of them?) use OpenGL coordinate system.

tancik:
More info here - https://docs.nerf.studio/en/latest/quickstart/data_conventions.html#coordinate-conventions

jkulhanek:
The two lines don’t change the camera coordinate system though, there is a line above the two which flips z,y to the from opencv to opengl. I thought that in colmap the orientation of world coordinate system is more or less arbitrary. Even when you run something like model_aligner the orientation is not defined. The orientation is overridden in the dataparser anyway, so it doesn’t really matter. I just thought it would be good to have a way to preserve the poses from colmap (when disabling the transformation code in the dataparser).

to the best of my knowledge, in multinerf (first referenced link) they don’t do it. They just switch from opencv to opengl here https://github.com/google-research/multinerf/blob/5d4c82831a9b94a87efada2eee6a993d530c4226/internal/datasets.py#L109

pwais:
you're right, the cited code in nerfstudio is not equivalent to the cited multinerf code.. at least not the precise cited lines.   The cited conversion appears to do a handedness change but then also flips z:

```
array([[ 0, 1,  0,  0],
       [ 1,  0,  0,  0],
       [ 0,  0,  -1,  0],
       [ 0,  0,  0,  1]])
```

(note I'm going to ignore re-scale / re-center)

So, COLMAP is right-handed, OpenGL camera model is right-handed, but OpenGL NDC is left-handed, no?  `np.diag([1, -1, -1, 1])` should account for the change in camera frames, but the nerfstudio code is doing a handedness too? this is very confusing because the *viewer* world frame indicator is definitely right-handed, at least assuming the RGB axes all indicate _positive_ xyz

I believe the original LLFF code used in original nerf does a handedness change as well:
[...]

PR #2793, opened 2024-01-19
Revisiting #1504 and fix colmap world coordinate transform

jb-ye:
Okay, I spent some time to research how we maintain world coordinate conventions between colmap and nerfstudio, and I found the current implementation is not correct (original introduced in a very earlier commit https://github.com/nerfstudio-project/nerfstudio/pull/338 ). Let me first describe my understanding, then what the existing code does and finally what should be the fix. This fix addresses a historic issue: https://github.com/nerfstudio-project/nerfstudio/issues/1504

## The Problem

Given a c2w transform read from colmap, we need apply two transforms, one multiplied from right, one multiplied from left.
1. opencv to opengl  convention transform A which was for changing camera coordinate
2. colmap to nerfstudio convention transform B which was for changing world coordinate
Ideally, the code should look like
```
c2w_new = B * c2w * A
```

Here 
```
A=array([
       [ 1, 0,  0,  0],
       [ 0, -1,  0,  0],
       [ 0, 0, -1,  0],
       [ 0, 0,  0,  1]])
```

Now, let's look at the existing code
```
# effectively multiply A from right
c2w[0:3, 1:3] *= -1
# effectively multiply B from left
c2w = c2w[np.array([1, 0, 2, 3]), :]
c2w[2, :] *= -1
```
Here as implemented in the current code:
```
# incorrect
B=array([
       [ 0, 1,  0,  0],
       [ 1, 0,  0,  0],
       [ 0, 0, -1,  0],
       [ 0, 0,  0,  1]])
```

However, the second step was not correctly implemented. Because in Colmap/Metashape world coordinate convention, the y direction is often interpreted as DOWN direction. This is because the bundle adjustment optimization often starts from using the first opencv camera frame as world frame. Assuming the optimization hasn't deviate the first frame's pose a lot, the y direction in colmap world would be close to DOWN direction. 

In order to transform the world convention from colmap to nerfstudio, the correct B should be
```
# the fix in this PR
B=array([
       [ 1, 0,  0,  0],
       [ 0, 0,  1,  0],
       [ 0, -1, 0,  0],
       [ 0, 0,  0,  1]])
```
where we map  (-y) in colmap world to z in nerfstudio world. And the code should be corrected effectively as
```
c2w[0:3, 1:3] *= -1
c2w = c2w[np.array([0, 2, 1, 3]), :]
c2w[2, :] *= -1
```

Later in nerfstudio commit history, we introduce new feature `--orientation_method` in colmap parser, and the choice of B becomes irrelevant if the `orientation_method` is not `none`. The viewer will always display correct orientation as long as the orientation_method is not none. So in order to test my fix, I run ns-train with `ns-train gaussian-splatting ... colmap --orientation_method=none` and open up my viewer to verify the initial world coordinate setup. See pictures below on mipnerf360 bicycle dataset: 

Without fix
Screenshot from 2024-01-18 23-15-30

With fix

Screenshot from 2024-01-18 22-07-33

## Summary of this change

* [No Effect] If `orientation_method` is not `none` (by default it is set to `up`), the change of this PR has no material impact on outcomes of train (e.g. checkpoint, ply exporter) or viewer. The scene will be oriented as instructed by the parameter. For regular user, whom never touch this parameter other than the default setting, this PR has no impacts for them. 
* [Effect 1] If `orientation_method` is `none`, this is a case we want to properly transform the world coordinate of colmap project to nerfstudio convention. 
* [Effect 2] We introduce a new option to colmap parser (`assume_colmap_world_coordinate_convention`): If `orientation_method` is none, and `assume_colmap_world_coordinate_convention` is False (by default it is true), we will not apply B transform at all.

## Why we need this change?

Users may want to orient the scene using some prior knowledge instead of using what is provided by the nerfstudio, therefore we need effect 1. User may want to orient the scene using nerfstudio world convention instead of colmap convention, then we need effect 2. 

## Other related issues that can find answer in this PR:
https://github.com/nerfstudio-project/nerfstudio/issues/2784


jb-ye:
@kerrj  @tancik Could you take a look? This is a fix for a historic issue.

brentyi:
Nit: I think you have a typo in your `A` matrix. (the x component should be positive).

But the details in the PR are appreciated and I agree with what's written.

brentyi (approved):

jb-ye:
> Nit: I think you have a typo in your `A` matrix. (the x component should be positive).
> 
> But the details in the PR are appreciated and I agree with what's written.

I fixed the typo in main thread. Thx

Issue #1480, opened 2023-02-22
Use pycolmap for all platforms


pwais:
`pycolmap` currently does not build very easily for Windows, and there's no public compiled wheel.  This ticket documents progress and related issues.  Eventually when windows wheels become available, search for this issue number in code to port fully to `pycolmap`. 

References:
 * Historical issue: https://github.com/nerfstudio-project/nerfstudio/issues/263
 * pycolmap added to docker for hloc support: https://github.com/nerfstudio-project/nerfstudio/pull/1478
 * Attempt to build windows wheels: https://github.com/colmap/pycolmap/pull/114
 * Request for help from COLMAP team: https://github.com/colmap/colmap/issues/1810

pwais:
closing this, notes in https://github.com/colmap/pycolmap/issues/111#issuecomment-1633078069

VasuAgrawal:
FYI, it seems `pycolmap` has Windows wheels now: https://pypi.org/project/pycolmap/#files

PR #2794, opened 2024-01-19
Disable projectaria tools on Windows

VasuAgrawal:
The `projectaria-tools` package doesn't have Windows binaries, which prevents `pip install -e .[dev]` from working. I've updated the specification to require a platform other than Windows for installation. I also updated the specification to not use `[all]` after a discussion with the library authors, and fixed an underscore typo which was probably being silently transformed by `pip`.

Previously (https://github.com/nerfstudio-project/nerfstudio/pull/2672), I had also disabled `pycolmap`, but it turns out that very recently (as of Jan 11), there's now Windows wheels published for `pycolmap` on PyPi (https://pypi.org/project/pycolmap/#files), so it actually installs just fine. I also posted about that in the associated issue (https://github.com/nerfstudio-project/nerfstudio/issues/1480) in case someone wants to pick up the task to enable `pycolmap` usage again.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,12 +91,15 @@ dev = [
     "typeguard==2.13.3",
     "ruff==0.1.13",
     "sshconf==0.2.5",
-    "pycolmap>=0.3.0",  # NOTE: pycolmap==0.3.0 is not available on newer python versions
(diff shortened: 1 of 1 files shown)


brentyi (approved):
makes sense to me, thanks!

PR #2798, opened 2024-01-20
Remove legacy viewer from GitHub actions

brentyi:
This is often blocking merges, and I don't think we need it anymore?


kerrj (approved):
sgtm

PR #2801, opened 2024-01-20
Update nerfstudio_dataparser.py

ichsan2895:
This is initial code:
https://github.com/nerfstudio-project/nerfstudio/blob/8952fb531ea09db00c481de95cafadbb80c262e6/nerfstudio/data/dataparsers/nerfstudio_dataparser.py#L76

I fix to `--load-3D-points` set to True by default.

Why I set True by default? Because I see another recommendation in this line for set it true
https://github.com/nerfstudio-project/nerfstudio/blob/8952fb531ea09db00c481de95cafadbb80c262e6/nerfstudio/data/dataparsers/colmap_dataparser.py#L96

Meanwhile, we already switching from colmap dataparser to nerfstudio dataparser https://github.com/nerfstudio-project/nerfstudio/pull/2791

### DIFF
--- a/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
+++ b/nerfstudio/data/dataparsers/nerfstudio_dataparser.py
@@ -73,7 +73,7 @@ class NerfstudioDataParserConfig(DataParserConfig):
     """The interval between frames to use for eval. Only used when eval_mode is eval-interval."""
     depth_unit_scale_factor: float = 1e-3
     """Scales the depth values to meters. Default value is 0.001 for a millimeter to meter conversion."""
-    load_3D_points: bool = False
+    load_3D_points: bool = True
     """Whether to load the 3D points from the colmap reconstruction."""
 
 


ichsan2895:
If we don't declare `colmap --load-3D-points`, it was False by Default
`
ns-train nerfacto --data ./Nerfstudio/COLMAP_PLY --output-dir ./Nerfstudio/outputs_v1 --pipeline.model.eval-num-rays-per-chunk 8192 --pipeline.datamanager.train-num-rays-per-batch 1024 --pipeline.datamanager.eval-num-rays-per-batch 1024
`

image

If I declare it
`
ns-train nerfacto --data ./Nerfstudio/COLMAP_PLY --output-dir ./Nerfstudio/outputs_v1 --pipeline.model.eval-num-rays-per-chunk 8192 --pipeline.datamanager.train-num-rays-per-batch 1024 --pipeline.datamanager.eval-num-rays-per-batch 1024 colmap --load-3D-points True
`
image

brentyi:
Hm yeah, it seems reasonable to make these two consistent. But this change only impacts non-splatting methods where the points aren't used, right? (like nerfacto)

For the nerfstudio dataparser, this is currently overridden to `True` by default for splatting: https://github.com/nerfstudio-project/nerfstudio/blob/8952fb531ea09db00c481de95cafadbb80c262e6/nerfstudio/configs/method_configs.py#L601-L603

I also verified this in the helptext from `ns-train splatfacto nerfstudio-data --help`.

ichsan2895:
> Hm yeah, it seems reasonable to make these two consistent. But this change only impacts non-splatting methods where the points aren't used, right? (like nerfacto)
> 
> For the nerfstudio dataparser, this is currently overridden to `True` by default for splatting:
> 
> https://github.com/nerfstudio-project/nerfstudio/blob/8952fb531ea09db00c481de95cafadbb80c262e6/nerfstudio/configs/method_configs.py#L601-L603
> 
> I also verified this in the helptext from `ns-train splatfacto nerfstudio-data --help`.

yeah, you are alright. The `--load-3D-points` is true when I use `ns-train splatfacto` by default. Thats why I make the other method same as `splatfacto` for consistency.

brentyi:
Are there any other methods that will use the points?

The one argument I see for keeping the discrepancy is that the colmap parser always has points available, while the nerfstudio parser only sometimes has points.

If we always set `load_3D_points=True` for the nerfstudio parser, there will be many cases where this warning will be hit, even for models like `nerfacto`:
https://github.com/nerfstudio-project/nerfstudio/blob/8952fb531ea09db00c481de95cafadbb80c262e6/nerfstudio/data/dataparsers/nerfstudio_dataparser.py#L392-L395

This might be confusing for people.

kerrj:
should this be closed?

ichsan2895:
Okay, let close it since you said this: 
```
If we always set load_3D_points=True for the nerfstudio parser, there will be many cases where this warning will be hit, even for models like nerfacto
```

PR #2804, opened 2024-01-22
Update splatfacto.py

Harr7y:
Shape error: The shape of batch['mask'] is [H, W, 1], not[H, W]

### DIFF
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -827,8 +827,9 @@ class SplatfactoModel(Model):
         # Set masked part of both ground-truth and rendered image to black.
         # This is a little bit sketchy for the SSIM loss.
         if "mask" in batch:
-            assert batch["mask"].shape == gt_img.shape[:2] == pred_img.shape[:2]
-            mask = batch["mask"][..., None].to(self.device)
+            # batch["mask"] : [H, W, 1]
+            assert batch["mask"].shape[:2] == gt_img.shape[:2] == pred_img.shape[:2]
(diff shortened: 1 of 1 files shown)


brentyi:
Thanks for the fix!

Just to confirm: what kind of data are you getting HxWx1 masks from? Do you know where in the code these are generated?

It looks like this error is my fault; I added the assertion in https://github.com/nerfstudio-project/nerfstudio/pull/2785, where I'm also creating HxW masks.

cc also @f-dy

f-dy:
I haven't tested the latest mask changes yet (we'll be doing it soon)

Harr7y:
> Thanks for the fix!
> 
> Just to confirm: what kind of data are you getting HxWx1 masks from? Do you know where in the code these are generated?
> 
> It looks like this error is my fault; I added the assertion in #2785, where I'm also creating HxW masks.
> 
> cc also @f-dy

I used the bear data mentioned in nerf2nerf. The mask is generated by myself, a picture with a black center and white sides.
I used the **full_images_datamanager**, and I think the code used for generating the mask should at https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/utils/data_utils.py#L25.

Tao-11-chen:
Hi, the same situation for me, it should be merged.

kerrj (approved):
thank you! @brentyi it looks like the rest of the repo expects HxWx1, so we should probably change aria to be consistent with that. see https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/utils/data_utils.py#L25

PR #2806, opened 2024-01-23
Add documentation on existing datasets

VasuAgrawal:
Add a new documentation page describing some existing datasets, to help improve discoverability and make it easier to compare different options.

There are a few minor things I wanted to do but wasn't able to figure out, which I'd appreciate guidance on:
* How to make a reference to the automatically generated `reference/cli/ns-download-data.html`, rather than hardcoding in a link to the online version (line 5).
* How to have a reference to another heading, so that the "Image Count<sup>1</sup>" and "Image Resolution<sup>2</sup>" table headings in the summary table can directly link to the larger tables at the bottom. The standard markdown syntax didn't seem to work here, and I couldn't get the [explicit targets](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#creating-explicit-targets) in MyST to work.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -82,6 +82,7 @@ We hope nerfstudio enables you to build faster 🔨 learn together 📚 and cont
 
 quickstart/installation
 quickstart/first_nerf
+quickstart/existing_dataset
 quickstart/custom_dataset
 quickstart/viewer_quickstart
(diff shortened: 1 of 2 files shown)


kerrj (approved):
Great thank you!

PR #2810, opened 2024-01-23
Make downloading more windows-friendly

VasuAgrawal:
I ran into a few issues while running the various download commands on my Windows box. This diff attempts to remedy those issues. In no particular order:

* Switching from `os.system` to `subprocess.run` calls, which have better handling of `'` vs `"` and other shell escapes. In particular, the `'` which were previously used (line 431) weren't working in CMD, which is used by `os.system` on my machine. They work fine in powershell, though.
* Updated `curl` call to use `-o` flag to directly output to file, rather than going through IO redirection.
* Updated all download functions to use unique temp directories, rather than sharing `.temp`. This allows multiple download jobs to run concurrently, since they're typically rate-limited by the upstream.
* More informative error message for when there's more than 1 inner folder, though I haven't run into this issue since switching the temp folder names.

### DIFF
--- a/nerfstudio/scripts/downloads/download_data.py
+++ b/nerfstudio/scripts/downloads/download_data.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 import json
 import os
 import shutil
+import subprocess
 import tarfile
 import zipfile
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


nerfstudio/scripts/downloads/download_data.py
```diff
@@ -143,10 +144,10 @@ def download_capture_name(save_dir: Path, dataset_name: str, capture_name: str,
     file_id_or_zip_url = capture_name_to_file_id[capture_name]
     if file_id_or_zip_url.endswith(".zip"):
         url = file_id_or_zip_url  # zip url
-        target_path = str(save_dir / f"{dataset_name}/{capture_name}")
+        target_path = str(save_dir / f"{dataset_name}" / f"{capture_name}")
```
cr333, line 147: Consider simplifying this and other similar changes:
```suggestion
        target_path = str(save_dir / dataset_name / capture_name)
```

VasuAgrawal: Oops, midnight brain wasn't thinking too hard. I'll fix this.

VasuAgrawal:
Fixed. Also rebased the changes to be on top of master again.

PR #2816, opened 2024-01-24
Change ns-export command to incorporate splatfacto naming

lanxinger:
Change ns-export command to incorporate "splatfacto" naming convention.


kerrj:
Thanks for the PR! We're hesitant to call the export option `splatfacto` since the export is compatible with standard `.ply` exports. splatfacto is a method, but the export type is still the same data format as others. That said, renaming it from `gaussian-splatting` to something shorter like `splat` would be nice to make tab completion more intuitive.

brentyi:
Something shorter or more specific sounds good to me, but `ns-export splat` seems confusing because of the common `.splat` file format (which is different from the `.ply` we're exporting).

PR #2787, opened 2024-01-18
Update colab for new viewer

ginazhouhuiwu:
Changes:
- deprecate old viewer elements in notebook
- use conda for colmap installation

TODO: Viser share link doesn't seem to be printing properly. `ns-train` cell sometimes stops executing after saving the configs.

Issue #2406, opened 2023-09-07
Colmap binary link in Colab is invalid


zeozeozeo:
**Describe the bug**

**To Reproduce**
1. Go to [the colab page](https://colab.research.google.com/github/nerfstudio-project/nerfstudio/blob/main/colab/demo.ipynb)
2. Run `Install Nerfstudio and Dependencies (~8 min)`
3. See error when downloading colmap from [this link](https://drive.google.com/u/0/uc?id=15WngFRNar_b8CaPR5R-hvQ3eAnlyk_SL&confirm=t)

**Expected behavior**
Colmap installing successfully

![image](https://github.com/nerfstudio-project/nerfstudio/assets/108888572/e88a448d-8ac0-40de-b7de-d54500815086)

cmaranes:
I have the same error

ShreelekhaR:
I have the same error, won't run.

ginazhouhuiwu:
Can you try replacing that link with this: https://drive.google.com/file/d/1HfqTvVEmryU-W9zUFvRYoPsRQqnOD7ta&confirm=t

```
# Installing COLMAP
%cd /content/
!gdown "https://drive.google.com/file/d/1HfqTvVEmryU-W9zUFvRYoPsRQqnOD7ta&confirm=t"
```

Let me know if it works!

cmaranes:
> Edit: Actually there is still an issue with installation after this, let me look into it further and update soon.
> 
> Can you try replacing that link with this: https://drive.google.com/file/d/1HfqTvVEmryU-W9zUFvRYoPsRQqnOD7ta&confirm=t
> 
> ```
> # Installing COLMAP
> %cd /content/
> !gdown "https://drive.google.com/file/d/1HfqTvVEmryU-W9zUFvRYoPsRQqnOD7ta&confirm=t"
> ```

Thanks for the answer! However, I'm trying to access the link but it says that the file doesn't exist

ginazhouhuiwu:
Can you try to go into the code block, delete that entire section under `# Installing COLMAP` and replace it with `!sudo apt-get install colmap`. You can check that colmap is installed correctly by running `!colmap -h` afterwards. This is slightly suboptimal due to some dependency issues still (need a bit more time to fix) but will at least allow you to run the colab demo for now!

ShreelekhaR:
I will try!

zeozeozeo:
[...]

PR #2823, opened 2024-01-25
fixing colmap installation to use apt-get since gdown on the URL unable to download colmap

AntonioMacaronio:
in the demo colab notebook, `gdown` is unable to install `colmap` because the gdrive link is not sharable or not available. 

This issue is also mentioned here:
1. https://github.com/nerfstudio-project/nerfstudio/pull/2787/files
- `!conda install -c conda-forge colmap` will raise a `PackagesNotFoundError`.
2. https://github.com/nerfstudio-project/nerfstudio/issues/2406#issuecomment-1746746583
- the fix mentioned here will also have a `PackagesNotFoundError` and then the runtime always crashes/restarts

To download colmap, using `apt-get` solved these issues and allowed me to process my data in the following cell - it is a quick one-line and it takes about 4minutes 16seconds to install nerfstudio and its dependencies

### DIFF
--- a/colab/demo.ipynb
+++ b/colab/demo.ipynb
@@ -3,8 +3,8 @@
     {
       "cell_type": "markdown",
       "metadata": {
-        "id": "view-in-github",
-        "colab_type": "text"
+        "colab_type": "text",
+        "id": "view-in-github"
       },
       "source": [
         "<a href=\"https://colab.research.google.com/github/nerfstudio-project/nerfstudio/blob/alex%2Fcolab-logo-fix/colab/demo.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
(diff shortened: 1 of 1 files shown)


ginazhouhuiwu (commented):
lgtm

brentyi:
I think that the `apt` version of colmap isn't GPU-accelerated, so maybe still revisiting again before #2787 is merged?

ginazhouhuiwu:
Will definitely do!

ginazhouhuiwu:
I think for now if it patches the issue (which it seem to do), can be merged in main. I personally didn't run into problems with `conda` install so I have to look into it more and see what the discrepancy is.

brentyi:
Sounds good! I think missing GPU support in `colmap` is possibly also fine if it makes the install much easier. We could maybe benchmark.

machenmusik:
I would recommend benchmarking, as my recollection of colmap non-GPU on larger datasets was... significantly slower execution.

PR #2824, opened 2024-01-25
Add alpha rendering to splatfacto, fix depth bug

kerrj:
Fixes a bug in depth rendering which ignores transmittance values of non-saturated pixels, adds alpha rendering to splatfacto with the `accumulation` output, which can be used for alpha supervision, compositing, etc.


vye16 (approved):
lgtm

PR #2825, opened 2024-01-25
Fix "mutable default" error in render.py

pyetras:
This fixes 

```
ValueError: mutable default <class 'nerfstudio.data.scene_box.OrientedBox'> for field obb is not allowed: use default_factory
```


brentyi (approved):
thanks @pyetras!

PR #2827, opened 2024-01-26
Added a gradio webui for nerfstudio

KevinXu02:
The webui is still a work in progress. I would appreciate it if someone could integrate the CLI into it. Currently, it only supports ns-train, with plans to add ns-view support later.

The webui can be opened by "python webui.py".

Known Bugs:
The command lines break when reach the end of training for there is a emoji, and the model won't be saved after training because of it.


hypercryptoman:
working perfectly for me <3

Issue #2815, opened 2024-01-24
Background color in splatfacto renders can be random if no crop is set


kerrj:
**Describe the bug**
When rendering splatfacto, if a crop isn't set the background color will be random

**To Reproduce**
`ns-render` any splatfacto model where the background is poorly defined (ie underreconstructed)

**Expected behavior**
same color as viewer will be in background

**Screenshots**
See orange in background of this scene
<img width="500" alt="Screenshot 2024-01-23 at 4 56 26 PM" src="https://github.com/nerfstudio-project/nerfstudio/assets/15806475/f3b33cb8-7364-46b1-9cc9-53bd3a6903c3">

**Additional context**
in render.py, BACKGROUND_COLOR_OVERRIDE is only set if the crop is set, this should be set whether the crop is used or not.

KevinXu02:
Hi! Could you check #2827 with the "flip" bg option on?

brentyi:
@KevinXu02 my understanding of the problem here is that the background is randomized during rendering for both training and exporting/evaluating, whereas we only want the former. This seems a bit different from what the `flip` option you're proposing tries to do, right?

KevinXu02:
Oh I know. Maybe we should just set the the background color to white or the viser color besides training to fix this.

KevinXu02:
> Oh I see. Maybe we should just set the the background color to viser color besides training to fix this. PR is updated to set the initialized background color the same as viser. The should won't affect training.

Tested and it should be working.
![image](https://github.com/nerfstudio-project/nerfstudio/assets/50019703/fda4c50d-858e-4d7a-b242-a1cdca890421)

kerrj:
#2831 patched this

PR #2831, opened 2024-01-26
Fix the background color bug when exporting video with the "random" background option

KevinXu02:
Set default bg to viser color in gs to fix strange bg color in rendering.

### DIFF
--- a/colab/demo.ipynb
+++ b/colab/demo.ipynb
@@ -3,8 +3,8 @@
     {
       "cell_type": "markdown",
       "metadata": {
-        "id": "view-in-github",
-        "colab_type": "text"
+        "colab_type": "text",
+        "id": "view-in-github"
       },
       "source": [
         "<a href=\"https://colab.research.google.com/github/nerfstudio-project/nerfstudio/blob/alex%2Fcolab-logo-fix/colab/demo.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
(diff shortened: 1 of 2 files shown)


bobye:
Training with random background is a feature not a bug.

KevinXu02:
> Training with random background is a feature not a bug.

Hi! This modification won’t affect training, the background is still randomly picked . It only change the background color when exporting video.

jb-ye:
I see. But the code here is quite confusing. Could you add some comments explaining the effect of setting this to a magic value?

KevinXu02:
> I see. But the code here is quite confusing. Could you add some comments explaining the effect of setting this to a magic value?

Oh, the value is just the same background color as the one for viser. The training background logic is defined seperately, so this value would only work when it is not in training.

jb-ye:
Yeah, I know. That's what I suggested for adding those comments above the line of change.

KevinXu02:
> Yeah, I know. That's what I suggested for adding those comments above the line of change.

Comments added. Thank you for your suggestion.

kerrj:
This works as a patch, but what might be a more permanent fix is to propagate the background color set in viser into render `crop_data` so that [this line ](https://github.com/nerfstudio-project/nerfstudio/blob/1bc5b424b41d889ed845f16e55bd7631b3175622/nerfstudio/scripts/render.py#L186)gets called even if the scene isn't cropped

kerrj (approved):
fixes #2815 , but might want to revisit how background is handled in the render `crop_data` so that users can define a background even if the 'crop scene' option is not selected

Issue #2272, opened 2023-07-26
Colab Notebook not generating transforms.json file for custom data


niconielsen32:
Have tried running the colab notebook for some days now without success. The notebook have been working for some weeks but suddenly stopped. When I try to upload and process the data it only creates the image folders but not the transforms.json file. It stops running after the image extraction phase.

Have been trying some different wrap arounds and using instant-ngp to generate the file and then manually upload it and reformat the transform file. Doesn't seem to work either. A bit weird since most versions for the dependencies are locked.

Any possible solutions?

Tianchong-Jiang:
I had the same issue. Have you tried to comment out all the lines with "CONSOLE" in the data processing part in nerfstudio? It seems to me that whenever there is an error and a call of CONSOLE, the program exits without error trace backs. In this case, there is a call of CONSOLE to print "Done copying images".

yankeesong:
I'm having the same issue. I dig a little bit and found that COLMAP installation has errors. I don't want to dig further since I'm not an expert. Could someone fix this please?

yixiang1120:
Same issue here. Could someone fix this please? Thanks!

sudoku-lord:
Yep, I'm getting the same issue in a notebook which previously had no issues with generating transforms.json. Was there some kind of update in the code which affected COLMAP versioning? @tancik (sorry for the tag, just hoping for a quick resolution)

tancik:
It sounds like out provided compiled colmap is no longer installing correctly. Likely some underlying colab dependencies changed. In the meantime you can try installing colmap using these instructions - https://gist.github.com/kwea123/f0e8f38ff2aa94495dbfe7ae9219f75c

PR #2833, opened 2024-01-27
Adding complete COLMAP installation and build with GPU support to colab notebook to address multiple issues with custom user data 

AntonioMacaronio:
### Problem and Background Issues

1. https://github.com/nerfstudio-project/nerfstudio/pull/2823/files 
Previous colmap installation issues were mentioned above and were addressed with a change to use `apt-get`. 
However, this version of colmap may not have GPU support (cuda), as mentioned by brentyi [here](https://github.com/nerfstudio-project/nerfstudio/pull/2823#issuecomment-1911140062).

2. https://github.com/nerfstudio-project/nerfstudio/issues/2272
In previous issues, Matthew Tancik also mentioned [here](https://github.com/nerfstudio-project/nerfstudio/issues/2272) that this could be related to the missing `transform.json` file when a user uploads their own custom data. This is now fixed! 😄 

3. Even if following the official COLMAP installation guide located [here](https://colmap.github.io/install.html), every user running this colab notebook will have a build failure because colab VM's run the Ubuntu 18.04 Linux Distribution, which will result in CMake linking errors. This is mentioned by various colab users [here](https://github.com/stevenlovegrove/Pangolin/issues/718#issuecomment-995887386) and [here](https://github.com/stevenlovegrove/Pangolin/issues/738).

### Changes

1. Added the installation of all dependencies (including `libcgal-qt5-dev`) as denoted on the official [COLMAP installation guide](https://colmap.github.io/install.html), which now has GPU support to address the concerns [here](https://github.com/nerfstudio-project/nerfstudio/pull/2823#issuecomment-1911364457)

2. Path of OpenGL is changed to `/usr/local/lib/libOpenGL.so` (before it's in `usr/lib/x86_64-linux-gnu/libOpenGL.so`). Added the installation of `libglynd` as the steps described [here](https://github.com/colmap/colmap/issues/1271#issuecomment-931900582). During the build process of `colmap`, this allows the linker to successfully resolve the "undefined reference to '_glapi_tls_Current'" error.

### Extra Notes
- With this, the "Install Nerfstudio and Dependencies" call now takes 35 minute to run (instead of 8 oof)
- The following cell "Downloading and Processing Data" can successfully generate the `transforms.json` file needed for training the NeRF as `colmap` is now successfully built
- A user can upload their own custom data now


ginazhouhuiwu:
You're building colmap from scratch right? Does everything together only take 1 minute?

AntonioMacaronio:
> You're building colmap from scratch right? Does everything together only take 1 minute?

I'm sorry I made a mistake (this is def too good to be true 💀), every time i tested these cells for time I was only just restarting my runtime, but colab may still cache some cell results. When I logged into another gmail account and tried this and it takes about 25 minutes to start and finish, so it is quite slower

PR #1486, opened 2023-02-23
Volinga model and NVOL documentation

Frivas97:
This Pull Request implements Volinga model, which is a simple adaptation of nerfacto for real-time. It also add the documentation for the conversion of ckpts into nvols that can be used in Unreal Engine.

Finally, this PR also changes some attributes of nerfacto from parameters to buffers. Since they are not being optimized, it is a better practice to use buffers (https://discuss.pytorch.org/t/what-is-the-difference-between-register-buffer-and-register-parameter-of-nn-module/32723/2).

PR #2834, opened 2024-01-27
Add average_init_density to improve robustness of nerfacto training

jb-ye:
This has been my personal recipe for training nerfacto: the basic issue of current nerfacto implementation is the scene is always initialized with an average density at 1.0 (due to the exponential activation of MLP). This is not a robust choice in general especially when scene depth is much higher than 1.0. 

One famous example is the mipnerf360 `stump` dataset, when training with the recommended setting before this PR
```
ns-train nerfacto --data stump \
  --pipeline.model.appearance_embed_dim 1 \
  --pipeline.model.camera_optimizer.mode off
```
The PSNR on eval set is only 18. With this PR, one can choose to train with a multiplier on top of density output. My personal experience is to set it 0.01 so the scene is very transparent initially. 
```
ns-train nerfacto --data stump \
  --pipeline.model.appearance_embed_dim 1 \
  --pipeline.model.camera_optimizer.mode off \
  --pipeline.model.average_init_density 0.01
```
The PSNR on eval set will be improved from 18 to 25.3. Similar improvement can also be observed in some of nerfstudio dataset.

At last, this feature would not produce more quality in the final render, but will avoid some bad optimization outcomes due to improper initialization.

### DIFF
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -97,6 +97,7 @@ method_configs["nerfacto"] = TrainerConfig(
         ),
         model=NerfactoModelConfig(
             eval_num_rays_per_chunk=1 << 15,
+            average_init_density=0.01,
             camera_optimizer=CameraOptimizerConfig(mode="SO3xR3"),
         ),
     ),
@@ -140,6 +141,7 @@ method_configs["nerfacto-big"] = TrainerConfig(
             max_res=4096,
             proposal_weights_anneal_max_num_iters=5000,
             log2_hashmap_size=21,
+            average_init_density=0.01,
             camera_optimizer=CameraOptimizerConfig(mode="SO3xR3"),
         ),
     ),
@@ -187,6 +189,7 @@ method_configs["nerfacto-huge"] = TrainerConfig(
             max_res=8192,
             proposal_weights_anneal_max_num_iters=5000,
             log2_hashmap_size=21,
+            average_init_density=0.01,
             camera_optimizer=CameraOptimizerConfig(mode="SO3xR3"),
         ),
     ),
--- a/nerfstudio/fields/density_fields.py
+++ b/nerfstudio/fields/density_fields.py
@@ -56,12 +56,14 @@ class HashMLPDensityField(Field):
         base_res: int = 16,
         log2_hashmap_size: int = 18,
         features_per_level: int = 2,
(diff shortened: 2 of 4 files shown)


nerfstudio/fields/density_fields.py
```diff
@@ -66,6 +67,7 @@ def __init__(
         self.register_buffer("max_res", torch.tensor(max_res))
         self.register_buffer("num_levels", torch.tensor(num_levels))
         self.register_buffer("log2_hashmap_size", torch.tensor(log2_hashmap_size))
+        self.register_buffer("average_init_density", torch.tensor(average_init_density))
```
jkulhanek, line 70: Why is this a buffer here?

jb-ye: This is my question as well, I saw other configurable parameters are set as buffer. I just tried to mimic the code style.

Anyone who can explain why `max_res`, `num_levels`, `log2_hashmap_size` are set as buffer?

jkulhanek: Seems a bit wasteful to me.

jkulhanek: Also it makes the new checkpoint not compatible with the previous one.

tancik: It was originally changed here - https://github.com/nerfstudio-project/nerfstudio/pull/1486

jkulhanek: In the case of aabb, it was a parameter before. It makes sense as it is a tensor. For scalars I think it’s better to keep them as floats?

brentyi: Probably Volinga wanted it so they could read config values from the checkpoint's state_dict? But it also seems weird to me.

nerfstudio/configs/method_configs.py
```diff
@@ -97,6 +97,7 @@
         ),
         model=NerfactoModelConfig(
             eval_num_rays_per_chunk=1 << 15,
+            average_init_density=0.01,
```
kerrj, line 100: maybe add these to nerfacto-big and nerfacto-huge too?

kerrj: oops I'm blind you already did

kerrj (approved):
sgtm!

PR #2843, opened 2024-01-28
Add flag to disable apperance embedding

jkulhanek:


nerfstudio/fields/nerfacto_field.py
```diff
@@ -239,17 +240,19 @@ def get_outputs(
...
         # appearance
-        if self.training:
-            embedded_appearance = self.embedding_appearance(camera_indices)
-        else:
-            if self.use_average_appearance_embedding:
-                embedded_appearance = torch.ones(
-                    (*directions.shape[:-1], self.appearance_embedding_dim), device=directions.device
-                ) * self.embedding_appearance.mean(dim=0)
+        embedded_appearance = None
+        if self.appearance_embedding_dim > 0:
```
brentyi, line 244: (2) and then condition this on `if self.embedding_appearance is not None`

because it seems statically safer for `self.embedding_appearance` to be typed as `Embedding | None`

brentyi (approved):
suggested one change, but also looks good to me in the current state!

jkulhanek:
Sorry, I had auto-merge turned on :( I would have updated it.

PR #2838, opened 2024-01-27
Add support for loading 3D points with ODM data processor

pierotofy:
Hello :wave:

this PR adds support for loading 3D points (useful for running `splatfacto`) from datasets imported from ODM.

Hope this can be useful to others.

PR #2847, opened 2024-01-29
Fix ODM data processor coordinate system

pierotofy:
Hello :wave:

I made a mistake in #2838 and while processing a different dataset I noticed an issue with the coordinate system conversion I wrote.

I tested these changes with multiple datasets and they work now.

image

image

Link to datasets exported with proposed change: https://drive.google.com/drive/folders/1tmeBSlLcUJKxpwqhFC2YqqSGUKlGoMtN?usp=sharing

### DIFF
--- a/nerfstudio/process_data/odm_utils.py
+++ b/nerfstudio/process_data/odm_utils.py
@@ -46,36 +46,50 @@ def rodrigues_vec_to_rotation_mat(rodrigues_vec: np.ndarray) -> np.ndarray:
     return rotation_mat
 
 
-def reconstruction_to_ply(reconstruction_file: Path, output_ply: Path):
+def get_reconstruction(reconstruction_file: Path):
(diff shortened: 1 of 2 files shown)


ichsan2895:
Hello @pierotofy  , I want to ask, why `ns-process-data odm` makes the images in `OUTPUT/images` folder has smaller size than original images? But the resolution is remain same.

For example, `ns-process-data odm --data ODM_RESULT --output-dir NERFSTUDIO_ODM

Then I compare the images in `ODM_RESULT/images` vs `NERFSTUDIO_ODM/images`. It was different but the resolution is same.

Is that case intended?
Is that case reduce the quality of reconstruction?

kerrj (approved):
sgtm

pierotofy:
You didn't process it with ODM (https://github.com/OpenDroneMap/ODM), you used NodeODM or WebODM. You'll need the opensfm folder, which is only available in ODM.

PR #2851, opened 2024-01-31
Update depth computation in nerfacto to not be torch.no_grad()

mattstrong-stanford:
This PR removes the torch.no_grad() from the depth computation in nerfacto because when we are using any kind of depth supervision that directly compares the ground truth depth and this depth, no updates will be made if depth is computed with torch.no_grad().

E.g., we can create a simple L1 loss on depth and ground truth depth but the effect will be 0.

### DIFF
--- a/nerfstudio/models/nerfacto.py
+++ b/nerfstudio/models/nerfacto.py
@@ -310,8 +310,7 @@ class NerfactoModel(Model):
         ray_samples_list.append(ray_samples)
 
         rgb = self.renderer_rgb(rgb=field_outputs[FieldHeadNames.RGB], weights=weights)
-        with torch.no_grad():
-            depth = self.renderer_depth(weights=weights, ray_samples=ray_samples)
+        depth = self.renderer_depth(weights=weights, ray_samples=ray_samples)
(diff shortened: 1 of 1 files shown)


OrangeSodahub:
@mattstrong-stanford Why you closed it?

mattstrong-stanford:
@OrangeSodahub the depth computed here uses a median to output the depth, so when we are taking the gradient, there is no influence. I instead had this PR [here](https://github.com/nerfstudio-project/nerfstudio/pull/2852) to take care of this fix.

mattstrong-stanford:
@OrangeSodahub the median depth would be use for visualization and can also be used to for depth mse in evaluation. You are correct about using  expected depth for depth supervision. You simply can't use median depth in depth supervision because no gradients will pass through a median operation. During visualization, I have noticed median depth to be a bit more fine than expected depth (the expectation smooths out values).

OrangeSodahub:
@mattstrong-stanford Thanks, I mean using median depth for depth supervision (defintely should enable the gradient), what I want to emphasize is the difference of median depth and expected depth, the latter is ok for depth supervision, but if I stick to using the former, is it a right way?

PR #2852, opened 2024-01-31
Have predicted depth as expected_depth so gradients can be computed

mattstrong-stanford:
In direct depth supervision methods, calling `depth_loss` will not work unless we can compute gradients on the predicted depth. The predicted depth right now is a median and there are no gradients that can be computed. I have changed it to expected depth to allow direct depth supervision methods to work.


AdamRashid96 (approved):
LGTM

PR #2855, opened 2024-01-31
Avoid OOM for caching undistorted images of splatfacto

jb-ye:
I found many users in discord complains about OOM error when training with > 1000 images. We probably need to enable this option by default in the `method_configs`.


kerrj (approved):
sgtm

PR #2856, opened 2024-01-31
Replace depth alpha mask logic with torch.where for gradients in splatfacto

kerrj:


ethanweber (approved):
LGTM!

nerfstudio/models/splatfacto.py
```diff
@@ -779,8 +779,7 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
                 W,
                 background=torch.zeros(3, device=self.device),
             )[..., 0:1]  # type: ignore
-            depth_im[alpha > 0] = depth_im[alpha > 0] / alpha[alpha > 0]
-            depth_im[alpha == 0] = 1000
+            depth_im = torch.where(alpha > 0, depth_im / alpha, depth_im.detach().max())
```
jb-ye, line 782: Just an unrelated comment, compute the mean(depth) has very high variance in estimation. I would prefer to compute mean(1/depth) as this is more related to disparity maps across views.

PR #2860, opened 2024-02-01
Make colmapDataParser compatible with 360_v2 dataset format

Jing1Ling:
In order to be compatible with the original format of the 360_v2 dataset without reprocessing it, I added an option to colmapDataParser to round up the image size when downscaling. This may change the image height and width in the camera.
But I'm not sure if it will have any negative impact on other parts?  If so, please let me know. Thank you!

### DIFF
--- a/nerfstudio/data/dataparsers/colmap_dataparser.py
+++ b/nerfstudio/data/dataparsers/colmap_dataparser.py
@@ -457,13 +457,20 @@ class ColmapDataParser(DataParser):
             out["points3D_points2D_xy"] = torch.stack(points3D_image_xy, dim=0)
         return out
 
-    def _downscale_images(self, paths, get_fname, downscale_factor: int, downscale_rounding_mode: str = "floor",nearest_neighbor: bool = False):
-        def calculate_scaled_size(original_width, original_height, downscale_factor, mode='floor'):
-            if mode == 'floor':
+    def _downscale_images(
+        self,
+        paths,
+        get_fname,
+        downscale_factor: int,
+        downscale_rounding_mode: str = "floor",
+        nearest_neighbor: bool = False,
+    ):
+        def calculate_scaled_size(original_width, original_height, downscale_factor, mode="floor"):
+            if mode == "floor":
                 return math.floor(original_width / downscale_factor), math.floor(original_height / downscale_factor)
-            elif mode == 'round':
+            elif mode == "round":
                 return round(original_width / downscale_factor), round(original_height / downscale_factor)
-            elif mode == 'ceil':
+            elif mode == "ceil":
                 return math.ceil(original_width / downscale_factor), math.ceil(original_height / downscale_factor)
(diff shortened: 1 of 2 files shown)


nerfstudio/data/dataparsers/colmap_dataparser.py
```diff
@@ -56,6 +56,8 @@ class ColmapDataParserConfig(DataParserConfig):
     """How much to scale the camera origins by."""
     downscale_factor: Optional[int] = None
     """How much to downscale images. If not set, images are chosen such that the max dimension is <1600px."""
+    downscale_rounding_mode: Literal["floor", "ceil"] = "floor"
```
jb-ye, line 59: The ceiling mode isn't compatible with the ffmpeg commands. 

See here: https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/dataparsers/colmap_dataparser.py#L464-L468

You probably need to change this part and test the auto-downscale implementation.

sib1:
Thank for this implementation @Jing1Ling.

I tried using it and discovered that the 360_v2 dataset does not round up when downscaling, but actually rounds to the nearest integer with the half up rule i.e. 7.5 -> 8, 7.7 -> 8, 7.3 -> 7 

You can see an example of this behaviour in the Garden scene. The original resolution of 5187 x 3361 gets downscaled by 2x to 2593.5 x 1680.5 -> 2594 x 1681. And it gets downscaled by 4x to 1296.75 x 840.25 -> 1297 x 840.

A solution which works for me is:

`self.height = torch.floor(0.5 + (self.height * scaling_factor)).to(torch.int64)`
`self.width = torch.floor(0.5 + (self.width * scaling_factor)).to(torch.int64)`

Jing1Ling:
Thanks for your reminder @sib1. For my implementation, maybe I can add a round mode to deal with the situation you mentioned. I will try to update it in the future.

nerfstudio/data/dataparsers/colmap_dataparser.py
Jing1Ling: Hi @jb-ye, I've changed this part. I ran 'ns-train nerfacto --data xxx colmap --downscale_rounding_mode round --downscale_factor 4' and it worked when there only original image folder and colmap folder under the data path. Do you have any other suggestions:)?

nerfstudio/data/dataparsers/colmap_dataparser.py
```diff
@@ -488,7 +514,7 @@ def get_fname(parent: Path, filepath: Path) -> Path:
         if self._downscale_factor is None:
             if self.config.downscale_factor is None:
                 test_img = Image.open(filepath)
-                h, w = test_img.size
+                w, h = test_img.size
```
jb-ye, line 517: Thanks for catching the error

jb-ye (approved):

PR #2861, opened 2024-02-01
Changed torch version to <2.2

simonbethke:
Changed torch version to <2.2, as <=2.1 will uninstall torch 2.1.2+cu118 and leave users without cuda in torch


brentyi (approved):
thanks!!!

PR #2864, opened 2024-02-02
Upper-bound torch version for dev installs only

brentyi:
Since we only upper-bound to suppress some type errors!


tancik (approved):
LGTM

PR #2867, opened 2024-02-02
Add multi-processing support for cache_image. Fixed the issue of insufficient precision leading to a completely black image when converting from RGBA to RGB.

KevinXu02:
Add multi-processing support for cache_images. Now caching image is way more faster.

(Kind of messed up with branches and comments, the only changed file is full_images_datamanager.py. Please bear with me.)


nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -172,14 +172,31 @@ def cache_images(self, cache_images_option):
...
-            self.eval_dataset.cameras.width[i] = image.shape[1]
-            self.eval_dataset.cameras.height[i] = image.shape[0]
+        result_list = []
+        CONSOLE.log("Caching / undistorting train images")
+        with ThreadPoolExecutor(max_workers=multiprocessing.cpu_count()) as executor:
+            for idx in range(len(self.train_dataset)):
+                result = executor.submit(process_train_data, idx)
+                result_list.append(result)
+            for data in track(result_list, description="Caching / undistorting train images", transient=True):
+                cached_train.append(data.result())
```
brentyi, line 190: Untested, but I think you can replace all of these loops with something like:

```python
cached_train = list(
    executor.map(process_train_data, track(range(len(self.train_dataset))))
)
```

brentyi (commented):
Looks reasonable to me! Just left two comments 🙂

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -172,14 +172,31 @@ def cache_images(self, cache_images_option):
...
 
-            self.eval_dataset.cameras.fx[i] = float(K[0, 0])
-            self.eval_dataset.cameras.fy[i] = float(K[1, 1])
-            self.eval_dataset.cameras.cx[i] = float(K[0, 2])
-            self.eval_dataset.cameras.cy[i] = float(K[1, 2])
-            self.eval_dataset.cameras.width[i] = image.shape[1]
-            self.eval_dataset.cameras.height[i] = image.shape[0]
+        result_list = []
+        CONSOLE.log("Caching / undistorting train images")
+        with ThreadPoolExecutor(max_workers=multiprocessing.cpu_count()) as executor:
```
brentyi, line 185: If you're using a thread pool I don't think `multiprocessing.cpu_count()` makes sense here; because of the [GIL](https://wiki.python.org/moin/GlobalInterpreterLock) there's not really any relationship between the optimal number of threads and the number of CPUs.

Since OpenCV will use multiple cores under-the-hood anyways, I think we could just set the number of threads to a high-enough constant values that allows us to parallelize I/O operations (reading images via `.get_data()`) and undistortion.

nerfstudio/data/datamanagers/full_images_datamanager.py
KevinXu02: Yes, this works with a faster speed. Thank you!

nerfstudio/data/datamanagers/full_images_datamanager.py
KevinXu02: The ThreadPoolExecutor's default num_workers is max_workers = min(32, (os.cpu_count() or 1) + 4). I think this could be reasonable and I just leave the init num empty.

KevinXu02:
> Looks reasonable to me! Just left two comments 🙂

Hi! I have updated the code based on the comments. Thank you for your suggestions！

KevinXu02:
Also fixed the issue of insufficient precision leading to a completely black image when converting from RGBA to RGB.

nerfstudio/data/datasets/base_dataset.py
```diff
@@ -103,7 +103,7 @@ def get_image_uint8(self, image_idx: int) -> UInt8[Tensor, "image_height image_w
             assert (self._dataparser_outputs.alpha_color >= 0).all() and (
                 self._dataparser_outputs.alpha_color <= 1
             ).all(), "alpha color given is out of range between [0, 1]."
-            image = image[:, :, :3] * image[:, :, -1:] / 255.0 + 255.0 * self._dataparser_outputs.alpha_color * (
+            image = image[:, :, :3] * (image[:, :, -1:] / 255.0) + 255.0 * self._dataparser_outputs.alpha_color * (
```
jb-ye, line 106: Thanks for catching this bug.

jb-ye:
lgtm

kerrj (approved):
sgtm!

Issue #2283, opened 2023-08-01
"Start training" stops early, renderer disconnects


chigozienri:
**Describe the bug**
"Start Training" cell runs for only a very short period of time, during which time the viewer shows only a small subset of the training images. After the "Start Training" cell has has finished running, the viewer shows "Renderer disconnected". This may be related to #2144 and #1832, although unlike the behaviour reported in those issues, there is some time where the viewer shows as connected.

**To Reproduce**
Environment: Google Colab free tier, t4 GPU, Python 3.10.10
1. Run cells up to "Downloading Data"
2. Upload custom images (in this case I had 39 frames)
3. Run cells up and including to "Start training"
4. Click "refresh page" on viewer, and see images start to appear
5. 15 seconds later, script exits, "renderer exits" on viewer
6. No errors appear, but no way to interact with images, no checkpoints generated

**Expected behavior**
Viewer to connect in a persistent way, nerfacto to start training and generating checkpoints

**Screenshots**
<img width="1150" alt="nerfstudio" src="https://github.com/nerfstudio-project/nerfstudio/assets/14337872/00a04e97-7293-4e18-8d01-267ccc55ab6a">

tancik:
It sounds like you are being kicked off of the port. Is it possible there is a firewall that blocks large traffic spikes?

sudoku-lord:
@tancik This looks like the same issue I'm getting on Google Colab (I'm assuming the poster is also using the Colab notebook). Have there been any updates on this bug? Thank you!

chigozienri:
I am not seeing this issue any more, and I am not sure why. I will keep the issue open for now and update if it reoccurs.

I was indeed using the Colab notebook.

tancik:
Might be related to this
https://github.com/nerfstudio-project/nerfstudio/issues/2328#issuecomment-1676038293

aliabusaleh:
Is there any update on how to fix that?

Issue #2166, opened 2023-07-03
ns-train help flag not displaying anything?


sudoku-lord:
Based on the docs, I've tried running `ns-train nerfacto --help` in order to acquire explanations for the parameters, but it prints absolutely nothing. Is there another way it should be run, or a new command which is more effective? Or is there a list of parameters in the docs somewhere? Thanks!

tancik:
Are there any errors? Can you post a screenshot? It's not in the docs because there are too many.

sudoku-lord:
Hi @tancik , there are no errors. It simply outputs nothing at all. This is how it looked after I finished running it:

<img width="401" alt="Screenshot 2023-07-03 at 6 11 46 PM" src="https://github.com/nerfstudio-project/nerfstudio/assets/15935019/ff834870-d502-4e9b-96a3-6ab4c726f839">

tancik:
Ahh, this is probably a colab specific problem. Can you try in the colab terminal?

Issue #1836, opened 2023-04-28
Colab Version does not train on custom data


MultiTrickFox:
I have uploaded and successfully colmap'ed the video into images and data (even checked against the given examples).

However the training just starts and immediately ends without any visible error;
just exiting..

<img width="933" alt="Screen Shot 2023-04-28 at 5 46 17 PM" src="https://user-images.githubusercontent.com/39859227/235206212-6c1ae8c2-834d-4259-980c-fc29071c6813.png">

ichsan2895:
Same issue occurs with me, 

**IMO** the problem appears because recently `Colab` upgrading their python version (3.9 to 3.10), so maybe `tiny-cuda-nn` is not compatible with it. Hopefully it will be updated to newest version for its dependencies.

tancik:
Yea, colab annoyingly keeps updating python versions without any warning. Until it is updated on our side, I believe there are ways to force colab to use specific versions of python.

AntonioMacaronio:
After experimenting more, this problem is actually not due to colab's python version - if you run `ns-train` on the colab terminal (requires colab pro), the model is able to train and will give you a sharable link. I was using the provided data and ran into the same issue, but it worked once I ran the command on the terminal instead of inside a jupyter cell

Issue #1832, opened 2023-04-27
demo.ipynb: renderer disconnected on colab


wayner0628:
I follow the instruction on demo.ipynb, but the viewer seems to be broken, both custom data and pre-built dozer fail.
The former blocks executed correctly, and it did output all processed data.
Due to the issue of viewer, I also could't train the model, it stopped at saving config.
Not knowing why these happen, please check the issue, thanks!

FuturaTino:
我和你遇到了一样的问题。这是由于colab默认pyhton版本3.10，tinycudann包版本是3.9`tinycudann-1.7-cp39-cp39-linux_x86_64.whl`，两者没有对应上，从而代码无法跑通。
解决方案如下：
`Colab界面->Tools > Command palette > Use fallback runtime version`
使用上一个python(也就是python3.9)，重启runtime，选择GPU设备，重新运行即可。

详情可见该文件：Quick fix for nerfstudio demo colab, issue #1077
https://github.com/vvdb101/nerfstudio_demo_issue_1077/blob/main/nerfstudio_demo_issue_1077.ipynb
希望能解决你的问题。

loftusa:
Hi, unless i'm misunderstanding something, the renderer is disconnected for me as well.

ichsan2895:
Seems similar issue with https://github.com/nerfstudio-project/nerfstudio/issues/1836 due to `Google Colab` upgrading their Python version from 3.9 to 3.10

FuturaTino:
> Seems similar issue with #1836 due to `Google Colab` upgrading their Python version from 3.9 to 3.10

Yeah. Using the fallback version of python will fix this.

sudoku-lord:
I'm having the same issue in Kaggle, but downgrading my Python version there leads me into a fresh batch of errors (there's no super easy workaround in Kaggle to downgrade from 3.10 to 3.9 afaik). Is there a fix which can be implemented by nerfstudio itself? Or an updated version of tiny-cuda-nn which does work with 3.10?

sudoku-lord:
As an update: I switched to Google Colab and found that the fallback runtime is no longer available, so I'm using this

```
!apt-get install python3.9
!ln -sf /usr/bin/python3.9 /usr/bin/python
```
before running the rest of my code, which has no errors. However, the renderer still shows up as disconnected. Can anyone render anything in the demo? I'd really love to get some insight on this, because I've been trying to get it working for quite a while now.

[...]

PR #2870, opened 2024-02-04
Adding Terminal Window to Jupyter Cells so all Colab users (free and pro) can train with Nerfstudio

AntonioMacaronio:
**Background and Issues:**

- All colab users (free and pro) were not be able to run `ns-train` in a jupyter cell, and this has been made apparent in several issues below. (Pro users can get around this, but it's still problematic and explained below)
- https://github.com/nerfstudio-project/nerfstudio/issues/2283
- https://github.com/nerfstudio-project/nerfstudio/issues/1836
- https://github.com/nerfstudio-project/nerfstudio/issues/1832
- https://github.com/nerfstudio-project/nerfstudio/issues/1113
- https://github.com/nerfstudio-project/nerfstudio/issues/2166

The culprit of this bug is that as of Feb 2024, Google Colab runs python 3.10, which has an issue with nerfstudio's rich console outputs. In particular, a python module called `shutil` will set the terminal size to 0 when dealing with rich console outputs. As a result, nothing gets printed, and Colab will shutdown the cell and terminate early before training can begin.

This is mentioned in other python modules which use the rich console output, and a complete breakdown on the issue is detailed [here](https://github.com/rsalmei/alive-progress/issues/157). 

**Overview of Changes**

- By using the `colab-xterm` module, we are able to create a terminal window in a jupyter cell output.
- After installing it, we clear the cell output to make it very user-friendly to follow the instructions.
- The command is printed and instructions are given to help the user run `ns-train` and they will be able to generate the viser/viewer link to see the training progress
- Removed the "Connecting to Viewer Cell" as users can simply use `ns-train`'s sharable URL link
-  Final change was to add the torchaudio dependency which was mentioned [here](https://github.com/nerfstudio-project/nerfstudio/pull/2787#discussion_r1466591294) so that users can train.

**Impact and Other Notes**
- While Pro Colab users can get around this by running `ns-train` in their terminal window, Colab only allows for 1 terminal window. Therefore, to run the render command, pro users need to Ctrl+C the training process hosting the viser/viewer link, which can be quite irritating - especially if they want to make more than just 1 render path.

- Therefore, even once Colab upgrades to python 3.11 in April 2024, there is still a need to have at least 2 terminal windows, which is possible with this change.

- To run `ns-train`, users (free and pro) simply need to follow the instructions to copy and paste the command into the terminal window. This is pretty clear to follow, and looks like the following:
image

- Below is what the cell terminal window looks like after copy+pasting the command on a free colab account:
image


ginazhouhuiwu:
Thank you Anthony!!! Going to test this soon and see if anything is needed, but it looks like everything is good!

ginazhouhuiwu (approved):
Everything seemed good to me, sorry for the delay and thank you again!!

PR #2873, opened 2024-02-06
Enable always returning depth for Gaussian Splatting

ethanweber:
I'm working on some depth losses, and having this functionality to always return depth, even during training, would be very useful for cleanliness in my codebase.


maturk (approved):

maturk:
I agree, LGTM!

PR #2874, opened 2024-02-06
Add default depth output when no gaussians visible

chungmin99:
There should be a default `"depth"` value returned alongside `"rgb"`, or the renderer may crash if there is no gaussians in the front.

More specifically, L173 and L189 will currently crash in nerfstudio if `"depth"` is undefined:
https://github.com/nerfstudio-project/nerfstudio/blob/52cda21186a9a50b29e7480d3dec20b5b15323d3/nerfstudio/viewer/render_state_machine.py#L170-L191

### DIFF
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -667,7 +667,11 @@ class SplatfactoModel(Model):
         if self.crop_box is not None and not self.training:
             crop_ids = self.crop_box.within(self.means).squeeze()
             if crop_ids.sum() == 0:
-                return {"rgb": background.repeat(int(camera.height.item()), int(camera.width.item()), 1)}
+                return {
+                    "rgb": background.repeat(int(camera.height.item()), int(camera.width.item()), 1),
+                    "depth": torch.ones(int(camera.height.item()), int(camera.width.item()), 1) * 10,
+                    "accumulation": torch.zeros(int(camera.height.item()), int(camera.width.item()), 1),
+                }
         else:
             crop_ids = None
         camera_downscale = self._get_downscale_factor()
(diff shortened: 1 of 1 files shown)


nerfstudio/models/splatfacto.py
```diff
@@ -732,7 +732,10 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
             tile_bounds,
         )  # type: ignore
         if (self.radii).sum() == 0:
-            return {"rgb": background.repeat(int(camera.height.item()), int(camera.width.item()), 1)}
+            return {
+                "rgb": background.repeat(int(camera.height.item()), int(camera.width.item()), 1),
+                "depth": torch.ones(int(camera.height.item()), int(camera.width.item()), 1) * 10,
```
brentyi, line 737: should we include `"accumulation"` also?

chungmin99: True. This line also needs to be edited:

https://github.com/nerfstudio-project/nerfstudio/blob/17873aa19551316bf9b8ee38f12a50d36a1ac2df/nerfstudio/models/splatfacto.py#L670

nerfstudio/models/splatfacto.py
```diff
@@ -734,7 +738,11 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
...
```
brentyi, line 743: also do we need to worry about the device that these are on? maybe we should do `background.new_ones()`, `background.new_zeros()` to make sure it's matched?

brentyi (approved):

PR #2880, opened 2024-02-07
Add more attributes to SplatFacto

maturk:
For custom models, we need to access some of the intermediate outputs. Lets make them attributes so its easier to call `outputs = super(SplatFacto).get_outputs(camera)` and still have access to them.

### DIFF
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -721,7 +721,7 @@ class SplatfactoModel(Model):
 
         colors_crop = torch.cat((features_dc_crop[:, None, :], features_rest_crop), dim=1)
 
-        self.xys, depths, self.radii, conics, num_tiles_hit, cov3d = project_gaussians(  # type: ignore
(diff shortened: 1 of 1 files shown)


kerrj:
This is a good idea, but its quite a messy interface since there's so many mutable effects within the get_outputs. I wonder if we should just return them as a payload in the outputs dict?

maturk:
@kerrj, does it make sense to put intermediates in the output dict, since then the viser viewer will try to visualize them as well? Actually, I am thinking that all outputs from project_gaussians `depths, radii, conics, and num_tiles_hit` should be easily accessible since we need them for any additional calls to rasterize_gaussians (e.g. if the user wants to rasterize some additional feats or normals etc).

jb-ye:
Not entirely related to this PR, but I am thinking if we can leverage this [TensorDataclass](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/utils/tensor_dataclass.py#L27) interface to manage intermediate variables and their slicing.

Issue #2881, opened 2024-02-07
ns-render dataset removes too much from the filename


f-dy:
**Describe the bug**
- In `ns-render dataset`, the suffix is removed from image names [here](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/scripts/render.py#L808) using `.with_suffix("")`
- Then a new suffix is added [here](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/scripts/render.py#L859) using `.with_suffix(".jpg")`

The issue is that if the image name contains another dot, the second with_suffix will remove everything after it::
```
>>> from pathlib import Path
>>> n = Path("a.b.png")
>>> ns = n.with_suffix("")
>>> nj = ns.with_suffix(".jpg")
>>> nj
PosixPath('a.jpg')
```
The fix is to remove the call to `.with_suffix("")` [here](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/scripts/render.py#L808)

PR #2882, opened 2024-02-07
Remove useless call to .with_suffix(""), which causes a bug if filename contains dots

f-dy:
Fixes https://github.com/nerfstudio-project/nerfstudio/issues/2881


kerrj (approved):
lgtm

PR #2884, opened 2024-02-07
support generic TDataset in paralleldatamanager

hturki:
As it stands one cannot use a custom Dataset class in ParallelDatamanager. This handles that usecase in the same manner as VanillaDataManager and FullImagesDataManager


brentyi (approved):
seems reasonable to me!

Issue #925, opened 2022-11-09
missing linear_to_srgb at the end of 


devernay:
**Describe the bug**
Volume rendering (just like any compositing operation) should be done in linear color space, so one can assume that the RGB representation in NeRF should be linear.

However, the loss is expressed on sRGB images, which are gamma-compressed.

The original NeRF code didn't care about that detail, by you can see that RawNeRF now properly handles this (it was necessary for the level of detail that RawNeRF is able to handle):
https://github.com/google-research/multinerf/blob/30005650e9e6a1d8a0f561aa848ea65d855fc787/internal/models.py#L599

The linear_to_srgb function is differentiable, and should be added at the end of RGBRenderer.forward(), around here:
https://github.com/nerfstudio-project/nerfstudio/blob/ca69adaaec02f347f650804371623198d30af562/nerfstudio/model_components/renderers.py#L113

This should improve details, especially near object boundaries with Mip-NeRF: a pixel covered with 50% white and 50% black should not have a value of 127, but... 255*(0.5^(1/2.2))~=186

devernay:
For reference, here are two pytorch implementations of linear_to_srgb:
- [NeRF-Factory](https://github.com/kakaobrain/NeRF-Factory/blob/f61bb8744a5cb4820a4d968fb3bfbed777550f4a/src/model/refnerf/helper.py#L25) (based on the multinerf code)
- [Kornia](https://github.com/kornia/kornia/blob/406f03aa3da709dddeee4a6992a1158d2054fc63/kornia/color/rgb.py#L205)

tancik:
Have you observed a noticeable difference when using `linear_to_srgb`?
Here are some initial tests 
<img width="1461" alt="image" src="https://user-images.githubusercontent.com/3310961/203164226-c1cfce97-cfe1-4a35-85d4-609dead897ad.png">

tancik:
Here are some rendered videos. The quality looks fairly similar to me.

Main:
https://user-images.githubusercontent.com/3310961/203196445-cbac6ec9-c3ea-4fe6-ace4-6e66ac1028b1.mp4

With change:
https://user-images.githubusercontent.com/3310961/203196458-d2abf459-3ba2-4020-9e77-e10e9d8770b5.mp4

devernay:
Yes, I did that same change in the original (vanilla) NeRF code a long time ago and that didn't change anything either. My guess is that it is compensated by the density field.
[...]

PR #2888, opened 2024-02-09
Implement new rasterization mode using opacity compensation factor

jb-ye:
See previous threads here: https://github.com/nerfstudio-project/gsplat/pull/117

The PR must be merged after this PR (version bump of gsplat lib, https://github.com/nerfstudio-project/gsplat/pull/124)

### DIFF
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -157,6 +157,16 @@ class SplatfactoModelConfig(ModelConfig):
     """
     output_depth_during_training: bool = False
     """If True, output depth during training. Otherwise, only output depth during evaluation."""
+    rasterize_mode: Literal["classic", "antialiased"] = "classic"
+    """
+    Classic mode of rendering will use the EWA volume splatting with a [0.3, 0.3] screen space blurring kernel. This
+    approach is however not suitable to render tiny gaussians at higher or lower resolution than the captured, which
+    results "aliasing-like" artifacts. The antialiased mode overcomes this limitation by calculating compensation factors
+    and apply them to the opacities of gaussians to preserve the total integrated density of splats.
+
+    However, PLY exported with antialiased rasterize mode is not compatible with classic mode. Thus many web viewers that
+    were implemented for classic mode can not render antialiased mode PLY properly without modifications.
+    """
 
 
 class SplatfactoModel(Model):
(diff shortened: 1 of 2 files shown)


ichsan2895:
`truck` dataset:
**Left =** Using `--pipeline.model.rasterize-mode 'standard'`
PSNR = 26,19 dB

**Right =** using `Splatfacto v1.0.1`, so it don't use this rasterization.
PSNR = 25,97 dB

image
image
image
image

jb-ye:
Currently only nerfstudio viewer is supported to display the results correctly. Most other viewers which were designed for the classic mode must accommodate the same change to show the right the results of this new feature.

ichsan2895:
> Currently only nerfstudio viewer is supported to display the results correctly. Most other viewers which were designed for the classic mode must accommodate the same change to show the right the results of this new feature.

for `truck` and `kitchen` dataset that I have shown before, it was a frame from video which come from `ns-render` result. So, it does not screenshot the 3DGS Web viewer.

for `playroom` dataset, yes it is the screenshot from `mkellogg` 3D WebGL Gaussian Splat.

kerrj (approved):
sgtm!

jb-ye:
The new mode name is called `antialiased`

f-dy:
IMHO this should become the default, because:
- "antialiased" hold a closer-to-reality physical representation of the scene
- it doesn't hurt eval results when eval images are at the same distance as train images (there's even a slight improvement)
- it would certainly improve eval results for images that are closer or farther

Note that the best test data for this is to use something with a fine or regular texture (eg coarse-thread fabric, things with grids).

As we are getting closer to a physical representation, I hope that my proposed change on working in linear space rather than gamma-compressed (https://github.com/nerfstudio-project/nerfstudio/issues/925) will have a visible impact some day.

f-dy:
The fact that third-party viewers don't render 2D Gaussians correctly is their own issue (some code don't even have the +0.3 trick)

gonzalle:
Is it the same rasterization than [this solution](https://niujinshuchong.github.io/mip-splatting-demo/) ???
it has a viewer based on mkkellog's one...  (unfortunatly, unavailable in the repo)

jb-ye:
@gonzalle It is not exactly the same implementation as mip-splatting, but is in similar spirit and simpler. I don't think you can use that viewer to display PLY exports from splatfacto.

jinhwanlazy:
@jb-ye Could you please clarify why the opacity compensation is not being applied to the rasterization of the depth image?

https://github.com/nerfstudio-project/nerfstudio/blob/29ddd2ca3a33c43fb8d44b90d2758aaba0436f57/nerfstudio/models/splatfacto.py#L814

jb-ye:
@jinhwanlazy Nice catch. I think you are right, this is a bug we have to fix. Could you send out a PR to fix it?

PR #2889, opened 2024-02-09
Add transparency carving to splatfacto

pa-la:
### WHY 
This PR is similar in spirit to https://github.com/nerfstudio-project/nerfstudio/pull/2165 and https://github.com/nerfstudio-project/nerfstudio/pull/2025 for Nerfacto.

It implements the transparency trick which allows to carve a gaussian splatting scene when the masks are known.

### SOLUTION

The process is well described by the aforementioned PR:
> 

    1. Make everything except for the foreground object transparent in each image.
    2. While rendering, pick a random color in each iteration, then use this random color as background during alpha compositing.
    3. When computing the RGB loss, make the color of transparent pixels match the randomly chosen color of the current iteration used for rendering.

Here is a comparison before/after on the lego scene from the Nerf Blender dataset:

https://github.com/nerfstudio-project/nerfstudio/assets/156442943/4381fe77-690a-4f6a-ab8b-67121e875227

Because white gaussians are visible from the evaluation views when this trick is not implemented, the original PSNR is 12dB. After implementing this trick, the PSNR becomes 33dB.

P.S for this trick to work, `alpha_color` needs to be set to `None` so that the dataloader keeps the alpha channel. On the blender dataparser, this parameter defaults to `white`. I can write a separate PR for that if needed. Training is then launched with `ns-train splatfacto --pipeline.model.background_color random blender-data --data /path/to/lego`

### DIFF
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -672,7 +672,7 @@ class SplatfactoModel(Model):
                 rgb = background.repeat(int(camera.height.item()), int(camera.width.item()), 1)
                 depth = background.new_ones(*rgb.shape[:2], 1) * 10
                 accumulation = background.new_zeros(*rgb.shape[:2], 1)
-                return {"rgb": rgb, "depth": depth, "accumulation": accumulation}
+                return {"rgb": rgb, "depth": depth, "accumulation": accumulation, "background": background}
         else:
             crop_ids = None
         camera_downscale = self._get_downscale_factor()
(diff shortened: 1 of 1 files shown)


f-dy:
Note that this is using scale 2.6 for the random pointcloud initialization, not the default value of 10 (see https://github.com/nerfstudio-project/nerfstudio/pull/2902#issuecomment-1939154216)

kerrj:
It might be simpler to just supervise the output accumulation from splatfacto with the alpha channel of the PNG directly if available, which would get rid of all the background color logic and just force transparent regions to have 0 opacity when rendered. Is there any reason this would be less desirable than compositing+supervising RGB like here?

kerrj:
Interesting! That makes sense, although it does feel a little clunky to output the background color from the outputs, is there any clean way to avoid this? It looks like the version in nerfacto generates a random color during the train step before compositing, but this would also be a bit confusing to have 2 versions of backgrounds floating around...

kerrj:
Hm I think returning is clearer than having a mutable variable, I'll approve for now but might be nice to refactor at some point. Thanks!

kerrj (approved):
lgtm!

PR #2890, opened 2024-02-09
[Fix] convert camera_idx to tensor

tobiasfshr:
Add a conversion of the camera index to tensor as the forward function expects `Int[Tensor, "camera_indices"]` as input.


maturk (approved):
lgtm!

Issue #2863, opened 2024-02-01
Splatfacto new POV image quality


pierremerriaux-leddartech:
Hi,
It not really a bug, but more a question about splatfacto image quality/parameters. If it is not the right place, just let me know.
I am working to reconstruct driving scene from pandaset. Camera poses are along a car trajectory. When I render an image from this trajectory (train or eval set) the quality is very good.
Below a example of one image from eval set:
![image](https://github.com/nerfstudio-project/nerfstudio/assets/42007976/d7adfd12-1b29-41e2-b3ab-b32c9a70e55b)

But if I move a bit the target vehicle and the ego camera (1m on left and 0.5m up), the camera direction stays very close to the original one. 
![image](https://github.com/nerfstudio-project/nerfstudio/assets/42007976/eda163c5-743e-4d5f-b395-b447e44ef89f)
The quality decrease very quickly. The quality is very sensitive to direction. 
I tried to reduce sh degree from 3 to 1, to avoid to much overfitting, with no really improvement. 
Sure that a driving scene images is less dense, than a unique object scene with video sequence around it. But few months ago, I did it with nerfacto, the quality from initial poses was less, but really less sensible to new POV. Below a short video of nerfacto reconstruction from 5 cameras (front, side, and front-side) on the vehicle:

https://github.com/nerfstudio-project/nerfstudio/assets/42007976/8b347e4e-325c-401e-bed8-27896f160cc1

I improved my results with nerfacto by using camera pose opt. So I tried to do it for splatfacto by cherry pick viewmat gradient backward from gsplat branch https://github.com/nerfstudio-project/gsplat/tree/vickie/camera-grads. But for now, it didn't improve it.
In the same way, if I tried with multi cameras (or just a side camera in place of the front), the quality is very less impressive. Below example with 5 cameras and cameras optimization (but I seem to not have a big influence).
Left view:
![image](https://github.com/nerfstudio-project/nerfstudio/assets/42007976/a59c3f2a-57a9-4473-bfb7-a896c201ee24)
Front view
![image](https://github.com/nerfstudio-project/nerfstudio/assets/42007976/31e3b7b3-d19d-43cc-9927-fbae149246d7)

Do you have any idea, about what I should try firstly to improve new POV synthesis and secondly to improve multi cameras reconstruction?
And did someone work one camera pose optimization for SplatFacto ?

And just for the fun, below an example of what we can do with objects and multiples splatFacto instance:

[...]

PR #2891, opened 2024-02-09
Camera pose optimization for Splatfacto

oseiskar:
_Updated description, April 7_: Camera pose optimization for Splatfacto, based on https://github.com/nerfstudio-project/gsplat/pull/127, which is included in gsplat 0.1.9+. See https://github.com/nerfstudio-project/nerfstudio/pull/2891#issuecomment-2041027334 for instructions and examples.

.
_Original description, outdated:_ ~This PR is intended as an extension of https://github.com/nerfstudio-project/nerfstudio/pull/2885, but I could not easily create a PR to that branch, so creating a new PR instead. A detailed description in a comment in the original PR: https://github.com/nerfstudio-project/nerfstudio/pull/2885#issuecomment-1936410849~
~@jh-surh : Feel free to merge / cherry-pick any of my commits in this branch to your original PR, [...]~

### DIFF
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -686,8 +686,10 @@ method_configs["splatfacto-big"] = TrainerConfig(
         },
         "quats": {"optimizer": AdamOptimizerConfig(lr=0.001, eps=1e-15), "scheduler": None},
         "camera_opt": {
-            "optimizer": AdamOptimizerConfig(lr=1e-3, eps=1e-15),
-            "scheduler": ExponentialDecaySchedulerConfig(lr_final=5e-5, max_steps=30000),
+            "optimizer": AdamOptimizerConfig(lr=1e-4, eps=1e-15),
+            "scheduler": ExponentialDecaySchedulerConfig(
+                lr_final=5e-7, max_steps=30000, warmup_steps=1000, lr_pre_warmup=0
+            ),
         },
     },
     viewer=ViewerConfig(num_rays_per_chunk=1 << 15),
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -632,8 +632,10 @@ method_configs["splatfacto"] = TrainerConfig(
         },
         "quats": {"optimizer": AdamOptimizerConfig(lr=0.001, eps=1e-15), "scheduler": None},
         "camera_opt": {
-            "optimizer": AdamOptimizerConfig(lr=1e-3, eps=1e-15),
-            "scheduler": ExponentialDecaySchedulerConfig(lr_final=5e-5, max_steps=30000),
+            "optimizer": AdamOptimizerConfig(lr=1e-4, eps=1e-15),
+            "scheduler": ExponentialDecaySchedulerConfig(
+                lr_final=5e-7, max_steps=30000, warmup_steps=1000, lr_pre_warmup=0
+            ),
         },
     },
     viewer=ViewerConfig(num_rays_per_chunk=1 << 15),
--- a/nerfstudio/cameras/camera_optimizers.py
+++ b/nerfstudio/cameras/camera_optimizers.py
@@ -151,15 +151,20 @@ class CameraOptimizer(nn.Module):
             raybundle.origins = raybundle.origins + correction_matrices[:, :3, 3]
             raybundle.directions = torch.bmm(correction_matrices[:, :3, :3], raybundle.directions[..., None]).squeeze()
 
-    def apply_to_camera(self, camera: Cameras) -> None:
-        """Apply the pose correction to the raybundle"""
-        if self.config.mode != "off":
-            assert camera.metadata is not None, "Must provide id of camera in its metadata"
-            assert "cam_idx" in camera.metadata, "Must provide id of camera in its metadata"
-            camera_idx = camera.metadata["cam_idx"]
-            adj = self(torch.tensor([camera_idx], dtype=torch.long, device=camera.device))  # type: ignore
-            adj = torch.cat([adj, torch.Tensor([0, 0, 0, 1])[None, None].to(adj)], dim=1)
-            camera.camera_to_worlds = torch.bmm(camera.camera_to_worlds, adj)
+    def apply_to_camera(self, camera: Cameras) -> torch.Tensor:
+        """Apply the pose correction to the world-to-camera matrix in a Camera object"""
+        if self.config.mode == "off":
+            return camera.camera_to_worlds
+
+        assert camera.metadata is not None, "Must provide id of camera in its metadata"
+        if "cam_idx" not in camera.metadata:
+            # Evalutaion cams?
+            return camera.camera_to_worlds
+
+        camera_idx = camera.metadata["cam_idx"]
+        adj = self(torch.tensor([camera_idx], dtype=torch.long, device=camera.device))  # type: ignore
+        adj = torch.cat([adj, torch.Tensor([0, 0, 0, 1])[None, None].to(adj)], dim=1)
+        return torch.bmm(camera.camera_to_worlds, adj)
 
     def get_loss_dict(self, loss_dict: dict) -> None:
         """Add regularization"""
--- a/nerfstudio/cameras/camera_optimizers.py
+++ b/nerfstudio/cameras/camera_optimizers.py
@@ -182,10 +182,12 @@ class CameraOptimizer(nn.Module):
     def get_metrics_dict(self, metrics_dict: dict) -> None:
         """Get camera optimizer metrics"""
         if self.config.mode != "off":
-            metrics_dict["camera_opt_translation_max"] = self.pose_adjustment[:, :3].norm(dim=-1).max()
-            metrics_dict["camera_opt_translation_mean"] = self.pose_adjustment[:, :3].norm(dim=-1).mean()
-            metrics_dict["camera_opt_rotation_mean"] = numpy.rad2deg(self.pose_adjustment[:, 3:].norm(dim=-1).mean())
-            metrics_dict["camera_opt_rotation_max"] = numpy.rad2deg(self.pose_adjustment[:, 3:].norm(dim=-1).max())
+            trans = self.pose_adjustment[:, :3].detach().norm(dim=-1)
+            rot = self.pose_adjustment[:, 3:].detach().norm(dim=-1)
+            metrics_dict["camera_opt_translation_max"] = trans.max()
+            metrics_dict["camera_opt_translation_mean"] = trans.mean()
+            metrics_dict["camera_opt_rotation_mean"] = numpy.rad2deg(rot.mean().cpu())
+            metrics_dict["camera_opt_rotation_max"] = numpy.rad2deg(rot.max().cpu())
 
     def get_param_groups(self, param_groups: dict) -> None:
         """Get camera optimizer parameters"""
--- a/nerfstudio/cameras/camera_optimizers.py
+++ b/nerfstudio/cameras/camera_optimizers.py
@@ -22,6 +22,7 @@ import functools
 from dataclasses import dataclass, field
 from typing import Literal, Optional, Type, Union
 
+import numpy
 import torch
 import tyro
 from jaxtyping import Float, Int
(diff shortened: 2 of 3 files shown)


ichsan2895:
Again, interesting PR. :+1: 
How about the result quality? Hopefully it increase the quality when using `colmap`.

Is it compatible with latest `gsplat==0.1.5` ? or we need git install from this repo https://github.com/nerfstudio-project/gsplat/pull/127 ?

ichsan2895:
**Sorry**, Unfortunatelly, it decrease the quality:

`Playroom` dataset processed by `colmap 3.8`: (Red line = this PR)
image
image
Camera optimization regularizer acts after 3k iters, then the PSNR begins to drop.

`desolation` dataset provided by `ns-download-data`: (Red line = this PR)
image
Camera optimization regularizer acts after 3k iters, then the PSNR begins to drop.

oseiskar:
> Sorry, Unfortunatelly, it decrease the quality:

This is expected. Like I commented here https://github.com/nerfstudio-project/nerfstudio/pull/2885#issuecomment-1936410849, there are two separate cases:
 1. The current NeRF/3DGS benchmark datasets which are still images taken on a tripod. For these COLMAP works very accurately and there's no need to further optimize the poses. If pose optimization is enabled it easily decreases quality. It might be possible to tune the hyperparameters so that this would not happen, but I have not attempted that
 2. Datasets collected with a moving mobile camera, similar to the one @jh-surh was testing with in the [other PR](https://github.com/nerfstudio-project/nerfstudio/pull/2885). The goal of this PR is to improve these cases, especially in scenarios, where COLMAP has been replaced with a faster alternative like ARKit's online poses or other systems. The issue is that there is no good standard benchmark set for these at the moment (as far as I know)

I now disabled the feature by default so that the assumed target use case remains as 1. (2. can be addressed by re-enabling this on the command line as `--pipeline.model.camera-optimizer.mode=SO3xR3`)

> Again, interesting PR. 👍 How about the result quality? Hopefully it increase the quality when using `colmap`.
> 
> Is it compatible with latest `gsplat==0.1.5` ? or we need git install from this repo [nerfstudio-project/gsplat#127](https://github.com/nerfstudio-project/gsplat/pull/127) ?

No. Need to install from Git since the view matrix gradients from https://github.com/nerfstudio-project/gsplat/pull/127 (or, alternatively, https://github.com/nerfstudio-project/gsplat/pull/123) are neeed.

ichsan2895:
> > Sorry, Unfortunatelly, it decrease the quality:
> 
> This is expected. Like I commented here [#2885 (comment)](https://github.com/nerfstudio-project/nerfstudio/pull/2885#issuecomment-1936410849), there are two separate cases:
> 
>     1. The current NeRF/3DGS benchmark datasets which are still images taken on a tripod. For these COLMAP works very accurately and there's no need to further optimize the poses. If pose optimization is enabled it easily decreases quality. It might be possible to tune the hyperparameters so that this would not happen, but I have not attempted that
> 
>     2. Datasets collected with a moving mobile camera, similar to the one @jh-surh was testing with in the [other PR](https://github.com/nerfstudio-project/nerfstudio/pull/2885). The goal of this PR is to improve these cases, especially in scenarios, where COLMAP has been replaced with a faster alternative like ARKit's online poses or other systems. The issue is that there is no good standard benchmark set for these at the moment (as far as I know)
> 
> 
> I now disabled the feature by default so that the assumed target use case remains as 1. (2. can be addressed by re-enabling this on the command line as `--pipeline.model.camera-optimizer.mode=SO3xR3`)
> 
> > Again, interesting PR. 👍 How about the result quality? Hopefully it increase the quality when using `colmap`.
> > Is it compatible with latest `gsplat==0.1.5` ? or we need git install from this repo [nerfstudio-project/gsplat#127](https://github.com/nerfstudio-project/gsplat/pull/127) ?
> 
> No. Need to install from Git since the view matrix gradients from [nerfstudio-project/gsplat#127](https://github.com/nerfstudio-project/gsplat/pull/127) (or, alternatively, [nerfstudio-project/gsplat#123](https://github.com/nerfstudio-project/gsplat/pull/123)) are neeed.

Thanks for your clarification, in my experiment before, I already use gsplat from this source:
```
git clone --branch approximate-viewmat-gradient https://github.com/SpectacularAI/gsplat
git submodule update --init --recursive
```

I think it is good if the camera opt can detect automatically if the poses already accurate or not. If it already accurate and it hurts quality, it will be turn off.

But I am sorry I can not coding since it too advance for me.

kerrj:
I think these changes look reasonable to me, delaying pose optimization until partial convergence makes sense. @jh-surh   have you had a chance to take a look?

jb-ye:
Regarding quantitative validation of this PR, Here is my suggestion:

(1) take a dataset which you think is suitable for pose optimization
(2) run a baseline with pose optimization disabled and obtain PSNR metrics on validation images.
(3a) run with pose optimization enabled for all frames and extract the optimized poses
(3b) use the optimized poses to re-run the same setting as done in (2) and compare the PSNR metrics on validation images.

If (3b) is better than (2), one has strong evidence that pose optimization works. 

What do you think? @oseiskar

jb-ye:
> I think these changes look reasonable to me, delaying pose optimization until partial convergence makes sense. @jh-surh have you had a chance to take a look?

Delay pose optimization until partial convergence might be useful for fixing a few bad local poses, but would not correct systematic drift in SfM/SLAM.

oseiskar:
> Regarding quantitative validation of this PR, Here is my suggestion:
> 
> (1) take a dataset which you think is suitable for pose optimization (2) run a baseline with pose optimization disabled and obtain PSNR metrics on validation images. (3a) run with pose optimization enabled for all frames (basically all images are used for training) and extract the optimized poses (3b) use the optimized poses to re-run the same setting as done in (2) and compare the PSNR metrics on validation images.
> 
> If (3b) is better than (2), one has strong evidence that pose optimization works.
> 
> What do you think? @oseiskar

I think this makes a lot of sense. I am planning to do that in the near-ish future.

To clarify, this PR is not yet in the state where it would be robust enough for general use. The main intention was to help https://github.com/nerfstudio-project/nerfstudio/pull/2885 so that it's not necessary to reinvent the wheel in all aspects there. In particular, I think if this commit https://github.com/nerfstudio-project/nerfstudio/pull/2891/commits/5fe47c9f9e933de34159f68560f5d403ffc8fe69 is probably a pre-requisite for getting anything working well. These new metrics https://github.com/nerfstudio-project/nerfstudio/pull/2891/commits/3b02a7f67707b723b622566aedd0773f225c69e6 can also be helpful for debugging/monitoring pose optimization.

jh-surh:
> I think these changes look reasonable to me, delaying pose optimization until partial convergence makes sense. @jh-surh have you had a chance to take a look?

I think the changes look good. I'll try them out.


> Regarding quantitative validation of this PR, Here is my suggestion:
> 
> (1) take a dataset which you think is suitable for pose optimization (2) run a baseline with pose optimization disabled and obtain PSNR metrics on validation images. (3a) run with pose optimization enabled for all frames (basically all images are used for training) and extract the optimized poses (3b) use the optimized poses to re-run the same setting as done in (2) and compare the PSNR metrics on validation images.
> 
> If (3b) is better than (2), one has strong evidence that pose optimization works.
> 
> What do you think? @oseiskar

Wouldn't (3a) showing a better PSNR already show that the pose optimization works?

jb-ye:
> Wouldn't (3a) showing a better PSNR already show that the pose optimization works?

(3a) uses all images to train, so the metrics calculated on the set of validation images is biased. Evaluation on hold out set of images for calculating metrics like (3b) is unbiased. 

In (3b), image data of validation sets are only used to optimize their poses, but not gaussian splat models.

MartinEthier:
> Regarding quantitative validation of this PR, Here is my suggestion:
> 
> (1) take a dataset which you think is suitable for pose optimization (2) run a baseline with pose optimization disabled and obtain PSNR metrics on validation images. (3a) run with pose optimization enabled for all frames (basically all images are used for training) and extract the optimized poses (3b) use the optimized poses to re-run the same setting as done in (2) and compare the PSNR metrics on validation images.
> 
> If (3b) is better than (2), one has strong evidence that pose optimization works.
> 
> What do you think? @oseiskar

I've been looking into ways to enable evaluation using pose optimization and I tried the method you outlined using nerfacto and nerfacto-big. For some reason, my results on the extracted optimized poses are worse than simply not using pose optimization, even though the training PSNR increases when I enable pose optimization, indicating that it helps. See my comment here for more details: https://github.com/nerfstudio-project/nerfstudio/issues/2863#issuecomment-1936631197. Do you have any ideas why this might be the case?

oseiskar:
@ichsan2895 @jb-ye I now rebased this on top of the current main branch (with #3019), removed deprecated workarounds and changed the hyperparameters to what we used in [Gaussian Splatting on the Move](https://spectacularai.github.io/3dgs-deblur/).

Here's how to test branch and verify that it works: First, install and download a suitable test dataset (our version of Deblur-NeRF, which contains a variant with artificial pose noise):
```bash
conda activate nerfstudio

# download a suitable dataset
git clone https://github.com/SpectacularAI/3dgs-deblur
cd 3dgs-deblur
python download_data.py --dataset synthetic
cd ..

# check out and install this branch
https://github.com/SpectacularAI/nerfstudio
cd nerfstudio
git checkout camera-pose-optimization
pip install -e .
```

Then, run a basline test with the noisy pose data but without pose optimization:

    ns-train splatfacto --data ../3dgs-deblur/data/inputs-processed/synthetic-posenoise/cozyroom

Then check Viser: Looks terrible (Figure 1). Now, let's run another test with pose optimization:

    ns-train splatfacto --data ../3dgs-deblur/data/inputs-processed/synthetic-posenoise/cozyroom \
        --pipeline.model.camera-optimizer.mode=SO3xR3

Check viser again: Looks great (Figure 2). (Bonus: also add `--pipeline.model.rasterize-mode=antialiased` for best results)

Now, the caveats:

 1. Evaluating this numerically is not trivial, for the following reasons:
     a) We would also need to optimize the evaluation image poses or most of the error in, e.g., PSNR is caused by just "misaligned" evaluation images. This topic is also discussed by @LingzheZhao [here](https://github.com/WU-CVGL/BAD-Gaussians/issues/3#issuecomment-2013160604).
     b) Even if we do, cases with very well-optimized poses probably degrade (this concrens of most academic benchmark data out there, which is mostly consists of still images taken on tripods, or synthetics). This is because we waste some iterations that could be spent refining the 3D model into refining the poses, especially early in the optimization and during this time, when camera poses can move a lot, the 3D model remains blurrier than without pose opt. This can be solved with a 2-pass approach, which we also used in the paper.
     c) If evaluation images are used during training, even in the first pass, one could argue that they improve the model and the results are overly optimistic. To avoid this, the two-pass optimization should actually work like this: 1st pass: train normally without evaluation images, 2nd pass: optimize evaluation image poses with fixed Gaussian parameters. I have not tried to implement this since I think it's really a bit overkill in this context.

 2. The pose optimization _can_ fail depending on the scene. This is sensitive to the following parameters: `pipeline.model.camera-optimizer.trans-l2-penalty` and `auto-scale-poses` (dataparser), which is relevant if the dataset has absolute metric scale (e.g., from VISLAM). It should be possible to improve this in the future with more carefully designed priors, e.g., penalizing maxium deviation from the initial poses more strongly.

In any case, I think this PR is now in a mergeable state despite the (minor) weaknesses above. The pose optimization is now anyway off by default.

### Figure 1: Baseline

image


### Figure 2: With pose optimization

image

This was actually with `--pipeline.model.rasterize-mode=antialiased`. [Here's](https://github.com/nerfstudio-project/nerfstudio/assets/1812432/1c254804-3987-4f17-a21a-f036f4ad0624) how it looks with the default rasterization. Still better than with noisy poses, but the new rasterization gives a nice super-resolution capability here (training data is 600x400)

jb-ye:
@oseiskar Thanks, I will try to take a closer look in the next few days.

nerfstudio/configs/method_configs.py
```diff
@@ -632,8 +632,10 @@
         },
         "quats": {"optimizer": AdamOptimizerConfig(lr=0.001, eps=1e-15), "scheduler": None},
         "camera_opt": {
-            "optimizer": AdamOptimizerConfig(lr=1e-3, eps=1e-15),
-            "scheduler": ExponentialDecaySchedulerConfig(lr_final=5e-5, max_steps=30000),
+            "optimizer": AdamOptimizerConfig(lr=1e-4, eps=1e-15),
```
jb-ye, line 635: Could you add this change to "splatfacto-big" config?

jb-ye (commented):
Looks good to me, some minor comments.

ichsan2895:
Hello, interesting PR to be implemented.

Here is my own dataset (already processed with colmap) with some of motion blur. Already processed by `colmap` & `colmap image_undistorter` Feel free for testing it.

[Lubuk_torok_scenery](https://drive.usercontent.google.com/download?id=1bM6JFvROo4c6xZP_kq_EIhIg23yESoAy&export=download)

oseiskar:
@jb-ye : Done. I tested very briefly (i.e., only with this same synthetic scene) that it looks reasonable and better than the old splatfacto-big camera pose opt. defaults. This was the case (result [here](https://github.com/nerfstudio-project/nerfstudio/assets/1812432/52325c40-8699-41c5-9d32-56bddc601c4c)) so these are probably a good starting point for further hyperparameter tuning.

@ichsan2895 To clarify, this PR does not implement motion blur compensation. This is just one of the building blocks that can help in that goal and other tasks as well. For testing the _Gaussian Splatting on the Move_ method with custom data, see [here](https://github.com/SpectacularAI/3dgs-deblur?tab=readme-ov-file#training-with-custom-data).

lxzbg:
@oseiskar Great work. I used ARkit's pose based data for testing, and it did work. 

However, I have some observations. 
1. When the pose error is small, it can be corrected very well; 
2. When the pose error is large, it cannot be completely corrected (I tried adjusting the number of iterations, learning rate, and warnup, but this only had a little effect); 
3. The pose optimization is more accurate in the surrounding shooting area, but in the peripheral area In slightly distant areas, it seems that pose optimization is not that effective (maybe this observation is incorrect). 

Is there anything that can be improved here?

jb-ye:
@oseiskar You need to rebase and the PR should be auto-merged.

oseiskar:
@jb-ye Great! I now rebased but auto-merge seems to have been auto-disabled

Issue #2849, opened 2024-01-30
[POTENTIAL BUG] Using ns-process-data gives worse result than convert to undistorted images to colmap


ichsan2895:
**Describe the bug**
I have two different test how to create splatfacto dataset.

First, using `ns-process-data images`

Second, using `python3 convert.py` from INRIA Gaussian-splatting's github repo. It do not use any code from GS so it safe for using it as colmap license. It uses colmap source code for extracting undistorted images and SfM information from input images. 

Second result is always gives better result. In PSNR and SSIM. I know this way from @jb-ye and he pointed out maybe a bug in nerfstudio dataparser.

## **To Reproduce first dataset called `Vasedeck`**
**First way:**

Get `vasedeck` dataset from `nerf_real_360.zip` in this site (https://drive.google.com/drive/folders/128yBriW1IG_3NJ5Rp7APSTZsJqdJdfc1). Then copy paste the images to `IMAGES` folder.
```
ns-process-data images --data IMAGES --output-dir VASEDECK

ns-train splatfacto --logging.steps-per-log 200 --vis viewer+wandb --viewer.websocket-port 7007 \
    nerfstudio-data --load-3D-points True \
    --data VASEDECK --downscale-factor 1
```

**Second Way:**

create a folder called `VASEDECK`, and a subfolder called `input` which contains the images.

Go to INRIA GS's github repo, download it, and run `convert.py`.
```
python3 convert.py -s VASEDECK

ns-train splatfacto --data VASEDECK \
    --logging.steps-per-log 200 --vis viewer+wandb --viewer.websocket-port 7007 \
    colmap --colmap_path VASEDECK/sparse/0  --downscale-factor 1
```
**Expected behavior**
Second way becomes default way in `ns-process-data images` since it was better for splat quality. In this dataset, **I get extra 0,9 dB PSNR with second way in this dataset. It gains PSNR from 23,9 to 24,8 dB.**

**Screenshots**
![image](https://github.com/nerfstudio-project/nerfstudio/assets/60387342/6a500f01-4395-49ea-8650-7662c7bc1580)
![image](https://github.com/nerfstudio-project/nerfstudio/assets/60387342/187e7ab6-8169-4ed7-809f-ce01b1c6887e)

First way (left) and second way (right) as seen with `mkellogg` gaussian splat viewer

**Additional context**

- Uses `runpod` environment with `Ubuntu` 22.04.
- `Python 3.10.16`
[...]

PR #2896, opened 2024-02-10
Update process_data.py to accomodate ns-process-data splatfacto

ichsan2895:
### **Why add this `ns-process-data splatfacto`?**
It mimics `python3 convert.py -s /path/to/input/folder` from Original Inria 3DGS

### **What is special things about `python3 convert.py`?**
1. original `ns-process-data images` make "images" in output folder become smaller despite has same resolution, it hurts quality.
2. It uses `exhausive_matcher` instead of `vocab_tree_matcher` in `ns-process-data images`
3. it uses hyperparameter of `ba_global_function_tolerance = 0.000001`
4. It uses `colmap image_undistorter`

### **So what I get if I use that?**
I got extra quality boost, about 0.3 - 1 dB in PSNR evaluation of all images. I tested it on different dataset. 

### **What is different between this implementation and `python3 convert.py.`?**
The original convert.py gives result of colmap format which does not have `transforms.json` and `sparse_pc.json`. This implementation creates that files so it compatible with `NerfstudioDataparser`.

Thanks for @jb-ye for give me inspiration of this PR.

### DIFF
--- a/nerfstudio/scripts/process_data.py
+++ b/nerfstudio/scripts/process_data.py
@@ -51,7 +51,7 @@ class ProcessRecord3D(BaseConverterToNerfstudioDataset):
 
     This script does the following:
 
-    1. Scales images to a specified snum_downscalesize.
+    1. Scales images to a specified size.
     2. Converts Record3D poses into the nerfstudio format.
     """
 
--- a/nerfstudio/scripts/process_data.py
+++ b/nerfstudio/scripts/process_data.py
@@ -504,7 +504,7 @@ class ProcessSplatfacto(BaseConverterToNerfstudioDataset):
             if self.eval_data is not None:
                 raise ValueError("Cannot use eval_data since cameras were already aligned with it.")
 
-            if skip_matching==False:
+            if self.skip_matching==False:
                 os.makedirs(output_path + "/distorted/sparse", exist_ok=True)
 
                 ## Feature extraction
--- a/nerfstudio/scripts/process_data.py
+++ b/nerfstudio/scripts/process_data.py
@@ -578,7 +578,7 @@ class ProcessSplatfacto(BaseConverterToNerfstudioDataset):
             
                         # Resize the image to half of its original resolution
                         height, width = img.shape[:2]
-                        resized_img = cv2.resize(img, (int(width * 1/factor), int(height * 0.5)))
+                        resized_img = cv2.resize(img, (int(width * 1/factor), int(height * 1/factor)))
             
                         # Write the resized image to the output folder
                         output_path = os.path.join(output_folder, file)
--- a/nerfstudio/scripts/process_data.py
+++ b/nerfstudio/scripts/process_data.py
@@ -28,6 +28,7 @@ import shutil
 
 import numpy as np
 import tyro
+import cv2
 from typing_extensions import Annotated
 
 from nerfstudio.process_data import (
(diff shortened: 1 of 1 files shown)


ichsan2895:
I constantly get better result with **Second Way**. Now I try another dataset.

### 1. `Truck` dataset from sample Inria 3DGS which I processed using `colmap 3.8`
**Blue line =** First Way
PSNR = 25,97
**Red line =** Second Way
PSNR = 26,66

image

### 2. `Purancak` dataset from my friend drone which I processed using `colmap 3.8`
**Blue line =** First Way
PSNR = 21,68

**Red line =** Second Way
PSNR = 22,47

image

### 3. `desolation` dataset from `ns-process-data`.
**Blue line =** First Way
PSNR = 21.62
**Red line =** Second Way
PSNR = 22.21

image

ichsan2895:
### 4. `desolation` dataset from `ns-download-data`.
**Blue line =** First Way
PSNR = 24.22

**Red line =** Second Way
PSNR = 24.97

image

kerrj:
Thanks for this! One thing to be careful of in metrics: since the ground truth images themselves are different (one is OpenCV undistorted, one is COLMAP undistorted), the metrics might not accurately reflect improvements since the base images have changed. It would be good to compare renders qualitatively to each other to confirm that there is in fact a quality improvement in the model itself.

Another note is I think it would be more elegant to include this within the standard COLMAP pipeline (eg `ns-process-data images`) and expose it as a parameter like `--use-colmap-undistort` or something. Having an additional pipeline involves more duplicated code and makes it harder to maintain.

ichsan2895:
> It would be good to compare renders qualitatively to each other to confirm that there is in fact a quality improvement in the model itself.

Thanks, I will try compare the result of qualitative image

But, let me ask question, why the size of ns-process-data images in "images" output folder is different than "images" in input folder. It has smaller size despite it has same resolution. I think it a little bit hurts the quality.

> think it would be more elegant to include this within the standard COLMAP pipeline (eg ns-process-data images) and expose it as a parameter like --use-colmap-undistort or something.

`ns-process-data images` uses `vocab_tree_matcher` by default, my implementation  uses `exhausive_,matcher` by default. It would be a conflict, isn't it?

kerrj:
> But, let me ask question, why the size of ns-process-data images in "images" output folder is different than "images" in input folder. It has smaller size despite it has same resolution. I think it a little bit hurts the quality.

Do you mean that the number of images is smaller? It's possible that COLMAP couldn't find poses for all the input images.

> `ns-process-data images` uses `vocab_tree_matcher` by default, my implementation uses `exhausive_,matcher` by default. It would be a conflict, isn't it?

This is exposed as a parameter in the data pipeline (see `matching_method` in `colmap_converter_to_nerfstudio_dataset.py`), so it could be toggled by a user if needed. I think we would like to keep the default to vocab_tree for speed, although I don't have much experience with the difference in speed/quality between the two methods.

jb-ye:
I agree with @kerrj suggestion. Let's first root cause the main factor that contributes to the quality difference observed in https://github.com/nerfstudio-project/nerfstudio/issues/2849

Once we know what is the major cause of the quality gap, we can reset default parameters in proper ways to reflect that or fix bugs if necessary.

ichsan2895:
Thank you for all attention. I will close my PR, but if somebody want to test `ns-process-data splatfacto`, please kindly git clone my fork of nerfstudio.

PR #2897, opened 2024-02-10
Call rescale_output_resolution before early return

pierotofy:
Hello :wave:

While looking through the splatfacto implementation, I noticed that in the case of an early exit in the forward pass due to a zero radii sum the camera does not get rescaled back to its original dimensions. I think this might be a mistake? I haven't found a dataset where the condition occurs, but it seems correct to make sure the camera is rescaled back before returning.

### DIFF
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -751,7 +751,7 @@ class SplatfactoModel(Model):
         camera.rescale_output_resolution(camera_downscale)
 
         if (self.radii).sum() == 0:
-            rgb = background.repeat(int(camera.height.item()), int(camera.width.item()), 1)
+            rgb = background.repeat(H, W, 1)
             depth = background.new_ones(*rgb.shape[:2], 1) * 10
             accumulation = background.new_zeros(*rgb.shape[:2], 1)
 
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -747,13 +747,13 @@ class SplatfactoModel(Model):
             tile_bounds,
         )  # type: ignore
 
+        # rescale the camera back to original dimensions before returning
+        camera.rescale_output_resolution(camera_downscale)
+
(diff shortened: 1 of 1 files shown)


chungmin99:
> in the case of an early exit in the forward pass due to a zero radii sum the camera does not get rescaled back

Thank you for catching it! This condition can actually happen quite frequently when there's no gaussians in front (e.g., due to cropping + moving the camera away from the cropbox). 

---

The fix looks great, but I'm just wondering if we can remove duplicate code, by moving the camera rescaling logic (L745, and L763) before the radii check (L738):

https://github.com/nerfstudio-project/nerfstudio/blob/c5520fb4b1ff45907f3a6a67acc76cdbcbf08ccb/nerfstudio/models/splatfacto.py#L737-L747

And we can change `rgb = background.repeat(int(camera.height.item()), int(camera.width.item()), 1)` to `rgb=background.repeat(H, W, 1)`. 

... unless there's any dependence on the `camera` object I'm missing between L750-760.

jb-ye:
Instead of adding this line to early return, could you move the line later to an earlier position: right after project_gaussian() call. We no longer need `camera` after that.

jb-ye:
You can test your code by enabling `--pipeline.model.num_downscales=2` because currently it is set to zero by default.

nerfstudio/models/splatfacto.py
```diff
@@ -747,10 +747,14 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
             tile_bounds,
         )  # type: ignore
 
+        # rescale the camera back to original dimensions before returning
+        camera.rescale_output_resolution(camera_downscale)
```
jb-ye, line 751: You need to remove the line 773-774

chungmin99: Ah yes, thanks for catching -- just fixed this

jb-ye:
lgtm

chungmin99 (approved):

PR #2899, opened 2024-02-10
Fix benchmarking script camera optimizer argument

gchhablani:
This PR fixes the camera optimizer argument in benchmarking scripts after #2092.


brentyi (approved):
lgtm, thanks!!

PR #2902, opened 2024-02-11
Add pointcloud loading for blender parser

gchhablani:
This PR adds pointcloud loading for blender parser.

The reason this is needed is because the [original gaussian splatting repo](https://github.com/graphdeco-inria/gaussian-splatting) supports blender format along with colmap.

With random initialization,`splatfacto` performance is not great (see below for lego).

image

[Here](https://github.com/graphdeco-inria/gaussian-splatting/blob/2eee0e26d2d5fd00ec462df47752223952f6bf4e/scene/gaussian_model.py#L124-L147) are the relevant lines of code where they initialize the gaussians from pointcloud if that helps.

**Note**:
- I am not sure if `transform_matrix` and `scaling_factor` are needed here. If yes, any pointers on how it should be done will help.

### DIFF
--- a/nerfstudio/data/dataparsers/blender_dataparser.py
+++ b/nerfstudio/data/dataparsers/blender_dataparser.py
@@ -21,6 +21,7 @@ from typing import Type
 
 import imageio
 import numpy as np
+import open3d as o3d
 import torch
 
 from nerfstudio.cameras.cameras import Cameras, CameraType
@@ -42,6 +43,11 @@ class BlenderDataParserConfig(DataParserConfig):
     """How much to scale the camera origins by."""
     alpha_color: str = "white"
     """alpha color of background"""
+    colmap_path: Path = Path("sparse_pc.ply")
+    """Path to the colmap reconstruction directory relative to the data path."""
+    load_3D_points: bool = True
+    """Whether to load the 3D points from the colmap reconstruction. This is helpful for Gaussian splatting and
+    generally unused otherwise, but it's typically harmless so we default to True."""
 
 
 @dataclass
@@ -64,6 +70,7 @@ class Blender(DataParser):
 
     def _generate_dataparser_outputs(self, split="train"):
         meta = load_from_json(self.data / f"transforms_{split}.json")
+        ply_path = self.config.data / self.config.ply_path
         image_filenames = []
         poses = []
         for frame in meta["frames"]:
@@ -93,6 +100,13 @@ class Blender(DataParser):
             cy=cy,
             camera_type=CameraType.PERSPECTIVE,
         )
+        
+        metadata = {}
+        if self.config.load_3D_points:
+            # Load 3D points
+                    # TODO: Check if we need to use `transform_matrix` and `scaling_factor`
+            metadata.update(self._load_3D_points(ply_path))
+
 
         dataparser_outputs = DataparserOutputs(
             image_filenames=image_filenames,
@@ -103,3 +117,16 @@ class Blender(DataParser):
         )
 
         return dataparser_outputs
+
+    def _load_3D_points(self, ply_file_path: Path):
+        pcd = o3d.io.read_point_cloud(str(ply_file_path))
+
+        points3D = torch.from_numpy(np.asarray(pcd.points, dtype=np.float32))
+        # TODO: Check if we need to use `transform_matrix` and `scaling_factor`
+        points3D_rgb = torch.from_numpy((np.asarray(pcd.colors) * 255).astype(np.uint8))
+
+        out = {
+            "points3D_xyz": points3D,
+            "points3D_rgb": points3D_rgb,
+        }
+        return out


brentyi:
Hi @gchhablani, thanks for the PR! Where are you getting the point cloud from? The point clouds are usually pulled from COLMAP, but we don't typically run COLMAP for synthetic data since the training images are just rendered from known poses.

For what it's worth, my `splatfacto` results with the default settings are also not great:
`ns-train splatfacto --data ./data/blender/lego blender-data`
image

But I can get much cleaner results by setting the background color of the model to white, which is a typical hack that NeRF methods evaluated on synthetic data use:
`ns-train splatfacto --data ./data/blender/lego --pipeline.model.background-color white blender-data`
image

brentyi (commented):
Hi @gchhablani, thanks for the PR! Where are you getting the point cloud from? The point clouds are usually pulled from COLMAP, but we don't typically run COLMAP for synthetic data since the training images are just rendered from known poses.

For what it's worth, my `splatfacto` results with the default settings are also not great:
`ns-train splatfacto --data ./data/blender/lego blender-data`
image

But I can get much cleaner results by setting the background color of the model to white, which is a typical hack that NeRF methods evaluated on synthetic data use:
`ns-train splatfacto --data ./data/blender/lego --pipeline.model.background-color white blender-data`
image

brentyi:
#2889 also seems related!

gchhablani:
@brentyi 
I agree, there might be ways to fix things for this particular dataset, but would it not be great to directly load ply files for pointcloud initializations for other blender format datasets?

gchhablani:
Btw, I get the pointcloud from the original nerf dataset.

brentyi:
> Btw, I get the pointcloud from the original nerf dataset.

Did you generate these yourself? I wasn't aware of any point clouds released in the original dataset, and can't seem to find them either (eg in this [Google Drive](https://drive.google.com/drive/folders/128yBriW1IG_3NJ5Rp7APSTZsJqdJdfc1)).

gchhablani:
@brentyi Thanks for asking me this :) 
I wouldn't have checked otherwise.

I downloaded the synthetic data using this:
```
wget http://cseweb.ucsd.edu/~viscomp/projects/LF/papers/ECCV20/nerf/nerf_example_data.zip
```
Not sure where I had found this.

However, I checked the pointcloud since you asked.
And this is the pointcloud:

image

Interestingly, I get slightly better performance with this random cube, but this could be random too, of course:

image

gchhablani:
Btw, setting `background-color` to `white` for `nerfacto` gives black-and-white results, I'm not sure why. Is this expected?

image

brentyi:
Assuming that's meant to be an RGB image: yeah, looks weird to me. It's interesting that the colors in the rendered eval image on the right look fine.

I'm not familiar with the tensorboard or wandb logging systems though.

gchhablani:
@brentyi Do you think this PR is worth pursuing?

ichsan2895:
Hello, seems good PR. Please let me know some my thought

If I use my own `blender` model, where can I get `sparse_pc.ply`?

If I manually use `ns-process-data images` (which utilizes `colmap`) to get `sparse_pc.ply`, of course I will use `nerfstudio dataparser` instead of `blender dataparser`

gchhablani:
@ichsan2895 @brentyi 

The idea is to have it consistent with the original repo. I have some synthetic data for which I do have pointcloud in `.ply` but random init will not work on it.

Here is my use case: 
- I have a set of images, corresponding intrinsic and extrinsics, and a pointcloud with holes in them. 
- I converted them to Blender format with scripts to run with `gaussian-splatting` repository because that was the easiest to do it.
- Now, to run things with `nerfstudio dataparser`, I will have to understand a new format and repeat the process. 

Would it not be easier to just have pointcloud support in Blender format?
Regardless, if this PR doesn't make sense, I'll close it.
However, I would greatly appreciate some pointers on how to integrate `transform_matrix` and `scaling_factor` so I can use it myself, at least.

f-dy:
@brentyi @gchhablani to get reasonable result with the blender dataset, you should set random_scale (here https://github.com/nerfstudio-project/nerfstudio/blame/main/nerfstudio/models/splatfacto.py#L144 ) to the right value. It is 2.6, not 10 in 3DGS, see here https://github.com/graphdeco-inria/gaussian-splatting/blob/main/scene/dataset_readers.py#L240 

That's something we noticed with @pa-la when working on https://github.com/nerfstudio-project/nerfstudio/pull/2889

I don't understand why it's 10 by default in nerfstudio? maybe for mipnerf360 scenes? I think we should set it to 2.6 by default, to be consistent with the original code.

nerfstudio/data/dataparsers/blender_dataparser.py
```diff
@@ -42,6 +43,11 @@ class BlenderDataParserConfig(DataParserConfig):
     """How much to scale the camera origins by."""
     alpha_color: str = "white"
     """alpha color of background"""
+    ply_path: Path = Path("sparse_pc.ply")
+    """Path to the pointcloud file relative to the data path."""
+    load_3D_points: bool = True
```
brentyi, line 48: can we replace these two fields with just one:

```python
ply_path: Optional[Path] = None
```

?

and indicate in the docstring that this is used for Gaussian splatting initialization?

brentyi (commented):
Sorry for the delay in getting back here! I've been hesitant here because it feels like the aim here would be to support non-standard dataset formats (AFAIK there's no public datasets in the Blender format that also provide point clouds).

But given how the uninvasive the changes are I'm fine with supporting it.

I think `transform_matrix` can be skipped (these are the camera poses? unless you're referring to something else), but the point should be scaled by `scale_factor`.

kerrj:
> I don't understand why it's 10 by default in nerfstudio? maybe for mipnerf360 scenes? I think we should set it to 2.6 by default, to be consistent with the original code.

We can change back to 2.6, the intention with setting to 10 was that if people use _real_ datasets without points available (eg Polycam), initializing to 10 would give reasonable results on backgrounds sometimes, while 2.6 would look bad. I'm fine changing the default to 2.6, since point pipelines for other data inputs should be added now anyways.

gchhablani:
@brentyi I have updated the PR. Requesting for another review.

jb-ye:
@brentyi @kerrj Any follow up on this? It looks like a simple non-invasive change.

brentyi:
Thanks for the bump; I fixed the scale factor and made the option opt-in. Should be okay with me now!

brentyi (approved):

PR #2910, opened 2024-02-13
Fix an issue where gradient accumulation could not be passed as argument due to a type error.

SharkWipf:
As discussed on Discord (Thanks Brent), `--gradient-accumulation-steps` did not properly cast its values to int and would error because it received a string.
There is still a separate bug where specifying `camera_opt` on nerfacto will cause an error, and the documentation on the --gradient-accumulation-steps could be better (accepted values are currently `proposal_networks`, `fields` and `camera_opts`, but there's no way to figure this out without without diving into the code AFAICT), but that's for another time/person/PR.


brentyi (approved):
looks great, thanks @SharkWipf!!

cc @kerrj for issue on gradient accumulation

PR #2919, opened 2024-02-16
Add nd-gaussians to Splatfacto

CardiacMangoes:
Adds support for rasterizing gaussians with extra dimensions

Adds parameters to gsplat:
- `use_nd` (bool)
- `nd_dim` (int)
- `nd_detach` (bool)

### DIFF
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -157,7 +157,7 @@ class SplatfactoModelConfig(ModelConfig):
     """
     output_depth_during_training: bool = False
     """If True, output depth during training. Otherwise, only output depth during evaluation."""
-    use_nd: bool = True # TODO: set this back to False. where to set this to true?
+    use_nd: bool = False
     """Whether to rasterize nd values."""
     nd_dim: int = 1
     """Dimension for nd rasterizing."""
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -157,6 +157,12 @@ class SplatfactoModelConfig(ModelConfig):
     """
     output_depth_during_training: bool = False
     """If True, output depth during training. Otherwise, only output depth during evaluation."""
+    use_nd: bool = False
+    """Whether to rasterize nd values."""
+    nd_dim: int = 8
+    """Dimension for nd rasterizing."""
+    nd_detach: bool = True
+    """Whether to detach gaussian from nd values."""
 
 
 class SplatfactoModel(Model):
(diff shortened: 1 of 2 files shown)


kerrj:
I know that internally we've used the term "nd" a lot, but it's not very precise from an outside perspective. Can we rename things to "render_features" or "render_latents" or something like that? then instead of "nd_values" it could output something like "feature_map" which is more clear what's going on.

nerfstudio/models/splatfacto.py
```diff
@@ -167,6 +167,12 @@ class SplatfactoModelConfig(ModelConfig):
     However, PLY exported with antialiased rasterize mode is not compatible with classic mode. Thus many web viewers that
     were implemented for classic mode can not render antialiased mode PLY properly without modifications.
     """
+    use_nd: bool = False
+    """Whether to rasterize nd values."""
```
kerrj, line 171: Expand on the docstrings a lot here, mention how each gaussians stores a feature vector of a certain dimension, and it gets rasterized as separate channels.

nerfstudio/models/splatfacto.py
```diff
@@ -660,17 +707,25 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
...
         assert camera.shape[0] == 1, "Only one camera at a time"
 
+        out = {}
+
         # get the background color
+        dims = self.config.nd_dim if self.config.use_nd else 3
         if self.training:
             if self.config.background_color == "random":
                 background = torch.rand(3, device=self.device)
+                nd_background = torch.rand(dims, device=self.device)
```
kerrj, line 717: Using random background for training only really makes sense for RGB since the gradients help opacity inflate into uncovered regions of the scene. I think the behavior for feature vectors should be to always use a torch.zeros background, then normalize with alpha similar to depth. (or maybe this step can be left out as optional if people want to add it)

nerfstudio/models/splatfacto.py
```diff
@@ -811,8 +874,24 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
                 background=torch.zeros(3, device=self.device),
             )[..., 0:1]  # type: ignore
             depth_im = torch.where(alpha > 0, depth_im / alpha, depth_im.detach().max())
+        if self.config.use_nd:
+            nd_values = self.get_nd_values()
+            nd_value = rasterize_gaussians(  # type: ignore
+                self.xys.detach() if self.config.nd_detach else self.xys,
+                depths,
```
kerrj, line 881: depths should also have the same `.detach()` logic

kerrj:
We should refactor this to integrate with #2946, it will simplify the logic.

PR #2920, opened 2024-02-16
fix crash when both mask and downscale_factor are set

jinhwanlazy:
Splatfacto crashes when mask is exist and num_downscale > 0, because downscale has not been applied to mask.

### DIFF
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -640,6 +640,7 @@ class SplatfactoModel(Model):
 
             # torchvision can be slow to import, so we do it lazily.
             import torchvision.transforms.functional as TF
+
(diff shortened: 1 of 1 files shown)


jb-ye:
Thanks for the catch. It looks good to me. 

May I ask when you found `num_downscale > 0` useful? I set it to zero because I didn't observe any benefits of that and the official Inria GS doesn't do it either.

jinhwanlazy:
I found no big difference in the end results, but it makes first training iterations sigificantly faster. So I think it like, why not?

jb-ye:
> I found no big difference in the end results, but it makes first training iterations sigificantly faster. So I think it like, why not?
Okay, I agree this is a bug to be fixed.

kerrj (approved):
thanks!

PR #2932, opened 2024-02-20
Adding data-processing support for the side grayscale fisheye lens on the Aria Glasses

AntonioMacaronio:


nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -118,25 +119,103 @@ def read_trajectory_csv_to_dict(file_iterable_csv: str) -> TimedPoses:
     )
 
 
+def undistort_image_and_calibration(
+    input_image: np.ndarray,
+    input_calib: calibration.CameraCalibration,
+    output_focal_length: int,
+) -> [np.ndarray, calibration.CameraCalibration]:
```
brentyi, line 126: is this a valid annotation? should it be `Tuple[np.ndarray, calibration.CameraCalibration]`?

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -118,25 +119,103 @@ def read_trajectory_csv_to_dict(file_iterable_csv: str) -> TimedPoses:
...
+    """
+    Return the rotated upright image and update both intrinsics and extrinsics of the camera calibration
+    NOTE: This function only supports pinhole and fisheye624 camera model.
+    """
+    output_image = np.rot90(input_image, k=3)
+    updated_calib = calibration.rotate_camera_calib_cw90deg(input_calib)
+
+    return output_image, updated_calib
+
+def generate_circular_mask(numRows: int, numCols: int, radius: float):
```
brentyi, line 167: return annotation would be nice!

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -76,11 +78,10 @@ class TimedPoses:
     t_world_devices: List[SE3]
 
 
-def get_camera_calibs(provider: VrsDataProvider) -> Dict[str, AriaCameraCalibration]:
+def get_camera_calibs(provider: VrsDataProvider, name="camera-rgb") -> AriaCameraCalibration:
```
brentyi, line 81: super nit: can we annotate the `name` with a `Literal` type? (also below)

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -118,25 +119,103 @@ def read_trajectory_csv_to_dict(file_iterable_csv: str) -> TimedPoses:
...
+    assert isinstance(src_calib, calibration.CameraCalibration), "src_calib is not of type CameraCalibration"
+    
+    # Get the image corresponding to this index and undistort it
     image_data = provider.get_image_data_by_index(stream_id, index)
-    img = Image.fromarray(image_data[0].to_numpy_array())
+    image_array, intrinsic = image_data[0].to_numpy_array().astype(np.uint8), [0, 0, 0, 0]
+    if pinhole:
+        f_length = 500 if camera_name == "camera-rgb" else 170
+        image_array, src_calib = undistort_image_and_calibration(image_array, src_calib, f_length)
+        intrinsic = [f_length, f_length, image_array.shape[1] // 2, image_array.shape[0] // 2]
```
brentyi, line 208: nit: can we use tuples instead if things don't need to be mutated?

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -68,6 +69,7 @@ class AriaImageFrame:
     file_path: str
     t_world_camera: SE3
     timestamp_ns: float
+    pinhole_intrinsic: List[int]
```
brentyi, line 72: maybe `tuple[float, float, float, float]`?

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -184,59 +292,126 @@ class ProcessProjectAria:
...
     """Path to the output directory."""
+    points_file: Optional[Path] = None
+    """Path to the point cloud file (usually called semidense_points.csv.gz) if not in the mps_data_dir"""
+    include_side_cameras: bool = False
+    """If True, include and process the images captured by the grayscale side cameras. If False, only uses the main RGB camera's data."""
+    vrs_file2: Optional[Path] = None
+    """Path to the second VRS file if provided"""
+    mps_data_dir2: Optional[Path] = None
+    """Path to the second MPS attachments if provided"""
+    points_file2: Optional[Path] = None
```
brentyi, line 303: it feels a little weird to me to hardcode things to only support 1 or 2 VRS files, instead of being able to do multiple

would it make sense to replace the original `vrs_file: Path` with `vrs_file: Tuple[Path, ...]`? as well as `mps_data_dir` with `mps_data_dir: Tuple[Path, ...]`?

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -184,59 +292,126 @@ class ProcessProjectAria:
     """Path to Project Aria Machine Perception Services (MPS) attachments."""
     output_dir: Path
     """Path to the output directory."""
+    points_file: Optional[Path] = None
+    """Path to the point cloud file (usually called semidense_points.csv.gz) if not in the mps_data_dir"""
```
brentyi, line 296: just wondering, when is the point cloud file to not in the MPS dir?

AntonioMacaronio: When I downloaded the Aria Everyday Activities Dataset, the pointcloud was in its own separate folder so i went ahead and added support for both cases

brentyi: aha that makes sense!

brentyi: is it one shared point cloud across all users?

AntonioMacaronio: actually it's multiple point cloud files, so rn main() will combine all of them since they should be using the same point of reference

PR #2931, opened 2024-02-19
Make time slider effective in viewer (Fixes #2928)

iSach:
This pull request makes the time slider in the Control panel of the viser viewer effective again, for dynamic models.

The Render panel still misses options to set a per-keyframe timestamp, but I will probably do it in a separate PR. This should do the job for now to monitor training.

For those interested, meanwhile, you can use the legacy viewer (`--vis viewer_legacy`) to create camera paths for dynamic scenes.

PR #2936, opened 2024-02-20
Full implementation of features for dynamic scenes into the Viser viewer.

iSach:
### Description

This pull request finalizes the feature fix in PR #2931.

To allow working with dynamic scenes on the new Viser viewer, the *Render* panel now allows for setting per-keyframe timestamps along with a default time value. When rendering, it will interpolate between the times of each keyframe (similar to FOV interpolation).

Also, unlike the legacy viewer which does not support previewing the dynamic render, this PR separates the default view (using the time in the Control panel) and the preview view so that one can preview (although blurry obviously :-)) their rendered dynamic scene!

### TL;DR

- The Control panel now has a slider for setting the time at which the (dynamic) scene is rendered.
- The Render panel now has a default time slider, which sets the default time for each keyframe.
- Each keyframe can override the default time value. At render (i.e., through the camera path JSON file), each frame between keyframes will interpolate the render time.
- The render preview now supports dynamic scenes, which allows for going into preview mode and see how the animated render will look like.

### Demo

https://github.com/nerfstudio-project/nerfstudio/assets/13520261/046cd76a-76fb-42b3-a232-ebaeb2a6c0ef

### DIFF
--- a/nerfstudio/viewer/render_panel.py
+++ b/nerfstudio/viewer/render_panel.py
@@ -149,6 +149,17 @@ class CameraPath:
                         initial_value=keyframe.override_time_val,
                         disabled=not keyframe.override_time_enabled,
                     )
+
+                    @override_time.on_update
+                    def _(_) -> None:
+                        keyframe.override_time_enabled = override_time.value
+                        override_time_val.disabled = not override_time.value
+                        self.add_camera(keyframe, keyframe_index)
+
+                    @override_time_val.on_update
+                    def _(_) -> None:
+                        keyframe.override_time_val = override_time_val.value
+                        self.add_camera(keyframe, keyframe_index)
                 delete_button = server.add_gui_button("Delete", color="red", icon=viser.Icon.TRASH)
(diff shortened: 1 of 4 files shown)


iSach:
Little fix incoming with JSON exports.

iSach:
Should be good now. :)

brentyi (approved):
looks good to me, thanks @iSach!!

nerfstudio/viewer/render_panel.py
```diff
@@ -902,8 +989,10 @@ def _(_) -> None:
                                 wxyz=pose.rotation().wxyz,
                                 # There are some floating point conversions between degrees and radians, so the fov and
                                 # default_Fov values will not be exactly matched.
-                                override_fov_enabled=abs(frame["fov"] - json_data.get("default_fov", 0.0)) < 1e-3,
+                                override_fov_enabled=abs(frame["fov"] - json_data.get("default_fov", 0.0)) > 1e-3,
```
brentyi, line 992: also thanks for fixing my bug 🎊

PR #2937, opened 2024-02-20
Update project, rasterize with block_width interface

kerrj:
In gsplat==0.1.6, the interface for tile_bounds is replaced by simply specifying a block size.


vye16 (approved):
lgtm

PR #2938, opened 2024-02-20
inference near plane option

ethanweber:


nerfstudio/models/splatfacto.py
```diff
@@ -764,6 +766,16 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
         else:
             rgbs = torch.sigmoid(colors_crop[:, 0, :])
 
+        if not self.training and self.config.inference_near_plane != 0.0:
```
kerrj, line 769: any update on doing this with the gsplat `project_gaussians` flag instead of manual masks?

PR #2939, opened 2024-02-21
Fix depth-nerfacto edge cases

ethanweber:
Depth nerfacto doesn't work unless num_rays_in_batch is divisible by 2. Also, the depth images are in a list and not just a batch tensor, so I think the changes I made are needed.

### DIFF
--- a/nerfstudio/data/pixel_samplers.py
+++ b/nerfstudio/data/pixel_samplers.py
@@ -272,9 +272,12 @@ class PixelSampler:
         # only sample within the mask, if the mask is in the batch
         all_indices = []
         all_images = []
+        all_depth_images = []
 
         if "mask" in batch:
             num_rays_in_batch = num_rays_per_batch // num_images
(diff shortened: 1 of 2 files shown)


nerfstudio/data/pixel_samplers.py
```diff
@@ -272,9 +272,12 @@ def collate_image_dataset_batch_list(self, batch: Dict, num_rays_per_batch: int,
         # only sample within the mask, if the mask is in the batch
         all_indices = []
         all_images = []
+        all_depth_images = []
 
         if "mask" in batch:
             num_rays_in_batch = num_rays_per_batch // num_images
+            if num_rays_in_batch % 2 != 0:
```
kerrj, line 279: this could be moved above if statement and 293-294 deleted right?

kerrj (approved):
other than small fix, lgtm

akristoffersen (approved):
LGTM

PR #2945, opened 2024-02-22
splatfacto-big

AdamRashid96:
-Added splatfacto-big which is just lower cull threshold and no culling after 15k steps
Those are the parameters I used for the splatfacto-big in the eval table


kerrj (approved):
lgtm

PR #2946, opened 2024-02-22
Refactor splatfacto gaussian params with a ParameterDict

kerrj:
This simplifies subclassing, and simplifies some of the logic for culling/splitting.

### DIFF
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -599,7 +599,6 @@ method_configs["splatfacto"] = TrainerConfig(
     steps_per_eval_all_images=1000,
     max_num_iterations=30000,
     mixed_precision=False,
-    gradient_accumulation_steps={"camera_opt": 100},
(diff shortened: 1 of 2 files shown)


jb-ye:
The change looks good to me. Some suggestions:

(1) test PLY exporter end-to-end to safeguard the workflow is fine.
(2) since the change disallow loading checkpoint from previous versions. I would suggest we bump a major version number like (1.1.0) to distinguish from previous versions. User can still use earlier versions to view their assets.

jb-ye:
One more thing: when rebasing on top of https://github.com/nerfstudio-project/nerfstudio/pull/2945 We need to change the parameter names for splatfacto-big.

kerrj:
This should be ready to merge, fixed all the mentioned changes and also added loading logic to make it backwards-compatible with old checkpoints.

maturk:
LGTM! Can we merge this?

PR #2950, opened 2024-02-23
Fix Nerfacto + Aria

brentyi:
#2783 fixed an issue where the `fisheye_crop_radius` metadata field was being accidentally deleted. Unfortunately this also broke `nerfacto` support for Aria, since it exposed some broken fisheye sampling logic from #2617 (which was previously not being run).

Specific fixed problems:
- Sampled x/y coordinates were unnecessarily scaled by `image_height // 2` and `image_width // 2`, since the crop radius is already in pixel units.
- The circle crop typically exceeds the bounds of the image, so sampling from the circle naively still has issues. We fix this with rejection sampling.

These changes + #2783 reduce artifacts.

Before:

https://github.com/nerfstudio-project/nerfstudio/assets/6992947/7c28e9ee-fa9d-4c4c-812f-cf38fa148de7

After:

https://github.com/nerfstudio-project/nerfstudio/assets/6992947/3add28c0-8ea9-4853-a3ec-2b6e4ed07196

### DIFF
--- a/nerfstudio/data/pixel_samplers.py
+++ b/nerfstudio/data/pixel_samplers.py
@@ -178,22 +178,50 @@ class PixelSampler:
         if isinstance(mask, torch.Tensor) and not self.config.ignore_mask:
             indices = self.sample_method(batch_size, num_images, image_height, image_width, mask=mask, device=device)
         else:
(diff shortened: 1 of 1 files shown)


kerrj (approved):
cool!

Issue #2909, opened 2024-02-13
Splat .ply export not supported by multiple viewers


robinsonkwame:
**Describe the bug**

Documentation [here](https://docs.nerf.studio/nerfology/methods/splat.html#exporting-splats) states
---
Nerfstudio’s splat export currently supports multiple third-party splat viewers:

[Polycam Viewer](https://poly.cam/tools/gaussian-splatting)

[Playcanvas SuperSplat](https://playcanvas.com/super-splat)

[WebGL Viewer by antimatter15](https://antimatter15.com/splat/)

[Spline](https://spline.design/)

[Three.js Viewer by mkkellogg](https://github.com/mkkellogg/GaussianSplats3D)
---

but an [exported splat .ply](https://drive.google.com/file/d/1l-w5KoCviKfA3y5knl00vLPCKmFA2FrZ/view?usp=sharing) generated from a sample Lion statue video ([here](https://replicate.delivery/mgxm/42fc2c94-74c0-4eee-b7ed-727a9b611c24/LionStatue.MOV)) on 2/9 fails to be viewed properly in all 3rd party viewers but Polycam.

**To Reproduce**
1. Download sample Lion statue video ([here](https://replicate.delivery/mgxm/42fc2c94-74c0-4eee-b7ed-727a9b611c24/LionStatue.MOV))
2. run `ns-process`, `ns-train splatfacto`, and `ns-export`
3. Go to each viewer listed above
4. Upload the .ply file, compare results to polycam

**Expected behavior**
All viewers at least approximately look similar

**Screenshots**
e.g.,

Polycam
![image](https://github.com/nerfstudio-project/nerfstudio/assets/8014151/e982749f-a258-42bd-95e4-df9ff7010477)

Supersplat
![image](https://github.com/nerfstudio-project/nerfstudio/assets/8014151/c9142ae6-0aa8-436f-b5dd-ca8c318ed6e5)

**Additional context**
Even if Nerfstudio can't support the listed viewer, it would be helpful to describe any transformations and changes to exported splat .ply files that are typically required to be viewable in, say, Supersplat and other viewers. This would support frontend UI development against NerfStudio splat output.

jb-ye:
Your splat can by rendered in supersplat ( you need to set splat size to 0 on the left panel). You just need to rotate and room a bit to see the lion ... 

The applied transform was lost in PLY exporter, if you use nerfstudio viewer, it will restore properly and see the lion on the startup frame. 

[...]

PR #2951, opened 2024-02-23
Support splat export in original dataset coordinates

brentyi:
The `.ply` export for Gaussians previously defaulted to saving in Nerfstudio's auto-oriented / auto-scaled coordinate frame.

I added support for exporting this in the original dataset coordinate frame, which the point cloud export supports via a "Save in world frame" checkbox.

I matched this in both the GUI and the export script CLI:

image

To me it makes the most sense to default this to `True` (I also flipped this for the point cloud export), but open to thoughts!

cc @jb-ye, #2909

### DIFF
--- a/tests/utils/test_spherical_harmonics.py
+++ b/tests/utils/test_spherical_harmonics.py
@@ -73,17 +73,12 @@ def test_spherical_harmonics_rotation_splatfacto(sh_degree):
     sh_coeffs = torch.rand(N, 3, num_coeffs)
     dirs = torch.rand(N, 3)
     dirs = dirs / torch.linalg.norm(dirs, dim=-1, keepdim=True)
-
     assert dirs.shape == (N, 3)
-    y_lm = gsplat_eval_sh_bases(num_coeffs, dirs)
-    color_original = (sh_coeffs * y_lm[..., None, :]).sum(dim=-1)
 
     rot_matrix = torch.tensor(ScR.random().as_matrix(), dtype=torch.float32)
     sh_coeffs_rotated = rotate_spherical_harmonics(rot_matrix, sh_coeffs, component_convention="-y,+z,-x")
     dirs_rotated = (rot_matrix @ dirs.T).T
     assert dirs_rotated.shape == (N, 3)
-    y_lm_rotated = gsplat_eval_sh_bases(num_coeffs, dirs_rotated)
-    color_rotated = (sh_coeffs_rotated * y_lm_rotated[..., None, :]).sum(dim=-1)
 
     torch.testing.assert_close(
         gsplat_spherical_harmonics(sh_degree, coeffs=sh_coeffs.swapaxes(-1, -2), dirs=dirs),


kerrj (dismissed):
seems good

jb-ye:
I don't think the option would work actually for spherical harmonic parameters.  The spherical harmonics are saved in the original frame and cannot be "re-orient" back to the original coordinates.

brentyi:
> I don't think the option would work actually for spherical harmonic parameters. The spherical harmonics are saved in the transformed frame and cannot be "re-orient" back to the original coordinates. (We can apply translation and scale, but not rotations, otherwise SH becomes inconsistent.)
> 
> My recommendation is to save the transforms as meta data as part of exporter.

Agree the current implementation is flawed, will revisit this after ECCV deadline!

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -180,26 +180,30 @@ def process_eval_data(idx):
...
                     ),
+                    description="Caching / undistorting train images",
+                    transient=True,
+                    total=len(self.train_dataset),
                 )
             )
 
         CONSOLE.log("Caching / undistorting eval images")
-        with ThreadPoolExecutor() as executor:
+        with ThreadPoolExecutor(max_workers=2) as executor:
```
pwais, line 197: please make this configurable? or maybe this is just for debugging

nerfstudio/scripts/exporter.py
```diff
@@ -497,7 +504,26 @@ def main(self) -> None:
...
-            positions = model.means.cpu().numpy()
+            if self.save_world_frame:
+                assert isinstance(pipeline.datamanager, FullImageDatamanager)
+                dataparser_outputs = pipeline.datamanager.train_dataparser_outputs
+                dataparser_scale = dataparser_outputs.dataparser_scale
+                dataparser_transform = dataparser_outputs.dataparser_transform.numpy(force=True)
+
+                output_scale = 1 / dataparser_scale
+                output_transform = np.zeros((3, 4))
+                output_transform[:3, :3] = dataparser_transform[:3, :3].T
```
pwais, line 515: pretty please don't do transform math w/out at least comments, this sort of code is 110% likely to put a future reader in transform hell

also pretty please use ` pipeline.datamanager.train_dataparser_outputs.transform_poses_to_original_space()` because 
(1) that's what's used elsewhere in this file
(2) using that function ensures future refactors won't break things, and most past nerfstudio refactors have indeed broken lots of things

nerfstudio/scripts/exporter.py
```diff
@@ -497,7 +504,26 @@ def main(self) -> None:
...
+                output_transform[:3, 3] = -dataparser_transform[:3, :3].T @ dataparser_transform[:3, 3]
+            else:
+                output_scale = 1
+                output_transform = np.zeros((3, 4))
+                output_transform[:3, :3] = np.eye(3)
+            inv_dataparser_quat = quaternion_from_matrix(output_transform[:3, :3])
+
+            positions = (
+                np.einsum("ij,bj->bi", output_transform[:3, :3], model.means.cpu().numpy() * output_scale)
+                + output_transform[None, :3, 3]
```
pwais, line 525: please don't do this, *ESPECIALLY* w/out comments.  instead consider:

```
positions = model.means.cpu().numpy()

poses = np.eye(4, dtype=np.float32)[None, ...].repeat(positions.shape[0], axis=0)[:, :3, :]
poses[:, :3, 3] = positions
poses = pipeline.datamanager.train_dataparser_outputs.transform_poses_to_original_space(
    torch.from_numpy(poses)
)
```

nerfstudio/scripts/exporter.py
```diff
@@ -518,11 +544,27 @@ def main(self) -> None:
...
             map_to_tensors["opacity"] = model.opacities.data.cpu().numpy()
 
-            scales = model.scales.data.cpu().numpy()
+            # Note that scales are in log space!
+            scales = model.scales.data.cpu().numpy() + np.log(output_scale)
             for i in range(3):
                 map_to_tensors[f"scale_{i}"] = scales[:, i, None]
 
-            quats = model.quats.data.cpu().numpy()
+            def quaternion_multiply(wxyz0: np.ndarray, wxyz1: np.ndarray) -> np.ndarray:
```
pwais, line 552: Please no.  First of all, `scipy.spatial.transform` already has quaternion multiply... at least this code is clear about scalar-first versus scalar-last.

this could be made more concise, but consider instead:
```
from scipy.spatial.transform import Rotation as ScR
# ns gplat says quaternions are [w,x,y,z] scalar-first format
# scipy is [x, y, z, w] scalar-last format

raw_quats = model.quats.data.cpu().numpy().squeeze()
R_quats = ScR.from_quat(raw_quats[:, [1, 2, 3, 0]])

# apply the inverse dataparser transform to the splat rotations
poses = np.eye(4, dtype=np.float32)[None, ...].repeat(raw_quats.shape[0], axis=0)[:, :3, :]
poses[:, :3, :3] = R_quats.as_matrix()
poses = pipeline.datamanager.train_dataparser_outputs.transform_poses_to_original_space(
    torch.from_numpy(poses)
)
rots_in_input = poses[:, :3, :3].numpy()
quat_in_input = ScR.from_matrix(rots_in_input)
quats = quat_in_input.as_quat()[:, [3, 0, 1, 2], None]
```


Again, this uses `transform_poses_to_original_space()`, which might amalgamate several different transforms and scales, who knows? and hopefully `transform_poses_to_original_space()` gets maintained.  But its really really important to be clear about frames etc

pwais (commented):
@jb-ye Many viewers ignore spherical harmonics, and in my experience doing a rigid transform on splats looks plenty fine.  I think ideally Nerfstudio just does a scale and re-center _internally_ *as part of the model* rather than changing any of the camera poses, initial points etc that are part of the dataset.  It should be up to the model to normalize as needed, not up to the user.

brentyi:
> I think ideally Nerfstudio just does a scale and re-center _internally_ *as part of the model* rather than changing any of the camera poses, initial points etc that are part of the dataset.

Tough to overcome the inertia here but I agree that this would solve a lot of problems!

nerfstudio/data/datamanagers/full_images_datamanager.py
brentyi: I'm actually not sure why this shows up in this diff, the change is from https://github.com/nerfstudio-project/nerfstudio/pull/2969. it speeds up undistortion a lot for big datasets I've been toying with!

it's hardcoded to 2 because we can really only expect benefits from one thread doing IO while the other thread is doing undistortion; the implementation is still weird given this (ideally we'd just have 1 worker doing IO while the main one is sequentially undistorting) but I'm a fan of not letting perfect be the enemy of... better 🤷

pwais: FWIW each worker might need `cv2.setNumThreads(1)` or else it can choke the CPU.  I seem to see way more than 200% util here hence why i brought it up so maybe it's just not tuned well for all users.

maybe in a future refactor this stuff will just get pushed to a torch dataloader... the pinned memory part breaks for me for large datasets anyways, literally I got a OOM and hard lock-up because too too too much much much pinned memory

pwais:
> Tough to overcome the inertia here but I agree that this would solve a lot of problems!

Yes inertia but at least the PR discussions leave breadcrumbs.  Things that fall off the rolling katamari ball become seeds for the next re-write.

jkulhanek:
Wait, spherical harmonics can easily be rotated by using Wigner matrices, right?

jb-ye:
@jkulhanek Could you share a gist of sample code of rotating spherical harmonics? I am not aware of Wigner matrices.

jkulhanek:
Just a note here. Since SH is a complete basis, the rotation is possible exactly. Here is the code:
https://github.com/jkulhanek/nerfbaselines/blob/develop/nerfbaselines/math_utils.py

I played it in a notebook, and it seems to rotate the SH correctly.

Frenchman997:
Is there any update on this? I would also be interested in the ability to export splats in their original world frame. I haven't looked into it yet at all but would be willing to contribute.

hardikdava:
@jkulhanek can you provide the notebook? Thanks

nerfstudio/scripts/exporter.py
brentyi: (note for when we revive this PR, which is planned) for the quaternion multiply if we don't want to deal with the xyzw/wxyz conversion of scipy we can also use `(vtf.SO3(wxyz0) @ vtf.SO3(wxyz1)).wxyz` with `import viser.transforms as vtf` where `viser>=0.1.30`

pwais: voicing a preference for the use of standard scipy / numpy / torch wherever possible

yes it's unfortunate that there are different quaternion encodings, different camera conventions, different euler angle conventions ....

hongsukchoi:
Spherical harmonics functions are equivariant to SO(3) and can be rotated.

I don't know the exact code to do it, but probably possible with `e3nn` api:

From ChatGPT
```Python
import torch
from e3nn.o3 import Irrep, Irreps
from e3nn.o3 import spherical_harmonics, Rotation

R = torch.tensor([ [0.36, 0.48, -0.8], [-0.8, 0.60, 0], [0.48, 0.64, 0.6] ])

l = 2  # Degree of spherical harmonics

directions = torch.tensor([[0.0, 0.0, 1.0]]) # z-axis unit vector 

Y_lm = spherical_harmonics(l, directions)

# Get the irreducible representation (Irrep) of the spherical harmonics of degree 
l irrep = Irrep(f"{l}e") # e denotes even parity 

# Apply the rotation to the spherical harmonics 
Y_lm_rotated = irrep.D_from_matrix(R) @ Y_lm
```

Explanation:
* spherical_harmonics(l, directions): Computes the spherical harmonics of degree l for the given directions.
* Irrep(f"{l}e"): Creates an irreducible representation for the spherical harmonics of degree l. The "e" stands for even parity, which is typical for standard spherical harmonics.
* D_from_matrix(R): Computes the Wigner D-matrix for the rotation matrix R, which describes how the spherical harmonics transform under rotation.
* @: Performs matrix multiplication between the Wigner D-matrix and the original spherical harmonics to obtain the rotated spherical harmonics.

Ben-Mack:
Is Playcanvas related code for SH rotation a good fit to unblock this PR: https://github.com/playcanvas/engine/blob/release-1.69/src/scene/gsplat/gsplat.js#L259

How can we get the matrix that represent the output coordinate? (Or the diff between original and output coordinate)

pwais:
The line you cite seems to just load splats?  rather than rotate the SHs given an arbitrary transform.

> Also, how can we get the matrix that represent the output coordinate? (Or the diff between original and output coordinate)

In nerfstudio there's a `dataparser_transform` that gets stored in the dataparser outputs / model metadata files.  AFAIK there's no "splat file format" that normally contains this transform.  The code in this change uses that transform to put the trained splats back into the training frame (right now the PR just has translations and rotations tho I believe, yeah).

jkulhanek:
You can use this code https://gist.github.com/jkulhanek/aae3ec12d779ffc729c72157315df0da

Ben-Mack:
> The line you cite seems to just load splats? rather than rotate the SHs given an arbitrary transform.

Sorry, this is the correct link, which they mentioned `Rotate spherical harmonics up to band 3 based on https://github.com/andrewwillmott/sh-lib`:
https://github.com/playcanvas/supersplat/blob/b008eca2bd28d2aae7d2defd67363ae6c4f7174b/src/sh-utils.ts#L42

pwais:
on first blush it's hard to compare the cited PlayCanvas impl vs @jkulhanek 's link (PlayCanvas looks possibly like an algebraic simplification somehow?).  

But it would be nice for any final solution to have some unit tests.  For example, if dataparser_transform breaks somehow, want to easily determine it's that instead of something in complex sh rotations.

PR #2954, opened 2024-02-24
Fix fisheye sampling to valid coordinates

sweeneychris:
I guess during a bug fix the exported "fisheye crop radius" changed from a 0 to 1 value to a pixel value, and so we no longer need to multiply by half the image size. Using a clamp here to ensure no values go out of bounds as well


brentyi:
I also made this fix last night: #2954. Do you have a preference on which to merge?

sweeneychris:
You can go for it @brentyi !

PR #2966, opened 2024-02-28
fix bug that opacity compentation not being applied to depth raterization

jinhwanlazy:
fix bug that opacity compentation has not being applied to depth raterization, found from @jb-ye 's new rasterization mode implementation pr

https://github.com/nerfstudio-project/nerfstudio/pull/2888#issuecomment-1961824489


jb-ye:
Thanks!

kerrj (approved):
sgtm!

Issue #2061, opened 2023-06-12
how add multiple cameras in dataparser and make use of them in datamanager?


Tejas-Deo:
I want to use the "CameraType.FISHEYE" and "CameraType.PERSPECTIVE" in dataparser as I have datasets for both these model types. How should I include both camera model types in a single dataparser file? Or should I create different dataparsers?

I want to train a single NeRF model on both datasets. Can you please guide me about the same?

tancik:
Unfortunately at this time the repo doesn't support this functionality. The way we currently handle distortions assumes all of the cameras are the same type so that it can be done in batch. This would need to change to support multiple camera types.

Tejas-Deo:
Is there a specific way that you would suggest to include this functionality as well?

PR #2976, opened 2024-03-01
Support multiple-camera COLMAP reconstructions

zacstewart:
This PR is an attempt to support using COLMAP reconstructions that were built from multiple cameras. The transforms.json produced by other formats adds per-image camera parameters directly to each `frame` object. I naively tried to do the same thing when processing COLMAP data, but admittedly I do not know if this is valid.

#2061


jb-ye (commented):
@zacstewart The unit test is failing, could you take a look?

jb-ye:
Close this as the dup of merged PR https://github.com/nerfstudio-project/nerfstudio/pull/3081

PR #2978, opened 2024-03-01
doc: fix code typo in 'NeRF' method doc

Crescent-Saturn:
In NeRF methods doc, old code component was found. So just fix this small typo issue.

Current one:
```python
from nerfstudio.renderers.renderers import RGBRenderer
```

Correct one:
```python
from nerfstudio.model_components.renderers import RGBRenderer
```

Thanks for reviewing!


brentyi (approved):
lgtm, thanks!!

PR #2982, opened 2024-03-03
[WIP] Depth nerfacto with visibility loss

ethanweber:


nerfstudio/fields/visibility_field.py
```diff
@@ -0,0 +1,90 @@
...
+        self.c2ws = cameras.camera_to_worlds
+        self.c2whs = torch.cat([self.c2ws, torch.zeros_like(self.c2ws[:, :1, :])], dim=1)
+        self.c2whs[:, 3, 3] = 1.0
+        self.w2chs = torch.inverse(self.c2whs)
+        self.K = cameras.get_intrinsics_matrices()
+        self.image_height = cameras.height
+        self.image_width = cameras.width
+
+    @torch.no_grad()
+    def forward(self, ray_samples: RaySamples, camera_chunk_size=50, ray_chunk_size=4096) -> torch.Tensor:
```
kerrj, line 40: does it maybe make sense to take in positions (Nx3) directly instead of ray_samples? this would make it compatible with other methods which might want to use visibility like splatting.

PR #2987, opened 2024-03-07
Fix: Masks having wrong dimension after undistortion (causes crash in splatfacto)

Zunhammer:
I noticed that masks in splatfacto are not working after undistortion. I get an error when masks are applied in the full images datamanager in line 881 (gt_img = gt_img * mask) as the shapes are (w, h, 3) and (w, h), however they need to be (w, h, 3) and (w, h, 1).
image

The resulting Error is: RuntimeError: The size of tensor a (3) must match the size of tensor b (3839) at non-singleton dimension 2
terminate called without an active exception

Please note that this issue is not present in nerfacto. Also it does not appear when using undistorted images (k- and p-params all 0)

I fixed it by adding the third dimension after undistortion. Tested both splatfacto and nerfacto as well, seems to be fixed to me.

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -27,20 +27,7 @@ from copy import deepcopy
 from dataclasses import dataclass, field
 from functools import cached_property
 from pathlib import Path
-from typing import (
-    Dict,
-    ForwardRef,
-    Generic,
-    List,
-    Literal,
-    Optional,
-    Tuple,
-    Type,
-    Union,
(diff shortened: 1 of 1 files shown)


maturk (approved):
This looks reasonable, the masks should indeed be (h,w,1) in nerfstudio. I am guessing this peculiarity comes from the cv2 side again... since it should be ensured in the dataparser side [here](https://github.com/nerfstudio-project/nerfstudio/blob/6e6cedd438348d03f390ba0d9b8afddaea4163c1/nerfstudio/data/datasets/base_dataset.py#L129)

PR #2991, opened 2024-03-08
Add 'within' function to SceneBox

tobiasfshr:
This PR adds a `within` function to the `SceneBox` class, analogous to the `within` function in the `OrientedBox` class.


maturk:
This looks reasonable, I guess the convention is that the maximum and minimum bounds are not included, but that is what the OrientedBox does as well. I suggest removing the test though, I think it is not necessary here.

kerrj (approved):
lgtm, thanks!

PR #3006, opened 2024-03-18
Support COLMAP reconstructions with non-flat image dirs

zacstewart:
The code was previously looking for the file.name, which drops all directory information from the path. Instead, we want to get the file path relative to the data dir, which will only drop parts of the path above the data dir, allowing us to recurse into subdirectories within the images dir

For example, you can use COLMAP reconstructions that were built using multiple captures and have an images directory structured like this:

```
images
└── DJI_0238
    ├── 000000.jpg
    ├── 000041.jpg
    ├── 000060.jpg
└── DJI_0286
    ├── 000000.jpg
    ├── 000041.jpg
    ├── 000060.jpg
```

### DIFF
--- a/nerfstudio/process_data/images_to_nerfstudio_dataset.py
+++ b/nerfstudio/process_data/images_to_nerfstudio_dataset.py
@@ -88,8 +88,8 @@ class ImagesToNerfstudioDataset(ColmapConverterToNerfstudioDataset):
                 image_rename_map_paths.update(eval_image_rename_map_paths)
 
             image_rename_map = dict(
-              (a.relative_to(self.data).as_posix(), b.name)
-              for a, b in image_rename_map_paths.items())
+                (a.relative_to(self.data).as_posix(), b.name) for a, b in image_rename_map_paths.items()
+            )
             num_frames = len(image_rename_map)
             summary_log.append(f"Starting with {num_frames} images")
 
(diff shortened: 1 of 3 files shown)


jb-ye:
@zacstewart Could you describe how you test this PR in step-by-step commands? My understanding is the change should support both single folder and multiple subfolder colmap project.

zacstewart:
@jb-ye Given you have a colmap reconstruction that has subdirectories in its images directory, there are no more steps than usual. This PR only effects how `ns-process-data images` will behave by allowing it to recurse into those subdirectories instead of only taking top-level image files. The tedious part of testing would be getting an applicable colmap reconstruction.

Do you need me to help prepare a small test dataset to demonstrate it? I can do so but I'll need a week before I can get to it.

#2976 is an orthogonal issue. Colmap allows you to build reconstructions from homogeneous camera images, but `ns-process-data` treats all images as if they are from a single camera and makes each image share intrinsics in the transforms.json it produces. It's not strictly related to this PR (#3006), but typically images from different cameras are in different subdirectories, so in practice they are both necessary if you want to train a nerf from a colmap reconstruction built using heterogeneous images.

In my case, I already have a colmap reconstruction that I built myself, outside of nerfstudio. It required special care to produce because the scene is very symmetrical and images tended to mis-register. Nerfstudio's internal invocation of colmap would not be able to reproduce it.

jb-ye (approved):
Thanks, this is a low risk change and bring clear benefits.

PR #3010, opened 2024-03-20
Fix SH

jkulhanek:
I think there is a bug in the SH pytorch impl. This PR should fix it. Please take a look to see if it's correct. I took the correct implementation from ingp.

### DIFF
--- a/tests/model_components/test_renderers.py
+++ b/tests/model_components/test_renderers.py
@@ -31,7 +31,8 @@ def test_sh_renderer():
     levels = 4
     num_samples = 10
 
-    sh = torch.ones((3, num_samples, 3 * levels**2))
+    sh = torch.zeros((3, num_samples, 3 * levels**2))
+    sh[..., 0] = 3.6
     weights = torch.ones((3, num_samples, 1))
     weights /= torch.sum(weights, dim=-2, keepdim=True)
     directions = torch.zeros((3, num_samples, 3))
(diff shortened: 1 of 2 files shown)


jb-ye:
Just to cite Inria gaussian splatting's implenentation: https://github.com/graphdeco-inria/gaussian-splatting/blob/main/utils/sh_utils.py

jkulhanek:
Inria’s impl us also correct. The ingp’s one, however, supports more degrees.

tests/model_components/test_renderers.py
```diff
@@ -31,7 +31,8 @@ def test_sh_renderer():
     levels = 4
     num_samples = 10
 
-    sh = torch.ones((3, num_samples, 3 * levels**2))
+    sh = torch.zeros((3, num_samples, 3 * levels**2))
+    sh[..., 0] = 3.6
```
kerrj, line 35: what is this for? maybe add a comment

jb-ye:
Just want to revisit this topic: Shall we remove SH evaluation in nerfstudio, and use the one from gsplat library? So it minimize our maintenance effort. @kerrj

aayushg55:
I don't think the existing SH encoding is necessarily wrong - it just follows a different convention for the basis function. Most other libraries appear to use the Condon-Shortely phase factor ((-1)^m) convention, which results in the sign flipping for odd-numbered coefficients in each band.

brentyi:
In #2951 the convention difference is adding some complexity for SH rotation with Wigner-D.

For removing SH evaluation in nerfstudio: one difference is that gsplat only supports NVIDIA GPUs in the public API. Nerfacto (should be) runnable on CPU, MPS, etc.

PR #3011, opened 2024-03-20
Fix rotation_matrix for opposite float32 vectors

jb-ye:
When tensors a and b are float32, the condition c < -1 + 1e-8 will never be triggered because of round off tolerance being too small. Also I use a more principled implementation for computing the rotation matrix when a, b are opposite. 

Test cases:

```
a=torch.rand(3, dtype=torch.float32)
b=-a
error=torch.norm(rotation_matrix(a, b) @ a[..., None] - b[..., None])
```


brentyi (approved):
I made a few minor changes:
- Reformulating to return "exact" solutions without `eps` in the denominator
- Renaming the function

If these makes sense to you @jb-ye the PR looks good to me 🙂

jb-ye:
@brentyi It looks good to me. Thanks for improving my PR.

PR #3016, opened 2024-03-23
Fix bug when trying to create a camera path in viewer

MildMystery:
This PR fixes the issue that I see when trying to add multiple keyframes in the viewer to generate a camera path for rendering. I get the following error:

```
  File "repos/nerfstudio/nerfstudio/viewer/render_panel.py", line 269, in spline_t_from_t_sec                                                                                                             interpolator = scipy.interpolate.PchipInterpolator(                                                                                                                                                            
AttributeError: module 'scipy' has no attribute 'interpolate'  
```
FYI, I don't see this error when trying to create a camera path running the legacy viewer i.e. when I try to create a path after running `ns-viewer --load-config outputs/test_data/nerfacto/2024-03-23_151802/config.yml  --vis viewer-legacy` instead of `ns-viewer --load-config outputs/test_data/nerfacto/2024-03-23_151802/config.yml`

### DIFF
--- a/nerfstudio/viewer/render_panel.py
+++ b/nerfstudio/viewer/render_panel.py
@@ -24,7 +24,7 @@ from pathlib import Path
 from typing import Dict, List, Literal, Optional, Tuple, Union
 
 import numpy as np
-import scipy
+from scipy import interpolate
 import splines
 import splines.quaternion
 import viser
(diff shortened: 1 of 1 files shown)


kerrj (approved):
Seems reasonable, I wonder if it's dependent on the version of scipy?

PR #3031, opened 2024-03-30
Added All pixel Sampler

IamMohitM:
A sampler that selects a specific image and samples all pixels from that image.

I believe this could be a useful class where the neural field is optimised per image. This is used in [nerf2mesh](https://github.com/ashawkey/nerf2mesh), where the pixels must be sampled from a single image to generate rays, which are then used to rasterise the image from the coarse mesh.

I [use](https://github.com/IamMohitM/nerf2mesh-nerfstudio/blob/12d43c429449b23441d1e1ef64de8b2d3aba0810/nerf2mesh/datamanager.py#L29) this pixel sampler in my [implementation of nerf2mesh](https://github.com/IamMohitM/nerf2mesh-nerfstudio/blob/12d43c429449b23441d1e1ef64de8b2d3aba0810/nerf2mesh/sampler.py#L13) using nerfstudio.

Happy to make any edits if required.

### DIFF
--- a/nerfstudio/data/pixel_samplers.py
+++ b/nerfstudio/data/pixel_samplers.py
@@ -531,3 +531,35 @@ class PairPixelSampler(PixelSampler):  # pylint: disable=too-few-public-methods
         pair_indices += indices
         indices = torch.hstack((indices, pair_indices)).view(rays_to_sample * 2, 3)
         return indices
+
+
+@dataclass
+class AllPixelSamplerConfig(PixelSamplerConfig):
+    _target : Type = field(default_factory=lambda: AllPixelSampler)
+
+class AllPixelSampler(PixelSampler):
+    config: AllPixelSamplerConfig
+
+    def sample(self, image_batch: Dict):
+        """
+        Samples all pixels from a single image. The image is selected randomly from the batch.
+        """
+        device = image_batch["image"][0].device
+        num_images, image_height, image_width, _ = image_batch["image"].shape
+
+        image_index = torch.randint(num_images, (1,), device=device)
+        i, j= torch.meshgrid(   
+            torch.arange(image_height, device=device),
+            torch.arange(image_width, device=device),
+            indexing='ij'
+        )
+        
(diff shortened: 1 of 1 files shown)


jb-ye:
Just some side question: I am also curious why we didn't have something like AllPixelSampler for implementing [next_eval_image](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/datamanagers/parallel_datamanager.py#L305).

I spent sometime digging into the code and realized that we currently use [RandIndicesEvalDataloader](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/utils/dataloaders.py#L251) to select a random camera and use [get_outputs_for_camera](https://github.com/nerfstudio-project/nerfstudio/blob/59d6acf9da718d1d513d260b8808a9bed1bbbd39/nerfstudio/models/base_model.py#L166) which involves a nested call of camera.[generate_rays](https://github.com/nerfstudio-project/nerfstudio/blob/59d6acf9da718d1d513d260b8808a9bed1bbbd39/nerfstudio/cameras/cameras.py#L320). 

The process to obtain rays is not following the typical workflow we set for next_train() and next_eval(), where a pixel sampler is used.

IamMohitM:
That makes sense. I didn't notice the RandIndicesEvalDataloader earler.

The AllPixelSampler implementation makes it compatible with the[ next_train()](https://github.com/nerfstudio-project/nerfstudio/blob/59d6acf9da718d1d513d260b8808a9bed1bbbd39/nerfstudio/data/datamanagers/base_datamanager.py#L530) and [next_eval()](https://github.com/nerfstudio-project/nerfstudio/blob/59d6acf9da718d1d513d260b8808a9bed1bbbd39/nerfstudio/data/datamanagers/base_datamanager.py#L541) function in the [VanillaDataManager](https://github.com/nerfstudio-project/nerfstudio/blob/59d6acf9da718d1d513d260b8808a9bed1bbbd39/nerfstudio/data/datamanagers/base_datamanager.py#L359) class. Both function assert a requirement for a pixel sampler.

PR #3034, opened 2024-04-02
Fix core tests with Pillow>=10.3.0

jb-ye:
The latest version update of Pillow will break nerfstudio since it is using a deprecated API field.


kerrj (approved):
lgtm!

PR #3043, opened 2024-04-04
Only load+undistort splatfacto images when they're used

### LINKED ISSUE #2817: Make viewer's loading time faster by skipping loading all images
Hello!
First of all, thanks for the v1.0.0 update! Looks amazing and promising very much.

I have a question / request: is it somehow possible to make model's loading time in the viewer faster?

It seems that the viewer loads all training images before rendering, as well as the checkpoint itself.
I assume that a certain part of the viewer's code requires the number of training images (maybe the appearance embedding?), which can be exhaustive and time-consuming when the dataset is very large.
If my assumption is true, can't we store the number of images as an attribute after loading them for training?
If not, what is the reason for loading the images and how can i manually override this?

Thank you very much in advance!

### LINKED ISSUE #2971: Out of memory using ns-render with splatfacto for large datasets
**Describe the bug**
I'm running the splatfacto method with Project Aria data using 1408x1408 images as input from a long video sequence. This trains fine using out-of-core sampling of the dataset and I can visualize the results in the Viser viewer perfectly. However, when I try to use ns-render to render out the dataset cameras it runs out of memory.  cc @brentyi 

**To Reproduce**
1. Train splatfactor with a large dataset (1000s of high res images)
2. Call ns-render dataset to render the dataset images with the trained model
3. CPU memory should explode and the process will be killed or your machine will freeze

**Expected behavior**
Here's what I think might be happening -- the ns-render seems to use the FullImagesDatamanager which attempts to cache *all* undistorted images in memory. This unnecessarily causes memory to explode during the rendering process.

A better solution would be to only load a fixed # of images at a time before rendering out, or otherwise manage the cache in FullImagesDataManager to proactively read batches of images into memory with a fixed memory budget.

**Additional context**
I'm happy to try and work on this, I just probably need a few pointers on the path of least resistance since there are probably multiple ways to do this (some more intrusive than others)

brentyi:
Some changes to only undistort splatfacto train/eval images when they're used. This should speed up:
- `ns-export`
- `ns-render`: fixes #2971
- `ns-viewer` with `--viewer.max-num-display-images 0`: fixes #2817

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -32,9 +32,6 @@ from typing import Dict, ForwardRef, Generic, List, Literal, Optional, Tuple, Ty
 import cv2
 import numpy as np
 import torch
-from rich.progress import track
-from torch.nn import Parameter
-
 from nerfstudio.cameras.camera_utils import fisheye624_project, fisheye624_unproject_helper
 from nerfstudio.cameras.cameras import Cameras, CameraType
 from nerfstudio.configs.dataparser_configs import AnnotatedDataParserUnion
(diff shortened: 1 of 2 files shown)


kerrj (approved):
nice!

Issue #2959, opened 2024-02-26
Cannot download poster dataset using ns-download-data


abrbhat:
**Describe the bug**
Getting an error while downloading poster dataset using ns-download-data

**To Reproduce**
1. Run ns-download-data nerfstudio --capture-name=poster

**Expected behavior**
Dataset should be downloaded

**Screenshots**
Error log:
```
(nerfstudio) PS C:\Users\Divya\nerfstudio> ns-download-data nerfstudio --capture-name=poster
Traceback (most recent call last):
  File "C:\Users\Divya\.conda\envs\nerfstudio\lib\site-packages\gdown\download.py", line 253, in download
    url = get_url_from_gdrive_confirmation(res.text)
  File "C:\Users\Divya\.conda\envs\nerfstudio\lib\site-packages\gdown\download.py", line 54, in get_url_from_gdrive_confirmation
    raise FileURLRetrievalError(
gdown.exceptions.FileURLRetrievalError: Cannot retrieve the public link of the file. You may need to change the permission to 'Anyone with the link', or have had many accesses. Check FAQ in https://github.com/wkentaro/gdown?tab=readme-ov-file#faq.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Divya\.conda\envs\nerfstudio\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\Divya\.conda\envs\nerfstudio\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\Divya\.conda\envs\nerfstudio\Scripts\ns-download-data.exe\__main__.py", line 7, in <module>
  File "C:\Users\Divya\.conda\envs\nerfstudio\lib\site-packages\nerfstudio\scripts\downloads\download_data.py", line 573, in entrypoint
    main(tyro.cli(Commands))
  File "C:\Users\Divya\.conda\envs\nerfstudio\lib\site-packages\nerfstudio\scripts\downloads\download_data.py", line 567, in main
    dataset.download(dataset.save_dir)
  File "C:\Users\Divya\.conda\envs\nerfstudio\lib\site-packages\nerfstudio\scripts\downloads\download_data.py", line 207, in download
    download_capture_name(save_dir, "nerfstudio", self.capture_name, capture_name_to_file_id=nerfstudio_file_ids)
[...]

Issue #3044, opened 2024-04-05
Cannot access GDrive link for test data


PAD2003:
**Describe the bug**
The link for test data is 404 not found

**To Reproduce**
When using the command "ns-download-data nerfstudio --capture-name=poster", I face error because the link is broken. Accessing through browser, I face 404 not found problem.

**Expected behavior**
Can you update the link for test data, please!

PR #3045, opened 2024-04-05
New Google Drive link for poster dataset

ginazhouhuiwu:
Hosting new dataset [here](https://drive.google.com/file/d/1FceQ5DX7bbTbHeL26t0x6ku56cwsRs6t/view?usp=sharing).

Addresses #3044


brentyi:
also fixes #2959!

brentyi (approved):

edwardchenomni:
Hi, thanks for your works!
But I found that the zip file you provided may cause another problem.
in your zip file, there exist not only the expected 'poster' folder but also a '__MACOSX' folder.
This triggers the assertion error at [this line](https://github.com/nerfstudio-project/nerfstudio/blob/babf57718c75e7d04f1b977f035db67c02a9c20a/nerfstudio/scripts/downloads/download_data.py#L174).
Moreover, depends on the runtime, the two folders may switch order, which in turn cause problem on [the next line](https://github.com/nerfstudio-project/nerfstudio/blob/babf57718c75e7d04f1b977f035db67c02a9c20a/nerfstudio/scripts/downloads/download_data.py#L175).

PR #3046, opened 2024-04-05
Changes for trainer.py to support the Gradio webui

KevinXu02:
Changes including step tracking. An early stop option for the webui (modifying the training_status won't pass the tests as it also requires changes in Viser, so I made it a flag). And stops the viser after training when quit_on_train_completion is on.


nerfstudio/engine/trainer.py
```diff
@@ -137,6 +137,12 @@ def __init__(self, config: TrainerConfig, local_rank: int = 0, world_size: int =
 
         self.viewer_state = None
 
+        # used to keep track of the current step
+        self.step = 0
+        # A flag to stop training early for gradio webui, added seperately to avoid
+        # interference with viewer
+        self.early_stop = False
```
brentyi, line 144: Can we add a method like:
```
def shutdown(self) -> None:
    """Stop down the trainer and stop all associated threads/processes (such as the viewer)."""
```
that sets this to true? Usage-wise a method seems more self-documenting than an attribute that just needs to be overwritten

nerfstudio/engine/trainer.py
```diff
@@ -137,6 +137,12 @@ def __init__(self, config: TrainerConfig, local_rank: int = 0, world_size: int =
 
         self.viewer_state = None
 
+        # used to keep track of the current step
+        self.step = 0
+        # A flag to stop training early for gradio webui, added seperately to avoid
+        # interference with viewer
```
brentyi, line 143: can we reword "interference with viewer"? it feels a little vague to me

nerfstudio/engine/trainer.py
```diff
@@ -291,12 +301,22 @@ def train(self) -> None:
...
-        # save checkpoint at the end of training
-        self.save_checkpoint(step)
+        # save checkpoint at the end of training, and write out any remaining events
+        self._after_train()
 
+    def shutdown(self) -> None:
+        """Stop the trainer and stop all associated threads/processes (such as the viewer)."""
+        self.stop_training = True  # tell the training loop to stop
+        if self.viewer_state is not None:
+            self._stop_viewer_server()
```
brentyi, line 311: perhaps we could get rid of `_stop_viewer_server()` and just write `self.viewer_state.viser_server.stop()` here?

brentyi (approved):

PR #3060, opened 2024-04-09
Fixed Unexpectedly huge RAM consumption of parallel datamanager

hyperzy:
Looks like caching images during the process construction consumes more memory than expected (at least 2x). The memory footprint looks fine when it's moved after the process starts. Can anyone explain this?


kerrj:
In my testing I see no difference, however this does make sense since the fork() happens in python when start() is called, which means there are 2 processes with duplicated memory. How much of a difference does this make for you and what dataset are you testing?

hyperzy:
I used a customized dataset that contains high resolution images. The peak memory usage without the proposed modification would exceeds my PC's upper bound as shown in the first pic, and the training process was OOM killed. On the other hand, the training goes okay with the modification
Selection_003
Selection_004

kerrj (approved):
lgtm, thanks!

Issue #2121, opened 2023-06-23
Export of generfacto trained models raises AssertionError


Zunhammer:
**Describe the bug**
Training of generfacto works fine using the stablediffusion model and looks visually promising in the viewer. After training is finished,  the geometry exports are not working and raise an AssertionError (tried poisson and pointcloud). Checkpoint seems to be loaded fine though.

**To Reproduce**
1. Train a generafto model (using stabeldiffusion): ```ns-train generfacto --prompt "a high quality photo of a pineapple"```
2. In the viewer click on export -> mesh -> setup your desired settings and copy the command.
3. Paste the command in the terminal to export: ```ns-export poisson --load-config outputs/generfacto/2023-06-23_075243/config.yml --output-dir exports/mesh/ --target-num-faces 500000 --num-pixels-per-side 2048 --normal-method open3d --normal-output-name normals --num-points 1000000 --remove-outliers True --use-bounding-box True --bounding-box-min -1 -1 -1 --bounding-box-max 1 1 1```
I also tried model_output for normals.
4. AssertionError is raised

**Expected behavior**
Successful export of the mesh or pointcloud of the model.

**Screenshots**
![image](https://github.com/nerfstudio-project/nerfstudio/assets/18286310/59a79374-119b-496a-93a3-49d5bd933720)

Using nerfstudio version 0.3.2.

Zunhammer:
Just checked, it is expecting a VanillaDataManagerConfig object but the actual type is nerfstudio.data.datamanagers.random_cameras_datamanager.RandomCamerasDataManager. Anyone knows why this happens?

terrancewang:
> Just checked, it is expecting a VanillaDataManager object but the actual type is nerfstudio.data.datamanagers.random_cameras_datamanager.RandomCamerasDataManager. Anyone knows why this happens?

looking into it rn, will submit a pr to fix later tonight!

Zunhammer:
Thanks, I found that the VanillaDataManager is actually a RandomCamerasDataManager, so importing this one an replacing it in line 288 fixes this problem. However, the RandomCamerasDataManager doesn't have a train_pixel_sampler. I commented out lines 289 and 290 and was able to export a point cloud and mesh as seen below with some weird effects/errors.
![image](https://github.com/nerfstudio-project/nerfstudio/assets/18286310/e59145e5-1899-47b0-8d20-98b86b99c53c)

terrancewang:
[...]

PR #3065, opened 2024-04-10
fix exporter.py for FullImageDataManager and RandomCamerasDataManager

dberga:
Just added the imports and assertions for solving the Generfacto case. 

This is fixing this issue:
https://github.com/nerfstudio-project/nerfstudio/issues/2121


brentyi (approved):
Thanks @dberga!!

PR #3067, opened 2024-04-11
Adding lock for Viser Element callbacks

Satvik1701:
I added a wrapper around the viser element callback so the render thread is blocked while the callback is called to avoid crashing. I also added a lock in run_viewer.py so this issue is addressed for both ns-train and ns-viewer. 

Test: I added time.sleep() in the callback and confirmed the render thread was blocked.

### DIFF
--- a/nerfstudio/scripts/viewer/run_viewer.py
+++ b/nerfstudio/scripts/viewer/run_viewer.py
@@ -16,6 +16,7 @@
 """
 Starts viewer in eval mode.
 """
+
 from __future__ import annotations
 
 import time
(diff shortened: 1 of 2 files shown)


jb-ye:
What kinds of crash did you see? I also experienced crashes in the viser viewer.

kerrj:
This PR fixes a bug with a race condition in custom viewer element callbacks (eg a ViewerButton which renders the current view), previously this might clash with the standard rendering thread and crash the callback or main renderer.

kerrj (approved):
lgtm!

jb-ye:
@kerrj I found this PR impacts the experience of ns-viewer when training gaussian splatting. The delay when I rotate the scene is so long that basically disallow me to interact with the scene.

kerrj:
That's odd I don't notice any difference. In fact it shouldn't affect things since the only code that was added to the lock was viser element callbacks, which splatfacto doesn't have any of.

Issue #3048, opened 2024-04-05
when using OpenCV's undistortion code, principal point should be adjusted


f-dy:
**Describe the bug**

[This undistortion code from nerfstudio (function `_undistort_image`)](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/datamanagers/full_images_datamanager.py#L351) overlooks the fact that COLMAP/nerfstudio's cameras have the center of the topleft pixel at (0.5,0.5), but in OpenCV it's (0,0)

```
            newK, roi = cv2.getOptimalNewCameraMatrix(K, distortion_params, (image.shape[1], image.shape[0]), 0)
            image = cv2.undistort(image, K, distortion_params, None, newK)  # type: ignore
...
            if np.any(distortion_params):
                mask = cv2.undistort(mask, K, distortion_params, None, newK)  # type: ignore
```
The effect of fixing this is probably *very minimal*, and will be hidden by experimental noise (my guess, I didn't check)

The fix should be:
- subtract 0.5 from the principal point coordinates in K before calling these
- add 0.5 to the principal point coordinates in newK after calling  these (e.g., just before the line `K = newK`)
- properly document in nerfstudio that the camera model uses (0.5,0.5) as the center of the topleft pixel

I opened this issue to check if there is a consensus about the issue and the proposed fix

Refs:
- https://github.com/opencv/opencv/blob/4.9.0/modules/calib3d/src/calibration.cpp#L1293
- https://answers.opencv.org/question/208969/theory-behind-center-of-top-left-pixel-considered-as-0-0/
- https://answers.opencv.org/question/87923/sub-pixel-coordinate-origin/

Somewhat related: https://github.com/nerfstudio-project/gsplat/pull/151

PR #3071, opened 2024-04-11
shift principle points by .5 in undistortion

kerrj:
addresses https://github.com/nerfstudio-project/nerfstudio/issues/3048


kerrj:
@f-dy @jb-ye If you could look at this that'd be great, I think this implements the convention difference between OpenCV/nerfstudio correctly, but I'm not sure

jb-ye:
@kerrj Let me test this change a bit and come back to you.

jb-ye (approved):
I spent some time testing the fix, and here is my findings. For images with minor distortions, the change has close to no impacts on the final rendering results. 

Why? because effect of missing (K[0, 2] - 0.5) in cv2.undistort actually cancel out the effect of missing (new[0, 2] + 0.5) in output! The effect of change might only be more visible for very distorted images, this certainly is not the case for perspective cameras. 

But I double verified that this logic is mathematically correct by checking the implementation of opencv.

PR #3088, opened 2024-04-17
Allow alpha transparency training for blender datasets input

jb-ye:
It is a simple change such that user can use the following commands to do [alpha transparency](https://github.com/nerfstudio-project/nerfstudio/pull/2889) training on blender dataset:
```
ns-train splatfacto-big --data materials/ \
--pipeline.model.background-color random \
--pipeline.model.random_scale 2.6 \
--logging.local-writer.max-log-size=0 \
blender-data --alpha_color=None
```


kerrj (approved):
sgtm!

PR #3100, opened 2024-04-22
Added a learnable pure background color for splatfacto

KevinXu02:
Added a learnable pure background color for splatfacto. Useful when the bg is pure color.

### DIFF
--- a/nerfstudio/process_data/hloc_utils.py
+++ b/nerfstudio/process_data/hloc_utils.py
@@ -26,29 +26,6 @@ from typing import Literal
 from nerfstudio.process_data.process_data_utils import CameraModel
 from nerfstudio.utils.rich_utils import CONSOLE
 
-try:
-    # TODO(1480) un-hide pycolmap import
-    import pycolmap
-    from hloc import (  # type: ignore
-        extract_features,
-        match_features,
-        pairs_from_exhaustive,
-        pairs_from_retrieval,
-        reconstruction,
-    )
-except ImportError:
-    _HAS_HLOC = False
-
-else:
-    _HAS_HLOC = True
-
-try:
-    from pixsfm.refine_hloc import PixSfM  # type: ignore
-except ImportError:
-    _HAS_PIXSFM = False
-else:
-    _HAS_PIXSFM = True
-
 
 def run_hloc(
     image_dir: Path,
@@ -88,6 +65,29 @@ def run_hloc(
         refine_pixsfm: If True, refine the reconstruction using pixel-perfect-sfm.
         use_single_camera_mode: If True, uses one camera for all frames. Otherwise uses one camera per frame.
     """
+
+    try:
+        # TODO(1480) un-hide pycolmap import
+        import pycolmap
+        from hloc import (  # type: ignore
+            extract_features,
+            match_features,
+            pairs_from_exhaustive,
+            pairs_from_retrieval,
+            reconstruction,
+        )
+    except ImportError:
+        _HAS_HLOC = False
+    else:
+        _HAS_HLOC = True
+
+    try:
+        from pixsfm.refine_hloc import PixSfM  # type: ignore
+    except ImportError:
+        _HAS_PIXSFM = False
+    else:
+        _HAS_PIXSFM = True
+
     if not _HAS_HLOC:
         CONSOLE.print(
             f"[bold red]Error: To use this set of parameters ({feature_type}/{matcher_type}/hloc), "


nerfstudio/models/splatfacto.py
```diff
@@ -668,11 +674,13 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
                 background = self.background_color.to(self.device)
         else:
             optimized_camera_to_world = camera.camera_to_worlds[0, ...]
-
-            if renderers.BACKGROUND_COLOR_OVERRIDE is not None:
-                background = renderers.BACKGROUND_COLOR_OVERRIDE.to(self.device)
+            if self.config.background_color == "auto":
+                background = (torch.sigmoid(self.learnable_bg).to(self.device) * 1.1).clamp(0, 1)
```
jb-ye, line 678: what is this 1.1 multiplier for?

nerfstudio/models/splatfacto.py
```diff
@@ -770,21 +778,41 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
...
+                depths,
+                self.radii,
+                conics,
+                num_tiles_hit,  # type: ignore
+                rgbs,
+                opacities,
+                H,
+                W,
+                BLOCK_WIDTH,
+                background=torch.zeros(3, device=self.device),
```
jb-ye, line 793: Is it because the rasterize_gaussians() doesn't have grad for background parameter? If so, I think the proper fix is to add bwd grad for background input in gsplat library.

jb-ye (changes requested):

nerfstudio/models/splatfacto.py
```diff
@@ -770,21 +778,41 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
...
+                return_alpha=True,
+            )  # type: ignore
+            # alpha blend rgb with background
+            alpha = alpha[..., None]
+            rgb = alpha * rgb + (1 - alpha) * (torch.sigmoid(self.learnable_bg).to(self.device) * 1.1).repeat(H, W, 1)
```
jb-ye, line 798: Also I think you no longer need the extra `alpha * rgb` here, I believe. When you call rasterize_gaussians() the output rgb is already alpha blended. 

But again, I believe the proper implementation is to implement grad for background in rasterize_gaussians()

nerfstudio/models/splatfacto.py
KevinXu02: sigmoid itself can hardly produce pure white (also pure black), like those for blender dataset, or overexposed images. So 1.1 just allows it to generate pure white

nerfstudio/models/splatfacto.py
KevinXu02: I have implimented one on bg backporp in gsplat to handle 2d bg and it just needs way more modifications than directly alpha blended them and it changes a lots of logics. I think alpha blends here might be easier and more elegant.

jb-ye: I agree that 2d bg is tricky to implement in gsplat. But for just backprop the grad for bg color, I noticed you don't really need to touch cuda implementation of gsplat. There is a thin layer of wrapper, where you can just have [one line change there](https://github.com/nerfstudio-project/gsplat/blob/main/gsplat/rasterize.py#L245): by computing the grad as 

```
(v_out_img.view(-1, 3).t() @ final_Ts.view(-1, 1)).squeeze()
```

(Above line is NOT tested though)

KevinXu02:
Wondering if i should seperate get_outputs_for_camera from get_outputs to git rid of lots of if self.training. And prevent viser from possibly overwrite bg color (might already fixed by the viewer lock, not sure

nerfstudio/models/splatfacto.py
```diff
@@ -657,8 +662,9 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
         # get the background color
         if self.training:
             optimized_camera_to_world = self.camera_optimizer.apply_to_camera(camera)[0, ...]
-
-            if self.config.background_color == "random":
+            if self.config.background_color == "auto":
+                background = torch.sigmoid(self.learnable_bg).to(self.device)
```
jb-ye, line 666: Do we need `to(self.device)` here? I would assume self.learnable_bg will be converted to `self.device` automatically as part of `model.to(device)` in __init__ call of VanillaPipeline.

KevinXu02: testing on this

jb-ye:
The PR looks almost good to me (except one minor comment), but it will need to wait until the next gsplat version update before merging this PR (maybe in two weeks or so?).

KevinXu02:
> The PR looks almost good to me (except one minor comment), but it will need to wait until the next gsplat version update before merging this PR (maybe in two weeks or so?).
> 
> @KevinXu02 Do you have some experiments that demonstrate the auto background learns the proper background of the scene (e.g. blue sky)?

this seems trying to simutanously learning the color of the ground and the sky, thus more likely result in a color in the middle. so this might only makes learning the objects floating in a pure color background easier.
I'm currently working on predict 2d bg and that helps a lot in cleaning up floaters, but my current approach is still a bit slow and i'm still optimizing it.
image

KevinXu02:
image
Here's the bicycle scene predict 2d bg instead

jb-ye:
My general thoughts on this: allowing background to be learnable should be used with some regularizations (e.g. minimizing L1 norm of accumulation map etc.) Otherwise this still is uncontrollable. Let's say you captured a tree from ground, the image should have both leaf, branch, and sky. Would it be nice to absorb sky into the background parameter, otherwise the model would have blue blobs between tree branches, which is super weird.

jb-ye:
@KevinXu02 The PR will be merged when we produced a new version of gsplat library that support background parameter grad.

KevinXu02:
> @KevinXu02 The PR will be merged when we produced a new version of gsplat library that support background parameter grad.

Got it. Thank you!

PR #3111, opened 2024-04-26
GaussianSplats export: allow uint8 in ply

### LINKED ISSUE #3110: Gaussian Splat does not export to PLY when sh-degree is set to 0
When sh-degree is set to 0 the exporter script tries to convert all of the color values to uint8, which is a great idea and should be implemented this way

https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/scripts/exporter.py#L576

However this is incompatible with the write_ply function, which requires all data types to be float

https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/scripts/exporter.py#L516

One or the other should change, as it is it will never export.

mathijshenquet:
Fixes #3110

### DIFF
--- a/nerfstudio/scripts/exporter.py
+++ b/nerfstudio/scripts/exporter.py
@@ -490,9 +490,7 @@ class ExportGaussianSplat(Exporter):
     def write_ply(
         filename: str,
         count: int,
-        map_to_tensors: OrderedDict[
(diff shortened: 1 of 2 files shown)


mathijshenquet:
I still have to test this fix

mathijshenquet:
ping @jb-ye

Issue #3073, opened 2024-04-13
Camera Optimizer has different pose adjustments when applied to Raybundle and Camera.


nexuslrf:
Hi developers,

I found a pretty tricky computation inconsistency issue on `camera_optimizer`'s `apply_to_*` functions:

https://github.com/nerfstudio-project/nerfstudio/blob/babf57718c75e7d04f1b977f035db67c02a9c20a/nerfstudio/cameras/camera_optimizers.py#L148-L168

The rotation and translation shown in the above two functions are completely different. 
Assume the original pose is `[R, t]`, and pose adjustment in `CameraOptimizer` is `[dR, dt]`

* `apply_to_raybundle`: simply translates the `origins` (`t+dt`), and rotates the ray `directions` (`dR @ R @ ray_vec`) 

* `apply_to_camera`: new camera is `[R@dR, t+R@dt]`, when using this new camera to generate rays `origins` will be `t+R@dt`, `directions` transformation will be `R @ dR @ ray_vec`.

I initially wanted to evaluate the quality of camera optimization of a trained NeRF model by applying this `apply_to_camera` function but found such a computation mismatch. I guess these functions would also be confusing for other users (for sure 🤣 

It would be better in the future version to add some caution notes and an option to switch to different forms of pose adjustments.
For example,
```python
# raybundle_cam: bool
if raybundle_cam:
    R = torch.bmm(adj[:, :3, :3], camera.camera_to_worlds[:, :3, :3])
    t = adj[:, :3, 3:] + camera.camera_to_worlds[:, :3, 3:]
    camera.camera_to_worlds = torch.cat([R, t], dim=-1)
else:
    adj = torch.cat([adj, torch.Tensor([0, 0, 0, 1])[None, None].to(adj)], dim=1)
    camera.camera_to_worlds = torch.bmm(camera.camera_to_worlds, adj)
```

oseiskar:
@jb-ye I did not actually check this in detail in https://github.com/nerfstudio-project/nerfstudio/pull/2891/commits/e8e05ff02f1c604124a0373afcca7691f2b88546, and just kept the multiplication order from the [original 3DGS PR](https://github.com/nerfstudio-project/nerfstudio/pull/2521/files#diff-574309e78cc47396f2fd1e783b1695cddbd093a62333ccdfa63bfd659b2a79dfR158)

[...]

PR #3113, opened 2024-04-27
Use AbsGrad to get better results with less gaussians

jb-ye:
I am following the parameter setting as recommended in https://arxiv.org/pdf/2404.10484

In order to experiment this PR, one need to install the customized gsplat library as in this PR https://github.com/nerfstudio-project/gsplat/pull/166
```
# uninstall previous gsplat and nerfstudio

pip install git+https://github.com/nerfstudio-project/gsplat.git@absgrad
pip install git+https://github.com/jb-ye/nerfstudio
```

I will follow up my experiments here. I recommend to experiment with one of the two settings (A, B) and use input image resolution higher than 1600 (e.g. 2k or 4k)

O splatfacto-big in latest main branch (--rasterize_mode antialiased)
A splatfacto-big of this PR (--rasterize_mode antialiased --densify_grad_thresh=0.0008)
B splatfacto-big of this PR (--rasterize_mode antialiased --densify_grad_thresh=0.0004)

O. baseline.
A. faster than B, high quality (expected to be better than O), smaller PLY asset.
B. slowest training, highest quality, largest PLY asset.

PR #3116, opened 2024-04-28
Export Splatfacto-optimized poses

oseiskar:
Currently, `ns-export cameras` does not export the optimized poses even if Splatfacto pose optimization is enabled. Similarly, `ns-eval` does not take the pose optimization into account. This is because the "adjustment" to the initial pose is stored as a separate variable in the camera optimizer, which allows properly using the initial poses as priors. The NeRF code uses a different approach.

This PR is a slightly cleaned-up version of the approach used in https://github.com/SpectacularAI/3dgs-deblur, which is basically ensuring that the "camera index" and an instance of the CameraOptimizer are available in the relevant places. However, this seems a bit ugly and fragile.

Any ideas of to improve this? Permanently "applying" the adjustments to the `camera_to_worlds` at the end of training could be one option, which would probably help with `ns-export cameras` (but not the eval PSNR metrics exported to, e.g., tensorboard, which are also fixed in this PR).

This is also somewhat related to https://github.com/nerfstudio-project/nerfstudio/issues/3073.

### Testing 

First, download suitable data and install this branch (similarly to [here](https://github.com/nerfstudio-project/nerfstudio/pull/2891#issuecomment-2041027334)):
```bash
conda activate nerfstudio

# download a suitable dataset
git clone https://github.com/SpectacularAI/3dgs-deblur
cd 3dgs-deblur
python download_data.py --dataset synthetic
cd ..

# check out and install this branch
https://github.com/SpectacularAI/nerfstudio
cd nerfstudio
git checkout export-splatfacto-optimized-poses
pip install -e .
```
Train (example):
```
ns-train splatfacto --data ../3dgs-deblur/data/inputs-processed/synthetic-posenoise/factory \
    --pipeline.model.camera-optimizer.mode=SO3xR3 \
    --pipeline.model.rasterize-mode antialiased \
    --max-num-iterations 15000 \
    --output-dir outputs/pose-opt-test \
    nerfstudio-data --eval-mode all
```
Check metrics:
```
ns-eval --load-config outputs/pose-opt-test/splatfacto/TIMESTAMP/config.yml
```
Then see the metrics in `output.json`. With this PR, should show high PSNR (> 32). Without this, shows something quite low (< 20) due to the misaligned evaluation poses.

In addition, the poses exported with the following command are different (with this PR, they are the optimized ones, otherwise the original ones, possibly after some auto-scaling/transformations by the dataparser):
```
ns-export cameras \
    --load-config outputs/pose-opt-test/splatfacto/TIMESTAMP/config.yml \
    --output-dir outputs/pose-opt-test/
```


ichsan2895:
I see.. When I activate SO3xR3, I checked the image eval result, the rendered image is little bit shifting due to camera optimization vs GT. I think it was cuprit for metrics (PSNR, LPIPS, SSIM) degradation. So, it corrected the eval images aligned with corrected poses.

BTW, @jb-ye has overhauled the splatfacto in this draft (https://github.com/nerfstudio-project/nerfstudio/pull/3113). Hopefully it wont break each other.

jb-ye:
I am hesitant about exporting the optimized eval poses. This is very much only for the evaluation metrics reporting for paper write-ups. In general, I'm afraid we don't have a proper framework design in nerfstudio to handle eval pose optimization. 

I think it would be useful to export both original training poses and optimized training poses. At least this is useful to inspect the results. What do you think?

oseiskar:
> I see.. When I activate SO3xR3, I checked the image eval result, the rendered image is little bit shifting due to camera optimization vs GT. I think it was cuprit for metrics (PSNR, LPIPS, SSIM) degradation. So, it corrected the eval images aligned with corrected poses.

@ichsan2895 Sorry, I did not fully get this. Did you mean that this happens in the current Splatfacto or with this PR. What should happen is that in the current main branch, if you activate SO3xR3, you get the misalignment and metric degradation, but this PR fixes the issue.

> I am hesitant about exporting the optimized eval poses. This is very much only for the evaluation metrics reporting for paper write-ups. In general, I'm afraid we don't have a proper framework design in nerfstudio to handle eval pose optimization.

@jb-ye I agree. Currently, the proper framework for optimizing eval poses (i.e., optimizing their poses while keeping all Gaussian parameters fixed) is missing. For this reason, I have only tested the `--eval-mode=all` option (other `--eval-mode`s are broken with pose optimization anyway). Then training and evaluation sets are the same. However, maybe also in that case, it would be a cleaner solution to only output the optimized training poses and leave the evaluation poses as the originals.
 
> I think it would be useful to export both original training poses and optimized training poses. At least this is useful to inspect the results. So just make this PR focus on training poses, no? What do you think?

Currently, both the original and optmized training poses are available in the saved model. I think exporting the optimized training poses is a more reasonable default for `ns-export cameras`, but it could also be possible to add a command line flag there that would also allow exporting the original poses.

The original poses are also presumably available in the input data, but they do not necessarily match the output of `ns-export cameras` due to the various transforms in data parsers, so I also agree that both unoptimized and optimized training poses are potentially useful export targets.

jb-ye:
> it would be a cleaner solution to only output the optimized training poses and leave the evaluation poses as the originals.

Yes, I think keep the PR simple is important for long term maintenance. 

> I think exporting the optimized training poses is a more reasonable default for ns-export cameras, but it could also be possible to add a command line flag there that would also allow exporting the original poses.

That could also work. Export the original poses (post various transforms), so they can be used to compare with the optimized pose.

oseiskar:
Update:
 * rebased on latest main
 * fixed linter warnings
 * exporting optimized poses only for training images

To summarize, the following aspects should now be fixed in Splatfacto pose optimization with the options `--pipeline.model.camera-optimizer.mode=SO3xR3 --eval-mode all`:
 * camera poses saved to `transforms_train.json` with `ns-export cameras`
 * metric output from `ns-eval`
 * "Eval Images Metrics" graphs in Tensorboard

The following are left as future improvements:
 * Correct handling of other eval modes (this is a bigger revamp, as mentioned [here](https://github.com/nerfstudio-project/nerfstudio/pull/3116#issuecomment-2083179764))
 * Command line parameter for exporting unoptimized poses in `ns-export cameras`

Also note that the `--eval-mode all` case is relevant for practical reconstruction cases where there is not separate test/evaluation set.

nerfstudio/exporter/exporter_utils.py
```diff
@@ -304,7 +308,15 @@ def collect_camera_poses_for_dataset(dataset: Optional[InputDataset]) -> List[Di
     # new cameras are in cameras, whereas image paths are stored in a private member of the dataset
     for idx in range(len(cameras)):
         image_filename = image_filenames[idx]
-        transform = cameras.camera_to_worlds[idx].tolist()
+        if camera_optimizer is None:
+            transform = cameras.camera_to_worlds[idx].tolist()
+        else:
+            # print('exporting optimized camera pose for camera %d' % idx)
+            camera = cameras[idx]
```
jb-ye, line 315: If I set `camera = cameras[idx:idx+1]` instead, would it simplify how I would handle input in `apply_to_camera()`? I don't really like the fact that `apply_to_camera()` can handle different input shapes as it is implemented now.

nerfstudio/exporter/exporter_utils.py
```diff
@@ -331,7 +343,14 @@ def collect_camera_poses(pipeline: VanillaPipeline) -> Tuple[List[Dict[str, Any]
     eval_dataset = pipeline.datamanager.eval_dataset
     assert isinstance(eval_dataset, InputDataset)
 
-    train_frames = collect_camera_poses_for_dataset(train_dataset)
+    from nerfstudio.models.splatfacto import SplatfactoModel
+
+    camera_optimizer = None
+    if isinstance(pipeline.model, SplatfactoModel):
```
jb-ye, line 349: This is a custom code for SplatfactoModel. But I also have Nerfacto model which have camera optimizer enabled by default. Can we support both?

nerfstudio/pipelines/base_pipeline.py
```diff
@@ -376,6 +376,9 @@ def get_average_eval_image_metrics(
             for camera, batch in self.datamanager.fixed_indices_eval_dataloader:
                 # time this the following line
                 inner_start = time()
+                if camera.metadata is not None:
+                    # fixes metrics with Splatfacto pose optimization and --eval-mode=all
+                    camera.metadata["cam_idx"] = idx
```
jb-ye, line 381: This is something I am not sure if we have correct "cam_idx". Looking at the code, "cam_idx" key was set here: https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/datamanagers/full_images_datamanager.py#L303

Are they the same? with the order of  `self.datamanager.fixed_indices_eval_dataloader`? I feel we really need to establish a global way to fetch indices of camera across different data managers. 

What about nerfacto case which uses `ParallelDataManager`?

nerfstudio/exporter/exporter_utils.py
oseiskar: Hmm. Good point. In theory, yes, but if we want a generic solution instead of separate ifs for Spaltfacto and Nerfacto, we would also need to fix this: https://github.com/nerfstudio-project/nerfstudio/issues/3073

nerfstudio/pipelines/base_pipeline.py
oseiskar: I did not actually test this with Nerfacto at all. This simply seemed to fix the metrics with `--eval mode all` in Splatfacto (if the indices were wrong, this would probably immediately destroy the PSNR metric).

> I feel we really need to establish a global way to fetch indices of camera across different data managers.

I fully agree :+1: This part of the PR is very fragile without such a mechanism

oseiskar: On a second thought, this part of the PR is not critical and it could be safer to leave it out for now. The ability to export the optimized poses does not depend on this

nerfstudio/exporter/exporter_utils.py
jb-ye: Maybe we should address #3073 first, and use the same API call for apply_to_camera()? The additional change seems very manageable in my opinion. 

> keep the current behavior for NeRFs, but match the splatfacto behavior for consistency

oseiskar: Ok. I'll get back to this a bit later

oseiskar: @jb-ye Done: I tested that I now can:
 1. optimize poses using either Splatfacto or Nerfacto (with `--eval-mode all` and `--pipeline.model.camera-optimizer.mode=SO3xR3`)
 2. `ns-export cameras`
 3. modify the input data by replacing the original noisy poses with the optimized poses
 4. train Splatfacto on the modified input data without pose optimization, and get good results

My scripts for step 3 are a bit messy (based on [this](https://github.com/SpectacularAI/3dgs-deblur/blob/main/combine.py), mostly), but at least in principle, it now works.

nerfstudio/cameras/camera_optimizers.py
```diff
@@ -157,15 +157,15 @@ def apply_to_camera(self, camera: Cameras) -> torch.Tensor:
...
             return camera.camera_to_worlds
 
         camera_idx = camera.metadata["cam_idx"]
-        adj = self(torch.tensor([camera_idx], dtype=torch.long, device=camera.device))  # type: ignore
-        adj = torch.cat([adj, torch.Tensor([0, 0, 0, 1])[None, None].to(adj)], dim=1)
-        return torch.bmm(camera.camera_to_worlds, adj)
+        adj = self(torch.tensor([camera_idx], dtype=torch.long)).to(camera.device)  # type: ignore
+        c2w4x4 = torch.cat([camera.camera_to_worlds, torch.Tensor([0, 0, 0, 1])[None, None].to(camera.device)], dim=1)
+        # apply correction in world coordinates so that the adjustments are consistent with apply_to_raybundle
+        return torch.bmm(adj, c2w4x4)
```
jb-ye, line 168: Based on #3073 's discussion, the apply_to_raybundle is neither multiplying transform from left or right. I believe, one should return
```
rotations = torch.bmm(adj[..., :3, :3], camera.camera_to_worlds[..., :3, :3])
translations = camera.camera_to_worlds[..., :3, 3] + adj[..., :3, 3]
```

oseiskar: Thanks for checking. You're right... Quite weird that this actually worked quite well in [my test](https://github.com/nerfstudio-project/nerfstudio/pull/3116#discussion_r1615140185). Perhaps the angular adjustments were so small in the synthetic test that the difference between the two formulas was negligible. I'll anyway change it to match...

oseiskar: @jb-ye Now actually fixed. Here's my test results that show the subtle difference
 1. Artificial pose noise, Splatfacto reconstruction without pose optimization
 2. Poses optimized with Nerfacto, reconstruction with Splatfacto without pose opt. *old version*
 3. Same as above but with the new version


### 1

no-pose-opt

### 2

old

### 3

new

PR #3120, opened 2024-05-01
Added the possibility to import a metashape ply file

simonbethke:
This should allow people to export the pointcloud along with the camera poses from Metashape and migrate them to the nerfstudio format so its digestable by the nerfstudio training.

@brentyi maybe you can take a look at this :)

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -157,9 +157,13 @@ class FullImageDatamanager(DataManager, Generic[TDataset]):
         def undistort_idx(idx: int) -> Dict[str, torch.Tensor]:
             data = dataset.get_data(idx, image_type=self.config.cache_images_type)
             camera = dataset.cameras[idx].reshape(())
-            K = camera.get_intrinsics_matrices().numpy()
+            assert data["image"].shape[1] == camera.width.item() and data["image"].shape[0] == camera.height.item(), (
+                f'The size of image ({data["image"].shape[1]}, {data["image"].shape[0]}) loaded '
+                f'does not match the camera parameters ({camera.width.item(), camera.height.item()})'
+            )
             if camera.distortion_params is None:
                 return data
+            K = camera.get_intrinsics_matrices().numpy()
             distortion_params = camera.distortion_params.numpy()
             image = data["image"].numpy()
 
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -649,6 +649,13 @@ class SplatfactoModel(Model):
             return resize_image(image, d)
         return image
 
+    @staticmethod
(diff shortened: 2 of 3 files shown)


nerfstudio/process_data/metashape_utils.py
```diff
@@ -188,10 +191,21 @@ def metashape_to_json(
 
     data["frames"] = frames
 
+    summary = []
+
+    if isinstance(ply_filename, Path):
```
jb-ye, line 196: I believe a proper implementation (similar to what we did for colmap) is to save a "applied_transform" in the data dict (e.g. see [this line](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/process_data/colmap_utils.py#L479) in colmap utils), and then apply the applied_transform to point cloud (see [this line](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/process_data/colmap_utils.py#L690-L692) in colmap case).

Do you have a sample metashape exported data that I can play? I believe I can help make it in the way I suggested.

simonbethke: Sure! I am not at home at the moment, but I can provide data in a few hours

simonbethke: Here is a very small example: https://drive.google.com/file/d/1GQBYRRQO7opLUqQ2u7z-udcxrfC7zHaj/view?usp=sharing

simonbethke: Damn there is at least one misaligned camera... maybe it still can serve this purpose here. Else I could revisit and fix it. I am not good at using metashape yet. Still in trial phase and thinking if I should buy it or not.

nerfstudio/scripts/process_data.py
```diff
@@ -253,6 +258,12 @@ def main(self) -> None:
         if self.eval_data is not None:
             raise ValueError("Cannot use eval_data since cameras were already aligned with Metashape.")
 
+        if isinstance(self.ply, Path):
```
brentyi, line 261: and then `if self.ply is not None`? the current condition looks like it will always be `True`

simonbethke: If just were speaking python ^^
It was supposed to be optional. Sure I simply messed up. I will try fixing that

jb-ye:
@simonbethke Could you give me access to your nerfstudio repo/branch, so I can push changes there?

nerfstudio/process_data/metashape_utils.py
jb-ye: Can you try my revision here (I built on top of your changes) and verify it works for you? https://github.com/jb-ye/nerfstudio/tree/metashape_splatfacto

jb-ye:
I somehow messed up this PR, see the new PR here: https://github.com/nerfstudio-project/nerfstudio/pull/3122

PR #3122, opened 2024-05-01
Support splatfacto training for metashape exports

jb-ye:
test commands:

```
ns-process-data metashape --xml floorboard.xml --data floorboard/ --ply floorboard.ply --output-dir floorboard_nerfstudio/
```

### DIFF
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -650,8 +650,8 @@ class SplatfactoModel(Model):
         return image
 
     @staticmethod
-    def get_empty_outputs(camera, background):
-        rgb = background.repeat(int(camera.height.item()), int(camera.width.item()), 1)
+    def get_empty_outputs(width: int, height: int, background: torch.Tensor):
+        rgb = background.repeat(height, width, 1)
         depth = background.new_ones(*rgb.shape[:2], 1) * 10
         accumulation = background.new_zeros(*rgb.shape[:2], 1)
         return {"rgb": rgb, "depth": depth, "accumulation": accumulation, "background": background}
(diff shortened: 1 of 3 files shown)


jb-ye:
@simonbethke Could you try this PR and see if it works? I manually tested your dataset, and it seems fine.

simonbethke:
Yea, I will test it

simonbethke:
Okay, i tested a little bit. The happy path works!!

The unhappy paths are rather bad. Not worse than the average quality but when taking that as benchmark it will stay bad forever :D

1) When the ply-file has not .ply extension a stacktrace is printed. As this case is handled and even has a message, it should not print the stacktrace
2) when providing a non-existing .ply file, it takes forever, then there are some Open3D warnings and in the end, it summarizes that it imported the non-existing file which is wrong.

simonbethke:
I know handling unhappy paths is tideous and no fun. If you like, just ignore what I wrote about them.

brentyi:
For the stack trace verbosity, maybe we can add `ValueError` to the `try`/`except` statement?

https://github.com/nerfstudio-project/nerfstudio/blob/41deed6a52a33996b34053eea58bdc54e714573f/nerfstudio/scripts/process_data.py#L531C1-L539C1

jb-ye:
> For the stack trace verbosity, maybe we can add `ValueError` to the `try`/`except` statement?
> 
> https://github.com/nerfstudio-project/nerfstudio/blob/41deed6a52a33996b34053eea58bdc54e714573f/nerfstudio/scripts/process_data.py#L531C1-L539C1

Yes, I sent out a new revision.

brentyi (approved):
Looks good to me, thanks @simonbethke @jb-ye!

simonbethke:
Thank you both for your help!

simonbethke:
Regarding this PR:
unfortunately there is still a gap. If I transform something in metashape, the data cannot be exported/imported to nerfstudio anymore. I think the reason for this is, that in the metashape xml, another transform appears that is not used in the nerfstudio importer. However, I already tried to get this working but I was not able to. 
I have to provide sample data for that and it would be great if somebody could look into this :)

jb-ye:
@simonbethke Could you start a new issue and provide sample data there?

PR #3129, opened 2024-05-08
Add viewerplot element

chungmin99:


nerfstudio/viewer/viewer_elements.py
```diff
@@ -710,3 +715,96 @@ def _create_gui_handle(self, viser_server: ViserServer) -> None:
...
+        self._create_gui_handle(viser_server)
+        assert self.gui_handle is not None
+
+    @property
+    def figure(self):
+        assert self.gui_handle is not None
+        return self.gui_handle.figure
+
+    @figure.setter
+    def figure(self, figure: go.Figure):
```
kerrj, line 762: it would be nice to automatically apply the dark mode+aspect to new figures as well, so that the user doesn't have to call dark_mode and aspect every time they update/change something

chungmin99: `aspect` doesn't need to be set for new figures, but you're right about `dark_mode`. :+1:

chungmin99: Should be addr'ed in https://github.com/nerfstudio-project/nerfstudio/pull/3129/commits/ca04cbc0d9e44a2dba41fe18f23d5d428071cb43.

PR #3131, opened 2024-05-08
Make RealityCapture csv/ply exports work!

### LINKED ISSUE #3118: Current Reality Capture Nerfstudio Import process producing very bad quality for Splatfacto.
**Describe the bug**
The current nerfstudio instructions on importing from Reality Capture at: https://docs.nerf.studio/quickstart/custom_dataset.html#realitycapture don't produce expected results.

I have a suspicion that the issue may be in using the exported CSV from `Internal/External camera parameters` export option.

~Does the CSV export take into account the **distortion** the images undergo as Reality Capture optimizes the alignment?
Or is that do in Nerfstudio?~

This problem does not happen if using the colmap `.txt` and `colmap.db` format generated from the bundler `.out`  and imagelist `.lst` exports with undistorted images, using the process shown here https://gist.github.com/jo-chemla/258e6e40d3d6c2220b29518ff3c17c40

Then taking that colmap format and using the [instant-ngp](https://github.com/NVlabs/instant-ngp/blob/de507662d4b3398163e426fd426d48ff8f2895f6/scripts/colmap2nerf.py) `colmap2nerf.py` script to convert to a `transforms.json` file with this section of that script disabled in the `colmap2nerf.py` script:

```python
if __name__ == "__main__":
	args = parse_args()
	#if args.video_in != "":  <---
	#	run_ffmpeg(args)  <---
	# if args.run_colmap:    <---
	#	run_colmap(args)  <---
	AABB_SCALE=int(args.aabb_scale)
	SKIP_EARLY=int(args.skip_early)
	IMAGE_FOLDER=args.images
```

Using this newely generate transforms.json file with the undistorted images, results is far greater quality (as shown in the screenshot) below.

**To Reproduce**

**Expected behavior**
Expecting quality to not be this bad.

**Screenshots**
image

**Additional context**

jb-ye:
fix #3118 

Changes in this PR:

1. fix cx,cy calculation which was done wrong previously
2. add --ply option to `ns-process-data realitycapture` to support splatfacto
3. safeguard brown4 tangential2 distortion model in full data manager, since opencv's undistort doesn't support it.

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -339,6 +339,10 @@ def _undistort_image(
 ) -> Tuple[np.ndarray, np.ndarray, Optional[torch.Tensor]]:
     mask = None
     if camera.camera_type.item() == CameraType.PERSPECTIVE.value:
+        assert distortion_params[3] == 0, (
+            "We doesn't support the 4th Brown parameter for image undistortion, "
+            "Only k1, k2, k3, p1, p2 can be non-zero."
+        )
         distortion_params = np.array(
             [
                 distortion_params[0],
--- a/nerfstudio/process_data/realitycapture_utils.py
+++ b/nerfstudio/process_data/realitycapture_utils.py
@@ -78,11 +78,11 @@ def realitycapture_to_json(
         frame["file_path"] = image_filename_map[basename].as_posix()
         # reality capture uses the 35mm equivalent focal length
         # See https://en.wikipedia.org/wiki/35_mm_equivalent_focal_length
-        v = max(width, height) / 36.0
-        frame["fl_x"] = float(cameras["f"][i]) * v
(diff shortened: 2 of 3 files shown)


shengjie-lin:
@jb-ye You mentioned "safeguard brown4 tangential2 distortion model in full data manager, since opencv's undistort doesn't support it.". However, from my testing, changing `k4` would affect the behavior of `cv2.undistort`. And sensors like Azure Kinect does provide k4 in the API. Can you please give some explanation on this? Thank you!

shengjie-lin:
Thanks for the explanation @jb-ye . Does this mean COLMAP or hloc tool box also estimate the distortion coefficients in a non-compatible format from what opencv expects (specifically for the k4)?
Edit: sorry, I just checked and realized hloc only estimates k1, k2, p1, p2.

jb-ye:
@shengjie-lin I am not sure about COLMAP, it support several cameras models. The FULL_OPENCV might be something you are looking for.

PR #3162, opened 2024-05-23
Fixing resize_image and adding additional distortion check

AntonioMacaronio:
There is a runtime error when boolean masks are fed into `splatfacto`: (For me, this happened when testing with aria data)

image

Additionally, when the distortion parameters are all 0 (meaning there is no distortion in the image), splatfacto will still try to distort the image because `distortion_params` are not None, so adding a quick check on that prevents images from being fed to the undistort function when there is no distortion.


jb-ye (approved):
Thx

Issue #3160, opened 2024-05-23
Extraction of demo capture dataset zip contains two directories so it fails


martin-kokos:
**Describe the bug**
Download and extraction of demo data per https://docs.nerf.studio/quickstart/first_nerf.html fails because there is a MACOS directory bundled in the zip

**To Reproduce**
Run `ns-download-data nerfstudio --capture-name=poster`
```
(nerfstudio) u@u-pc-q35-8-1:~$ ns-download-data nerfstudio --capture-name=poster
Downloading...
From (original): https://drive.google.com/uc?id=1FceQ5DX7bbTbHeL26t0x6ku56cwsRs6t
From (redirected): https://drive.google.com/uc?id=1FceQ5DX7bbTbHeL26t0x6ku56cwsRs6t&confirm=t&uuid=537d8006-5a22-4b91-9ca6-ec346850bbf1
To: /home/u/data/nerfstudio/poster.zip
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 749M/749M [01:26<00:00, 8.68MB/s]
Traceback (most recent call last):
  File "/home/u/miniconda3/envs/nerfstudio/bin/ns-download-data", line 8, in <module>
    sys.exit(entrypoint())
  File "/home/u/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/nerfstudio/scripts/downloads/download_data.py", line 573, in entrypoint
    main(tyro.cli(Commands))
  File "/home/u/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/nerfstudio/scripts/downloads/download_data.py", line 567, in main
    dataset.download(dataset.save_dir)
  File "/home/u/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/nerfstudio/scripts/downloads/download_data.py", line 207, in download
    download_capture_name(save_dir, "nerfstudio", self.capture_name, capture_name_to_file_id=nerfstudio_file_ids)
  File "/home/u/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/nerfstudio/scripts/downloads/download_data.py", line 174, in download_capture_name
    assert len(inner_folders) == 1, f"There is more than one folder inside this zip file: {inner_folders}"
AssertionError: There is more than one folder inside this zip file: ['poster', '__MACOSX']
```

**Expected behavior**
demo data is downloaded

PR #3164, opened 2024-05-24
Fix "poster" dataset extraction

martin-kokos:
Skip over `__MACOSX` directory

Fixes https://github.com/nerfstudio-project/nerfstudio/issues/3160


maturk (approved):
Thanks for the fix, LGTM!

PR #3172, opened 2024-05-27
Fix: when --use-sfm-depth is enabled and the model is given by --colm…

Serenitysmk:
in `ns-process-data`
if the user provides an existing colmap reconstruction model by `--colmap-model-path` and also enables `--use-sfm-depth`, nerfstudio still tries to search `points3D.bin` in the output colmap reconstruction directory instead of the user provided colmap model path.

This is a small fix for it


jb-ye (approved):
Thanks

PR #3177, opened 2024-05-29
Introduce FPSampler for train cameras

jb-ye:
Currently, the train cameras are sampled uniformly random without replacement in full_images_manager. There are cases when one wants to implement a different camera sampler other than uniformly random sampler. For example, this PR implements a type of sampler as an alternative option that can avoid oversampling cameras that are very close to each other (using farthest distance sampling), because I observed that oversampling train cameras from very similar views can potential create unwanted floaters. 

Choosing an optimal distribution over a given set of cameras to train is a non-trivial problem and have been discussed in literature (e.g. https://arxiv.org/pdf/2109.02369 ). But this topic goes beyond the scope of this PR. 

This PR also fixed the seed of train camera sampler for improving reproducibility of experiments. One can also set a different seed from command argument.

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -72,12 +72,13 @@ class FullImageDatamanagerConfig(DataManagerConfig):
     train_cameras_sampling_strategy: Literal["random", "fps"] = "random"
     """Specifies which sampling strategy is used to generate train cameras, 'random' means sampling 
     uniformly random without replacement, 'fps' means farthest point sampling which is helpful to reduce the artifacts 
-    due to oversampling subsets of cameras that are very closer to each other."""
+    due to oversampling subsets of cameras that are very close to each other."""
     train_cameras_sampling_seed: int = 42
     """Random seed for sampling train cameras. Fixing seed may help reduce variance of trained models across 
     different runs."""
     fps_reset_every: int = 100
-    """The number of iterations before one resets fps sampler repeatly"""
+    """The number of iterations before one resets fps sampler repeatly, which is essentially drawing fps_reset_every
+    samples from the pool of all training cameras without replacement before a new round of sampling starts."""
 
 
 class FullImageDatamanager(DataManager, Generic[TDataset]):
@@ -162,6 +163,10 @@ class FullImageDatamanager(DataManager, Generic[TDataset]):
             )
             n = self.config.fps_reset_every
             if num_train_cameras < n:
+                CONSOLE.log(
+                    f"num_train_cameras={num_train_cameras} is smaller than fps_reset_ever={n}, the behavior of "
+                    "camera sampler will be very similar to sampling random without replacement (default setting)."
+                )
                 n = num_train_cameras
             kdline_fps_samples_idx = fpsample.bucket_fps_kdline_sampling(data, n, h=3)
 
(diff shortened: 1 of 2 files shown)


maturk:
@jb-ye, very cool idea! Do you have examples showing what kind of differences once could potentially expect using the random or fps method?

jb-ye:
@maturk It highly depends on the distribution of train cameras. Most standard benchmark dataset you probably won't notice any difference. But on many of our internal datasets (acquired from a scan machine), you see improvements. Here is a nice example to your ask:

https://github.com/nerfstudio-project/nerfstudio/assets/132313008/cce95487-52ca-44b5-b71a-10460852d5ab

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -68,6 +69,15 @@ class FullImageDatamanagerConfig(DataManagerConfig):
...
     """The maximum number of threads to use for caching images. If None, uses all available threads."""
+    train_cameras_sampling_strategy: Literal["random", "fps"] = "random"
+    """Specifies which sampling strategy is used to generate train cameras, 'random' means sampling 
+    uniformly random without replacement, 'fps' means farthest point sampling which is helpful to reduce the artifacts 
+    due to oversampling subsets of cameras that are very closer to each other."""
+    train_cameras_sampling_seed: int = 42
+    """Random seed for sampling train cameras. Fixing seed may help reduce variance of trained models across 
+    different runs."""
+    fps_reset_every: int = 100
+    """The number of iterations before one resets fps sampler repeatly"""
```
maturk, line 80: @jb-ye, this option is great, but I think the docstring should be improved, it took me some time to understand the logic here.  I have a small doubt, if `fps_reset_every < num_train_cameras` then is it possible to construct a setting under which it is possible to never sample a training camera(s) during training using this method, taking into account the weighting of the `train_unsampled_epoch_count` cameras in L:161?

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -123,12 +133,44 @@ def __init__(
...
+            camera_origins = self.train_dataset.cameras.camera_to_worlds[..., 3].numpy()
+            # We concatenate camera origins with weighted train_unsampled_epoch_count because we want to
+            # increase the chance to sample camera that hasn't been sampled in consecutive epochs previously.
+            # We assume the camera origins are also rescaled, so the weight 0.1 is relative to the scale of scene
+            data = np.concatenate(
+                (camera_origins, 0.1 * np.expand_dims(self.train_unsampled_epoch_count, axis=-1)), axis=-1
+            )
+            n = self.config.fps_reset_every
+            if num_train_cameras < n:
+                n = num_train_cameras
```
maturk, line 165: Should there be a small warning in place here? Because if the user accidentally sets the condition `num_train_cameras < fps_reset_every` then the fps method is almost identical to the default random sampling of unseen cameras method? Although the ordering of returned cameras from next_train() is different, all cameras are still sampled during a sampling epoch. Is my understanding correct?

maturk (commented):
Thanks, @jb-ye, I appreciate the PR and your efforts in merging this kind of stuff upstream, I learned a lot from this. I left some comments and questions. I think it LGTM, just some minor concerns that could be better explained in the docstrings.

nerfstudio/data/datamanagers/full_images_datamanager.py
jb-ye: I haven't tested your hypothetic setting. In theory, when fps_reset_every is very small, the sampling would be more similar to random with replacement except that the effect of using `train_unsampled_epoch_count` we can at least bound its maximum though the use of weighting.

nerfstudio/data/datamanagers/full_images_datamanager.py
jb-ye: Correct, let me add a new warning.

maturk (approved):
LGTM

PR #3182, opened 2024-05-30
[Draft] [Do Not Review] Depth Supervision for Gaussian Splatting

mattstrong-stanford:
TBD

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@
     <img alt="nerfstudio" src="https://docs.nerf.studio/_images/logo.png" width="400">
 </p> -->
 
+test
+
 <p align="center"> A collaboration friendly studio for NeRFs </p>
 
 <p align="center">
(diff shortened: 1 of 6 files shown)


kerrj:
Hi! you might also find the work from @maturk interesting here: https://maturk.github.io/dn-splatter/

maturk:
Hi @mattstrong-stanford, would you be interested in making a pr with your depth-ranking implementation within dn-splatter? In that project, I tried to support depth supervision with real sensors (either z-depths or euclidian/ray depths) as well as some minimal monocular supervision (via alignment of scale/shift). It also supports normal supervision within the 3DGS context, but I also plan to support a 2DGS variant in the future. I already have a few different [depth losses](https://github.com/maturk/dn-splatter/blob/bc83595b0685f0fba0f0b440d510bb9123a8fdb2/dn_splatter/losses.py#L19) supported, although the list is not exhaustive. Let me know what you think, happy to collab on depth GS stuff.

peasant98:
Hi @maturk, I would love to collaborate on depth-GS (I was using my other Github account because of different forked versions).

I and another robotics graduate student are planning to hopefully integrate 2D-GS into Nerfstudio soon, so I bet we could get that done as well.

Maybe we could sync up to collab via email? My email is mastro1@stanford.edu.

tripp528:
Is there a plan to get this merged? I would love this feature

Issue #3195, opened 2024-06-04
install_checks doesn't use colmap_cmd


SimonF24:
**Describe the bug**
When using ColmapConverterToNerfstudioDataset from nerfstudio.process_data.colmap_converter_to_nerfstudio_dataset.py in __post_init__ the colmap install check from nerfstudio.utils.install_checks.check_colmap_installed() doesn't use the colmap_cmd attribute, instead using shutil.which('colmap') so effectively the only colmap_cmd allowed is 'colmap' and colmap has to be on the PATH. 

**To Reproduce**
Remove colmap from your PATH and provide an alternate colmap_cmd to an instance of ImagesToNerfstudioDataset() e.g. on Windows
```
from nerfstudio.process_data.images_to_nerfstudio_dataset import ImagesToNerfstudioDataset
from pathlib import Path

colmap_output_dir = Path("/colmap/")
colmap_cmd = "C:/Users/<user>/Downloads/COLMAP-3.9.1-windows-cuda/COLMAP-3.9.1-windows-cuda/colmap"
data_dir = Path("C:/Users/<user>/Downloads/data/images/")
 
images_to_nerfstudio_dataset = ImagesToNerfstudioDataset(
    colmap_cmd=colmap_cmd,
    data=data_dir,
    output_dir=colmap_output_dir
)
```

**Expected behavior**
Providing an alternate colmap_cmd should allow ImagesToNerfstudioDataset to run normally

**Additional context**
Either colmap_cmd should be removed from ImagesToNerfstudioDataset and the colmap command should be assumed to be "colmap" or the install check should be made to also use colmap_cmd.

jb-ye:
You are right, could you send a fix for it? I think add the additional parameter to `check_colmap_installed` make senses.

PR #3204, opened 2024-06-09
Making install_checks use colmap_cmd

SimonF24:
check_colmap_installed() in nerfstudio.utils.install_checks.py previously didn't use the colmap_cmd attribute of ColmapConverterToNerfstudioDataset. This change addresses that so an alternative colmap_cmd can actually be used. This addresses issue #3195


brentyi (approved):
Looks reasonable to me!

PR #3211, opened 2024-06-11
Adds back screen size too bigs culling as an option

KevinXu02:


nerfstudio/models/splatfacto.py
```diff
@@ -757,16 +783,27 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
...
+
         if render_mode == "RGB+ED":
             depth_im = render[:, ..., 3:4]
             depth_im = torch.where(alpha > 0, depth_im, depth_im.detach().max()).squeeze(0)
         else:
             depth_im = None
-        return {"rgb": rgb.squeeze(0), "depth": depth_im, "accumulation": alpha.squeeze(0), "background": background}  # type: ignore
+
+        if background.shape[0] == 3:
+            background = background.expand(H, W, 3)
```
jb-ye, line 799: Why do you want to expand background dims?

nerfstudio/models/splatfacto.py
```diff
@@ -757,16 +783,27 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
...
+
+        return {
+            "rgb": rgb.squeeze(0),  # type: ignore
+            "depth": depth_im,  # type: ignore
+            "accumulation": alpha.squeeze(0),  # type: ignore
+            "background": background.squeeze(0),  # type: ignore
```
jb-ye, line 805: I don't understand if you squeeze(0) for tensor shape in (H, W, 3), what would you expect?

Also can you also validate if the current code allow alpha transparency training on blender dataset. See [this](https://github.com/nerfstudio-project/nerfstudio/pull/3088) for instructions.

jb-ye (commented):
Thanks for bring in the culling logic back!

nerfstudio/models/splatfacto.py
KevinXu02: I copied this part from my splatw pr and i forget to clean it all up(

KevinXu02: Yes, transparency training on blender dataset still works

nerfstudio/models/splatfacto.py
KevinXu02: Becaue the viser will broke if tries to check the background without expanding it.

jb-ye (approved):
LGTM

PR #3216, opened 2024-06-12
Dataloading Revamp

AntonioMacaronio:
**Problems and Background**
- With a sufficiently large enough dataset, the current `parallel_datamanager.py` will try to cache the entire dataset into RAM, which will lead to an out-of-memory (OOM) error 
- `parallel_datamanager.py` only uses one worker to generate ray bundles. Since various subprocesses such as unprojecting during ray generation, or pixel sampling within a custom mask can be a CPU-intensive task, it may be better suited to parallelize this. While `parallel_datamanager.py` does support multiple workers, each worker caches the entire dataset to RAM and it does not support massive datasets, leading to duplicate copies of the dataset in computer memory. It also implements parallelism from scratch and is not friendly to build off.
- Additionally, both `VanillaDataManager` and `ParallelDataManager` rely on CacheDataloader, which subclasses `torch.utils.data.DataLoader`, which is a strange coding practice, and actually serves no particular use in the current nerfstudio implementation.
- Similarly for `full_images_datamanager.py`: As we can not fit the entire dataset in RAM, the current implementation loads in entire dataset into the `FullImageDataloader`'s `cached_train` attribute. To do this efficiently, we need multiprocess parallelization to load in images, undistort them, and do this quickly to keep up with GPU's forward and backward passes of the model.

**Overview of Changes**
- Replacing `CacheDataloader` with `RayBatchStream`, which subclasses `torch.utils.data.IterableDataset`. The goal of this class is to generate ray bundles directly without caching all images to RAM. This is done by collating a sampled batch of images to sample from. A new `ParallelDatamanager` class is written which is available side-by-side but can completely replace the original `VanillaDatamanager`
- Adding an `ImageBatchStream` to create a parallel, OOM-resistant version of `FullImageDataManager`. This can be configured to load from the disk by setting `cache_images` config variable to disk.
- A new `pil_to_numpy()` function is added. This function reads a PIL.Image's data buffer and fills an empty numpy array while reading, hastening the conversion process and removing an extra memory allocation. It is the fastest way to get from a PIL Image to a Pytorch tensor averaging ~2.5ms for a 1080x1920 image (~40% faster)
- A new flag called `cache_compressed_imgs` now caches your images to RAM in their compressed form (for example, caching) and relies on parallelized CPU dataloading to efficiently decode them into pytorch tensors to be used in training.
- Resolving some pyright issues: Within `nerfstudio/scripts/exporter.py`, the assertions for `ExportPointCloud` and `ExportPoissonMesh` were modified because these are only used on NeRFs, so exporting for splats (has its own export method) and RandomCameraDatamanger (outdated) were removed. Similarly, some "# type: ignore" were added to various runtime checked locations that pyright could not detect. This was in `base_pipeline.py` and `trainer.py`.

**Impact**
- Checkout [these comparisons](https://antoniomacaronio.github.io/cs180-webpages/random/index.html)! The left was trained on 200 images of a 4k video, while the right was trained on 2000 images of the same 4k video. 

**Additional Diagrams**
NOTE: THESE TESTS WERE CONDUCTED ON A 24GB 3090Ti
image
image
image
image
image


pwais (commented):
nice progress!  sorry its not fast but i think i know why:

i think the main reason this is slower than expected is because `_get_collated_batch()` gets called *per raybundle* and sadly `_get_collated_batch()` is AFAIK needlessly slow.  

 * take note about how the current `CachedDataloader` avoids doing `_get_collated_batch()` per raybundle.  it would have been nice for the author to have left some notes about how slow `_get_collated_batch()` is, but evidently that author found it's necessary to _not collate images per raybundle_ .
 * in my impl, I just `_get_collated_batch()` once on a small set of images an keep that batch cached.  the main problem I saw is that `_get_collated_batch()` on thousands of images seemed to use 2x or 3x as much RAM as actually needed and thus cause many minutes of swapping and stuff

Even if you only call `_get_collated_batch()` once tho, you might need a bigger prefetch factor and/or more workers depending on the model.

IMO it's worth trying to find a way to get the result of `nerfstudio_collate` on cameras (I think the cameras do need to be collated because they can be ragged? i could be wrong and they don't need collation)  but on _images_ just have the worker read image files / buffers and never call collate on those tensors.

Just to be clear, this is the line where collate on images can go nuts and start taking forever to allocate 200GB or more of RAM for many images in code in `main`: https://github.com/nerfstudio-project/nerfstudio/blob/3d27dd4609faef5acc5784dd1ef335ec7f329c15/nerfstudio/data/utils/nerfstudio_collate.py#L101 

So! If a worker is just emitting _raybundles_ then the images never need to be in shared tensor memory then eh?  Thus should be able to save some RAM and CPU by skipping that line for images.  Still need to think about the cost of reading the images themselves, but collate is definitely a troublemaker.

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -313,9 +317,9 @@ class VanillaDataManagerConfig(DataManagerConfig):
     """Specifies the dataparser used to unpack the data."""
     train_num_rays_per_batch: int = 1024
     """Number of rays per batch to use per training iteration."""
-    train_num_images_to_sample_from: int = -1
```
pwais, line 316: this is just for tesing / hacking, or are you proposing these params could be good for everybody?  Either way if you happen to decide on new parameters you think are good for everybody, good to highlight and discuss that in the summary and/or discussion portion of the Pull Request

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -335,6 +339,16 @@ class VanillaDataManagerConfig(DataManagerConfig):
...
+    prefetch_factor: int = 4 # increasing prefetch_factor was not beneficial
+    """The limit number of batches a worker will start loading once an iterator is created. 
+    More details are described here: https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader"""
+    dataloader_num_workers: int = 1
+    """The number of workers performing the dataloading from either disk/RAM, which 
+    includes collating, pixel sampling, unprojecting, ray generation etc."""
+    use_ray_train_dataloader: bool = True
+    """Allows parallelization of the dataloading process with multiple workers."""
+    cache_binaries: bool = True
+    """If True, caches the images as binary strings to RAM"""
```
pwais, line 351: in python 3.x there's `bytes` type that helps differentiate versus in python 2.x one had to have "binary data in strings"

so maybe just "When enabled, cache raw image files to RAM"

nerfstudio/data/utils/data_utils.py
```diff
@@ -31,7 +63,7 @@ def get_image_mask_tensor_from_path(filepath: Path, scale_factor: float = 1.0) -
         width, height = pil_mask.size
         newsize = (int(width * scale_factor), int(height * scale_factor))
         pil_mask = pil_mask.resize(newsize, resample=Image.Resampling.NEAREST)
-    mask_tensor = torch.from_numpy(np.array(pil_mask)).unsqueeze(-1).bool()
+    mask_tensor = torch.from_numpy(pil_to_numpy(pil_mask)).unsqueeze(-1).bool()
```
pwais, line 66: what was wrong with like `numpy.asarray(pil_image)` or can we maybe use https://pytorch.org/vision/main/generated/torchvision.transforms.functional.pil_to_tensor.html  ?   `pil_to_numpy()` looks odd.. tho i understand there's PIL resize() for some reason, i wonder if the resize could actually not use PIL (nerfstudio uses cv2.resize() a lot ) ...  not sure why the orig authors chose PIL resize

nerfstudio/data/utils/dataloaders.py
```diff
@@ -133,6 +134,7 @@ def __iter__(self):
                 collated_batch = self.cached_collated_batch
             elif self.first_time or (
                 self.num_times_to_repeat_images != -1 and self.num_repeated >= self.num_times_to_repeat_images
+                # if it's the first time, we need to
```
pwais, line 137: need to ... ?

pwais (commented):
just took a quick look (can't do a full review right now), so cool to see this coming along!!

Sounds like this change will target the case that _uncompressed image tensors_ can't fit in RAM, but the raw image files (typically jpeg) do fit in RAM.  In that case I guess we do want each worker to literally load the file bytes into Python RAM (as implemented) versus let the OS disk cache work, because the idea is that the uncompressed image tensors will otherwise blow out the disk cache.  

I think it would be important to test in the end like a case where the user only has limited RAM (say 16GB) and e.g. a 8GB laptop graphics card, in that case I think there are moderate or larger image datasets where the whole thing would OOM when using the current cache impl.  In that case, it would be helpful to have some way to disable the cache, or just communicate to the user that they simply have too weak of a machine for the dataset (e.g. just a `CONSOLE.print("[bold yellow]Warning ...")`  in the line where the workers start reading image files into RAM.

nerfstudio/data/utils/dataloaders.py
AntonioMacaronio: ahh this is just a random comment i wrote for myself, will remove this 😅

nerfstudio/data/datamanagers/base_datamanager.py
AntonioMacaronio: Yep this is just some hacking for now - I have not done too many tests changing these figures around

nerfstudio/data/utils/data_utils.py
AntonioMacaronio: After some digging + discussion, doing `asarray` leads an unwriteable numpy array, which could have conflicts when converting to Pytorch tensors described [here](https://github.com/pytorch/pytorch/blob/07adae3dac85d51303e52bbaaa3d45df6da54ae5/torch/csrc/utils/tensor_numpy.cpp#L205)

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -86,14 +90,38 @@ def variable_res_collate(batch: List[Dict]) -> Dict:
...
+        pixel_area=torch.cat([ray_bundle_i.pixel_area for ray_bundle_i in ray_bundle_list], dim=0),
+        camera_indices=torch.cat([ray_bundle_i.camera_indices for ray_bundle_i in ray_bundle_list], dim=0),
+        metadata=combined_metadata,
+    )
+    concatenated_batch = {
+        "image" : torch.cat([batch_i["image"] for batch_i in batch_list], dim=0),
+        "indices": torch.cat([batch_i["indices"] for batch_i in batch_list], dim=0),
+    }
+    # end = time.time()
+    # print((end - start) * 1000)
```
pwais, line 121: curious, this function is pretty fast right?

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -178,6 +206,7 @@ def __init__(self):
         self.train_count = 0
         self.eval_count = 0
         if self.train_dataset and self.test_mode != "inference":
+            # print(self.setup_train) # prints <bound method ParallelFullImageDatamanager.setup_train of ParallelFullImageDatamanager()>
```
pwais, line 209:
```suggestion
```

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -355,6 +394,204 @@ def __post_init__(self):
...
+        self.device = device
+        # self.collate_fn = variable_res_collate # variable_res_collate avoids collating images, which is much faster than `nerfstudio_collate`
+        self.collate_fn = collate_fn
+        print("collate_fn", self.collate_fn)
+        print("self.device", self.device)
+        self.num_image_load_threads = num_image_load_threads  # kwargs.get("num_workers", 4) # nb only 4 in defaults
+        self.exclude_batch_keys_from_device = exclude_batch_keys_from_device
+        # print("self.exclude_batch_keys_from_device", self.exclude_batch_keys_from_device) # usually prints ['image']
+        self.datamanager_config = datamanager_config
+        self.pixel_sampler: PixelSampler = None
```
pwais, line 442:
```suggestion
        self.pixel_sampler: Optional[PixelSampler] = None # lazy init
```

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -355,6 +394,204 @@ def __post_init__(self):
...
+        self.ray_generator: RayGenerator = None
```
pwais, line 443:
```suggestion
        self.ray_generator: Optional[RayGenerator] = None # lazy init
```

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -547,5 +563,193 @@ def _undistort_image(
...
+## Let's implement a parallelized splat dataloader!
+def undistort_idx(idx: int, dataset: TDataset, image_type: Literal["uint8", "float32"] = "float32") -> Dict[str, torch.Tensor]:
+    """Undistorts an image to one taken by a linear (pinhole) camera model and updates the dataset's camera intrinsics to a linear camera model"""
+    data = dataset.get_data(idx, image_type)
+    camera = dataset.cameras[idx].reshape(())
+    # dataset.cameras.width[idx] = data["image"].shape[1]
+    # dataset.cameras.height[idx] = data["image"].shape[0]
+    if idx == 48:
+        # breakpoint()
+        print("beginning", camera.width, camera.height)
```
pwais, line 579:
```suggestion
```

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -547,5 +563,193 @@ def _undistort_image(
...
+        data["mask"] = mask
+    # dataset.cameras.fx[idx] = float(K[0, 0])
+    # dataset.cameras.fy[idx] = float(K[1, 1])
+    # dataset.cameras.cx[idx] = float(K[0, 2])
+    # dataset.cameras.cy[idx] = float(K[1, 2])
+    # dataset.cameras.width[idx] = image.shape[1]
+    # dataset.cameras.height[idx] = image.shape[0]
+    # dataset.cameras.distortion_params = None
+    if idx == 48:
+        print("ending", camera.width, camera.height)
```
pwais, line 603:
```suggestion
```

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -547,5 +563,193 @@ def _undistort_image(
...
+        return data
+    K = camera.get_intrinsics_matrices().numpy()
+    distortion_params = camera.distortion_params.numpy()
+    image = data["image"].numpy()
+    K, image, mask = _undistort_image(camera, distortion_params, data, image, K)
+    # print(image.shape[1]) # outputs 539
+    # print(cameras[48].reshape(()).width.item()) # outputs 540
+    data["image"] = torch.from_numpy(image)
+    if mask is not None:
+        data["mask"] = mask
```
pwais, line 594: hmm, doesn't the mask need to be rectified / undistorted too? not sure if the code in `main` does this

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -547,5 +563,193 @@ def _undistort_image(
...
+                i = 0
+            idx = worker_indices[i] # idx refers to the actual datapoint index this worker will retrieve
+            data, camera = undistort_view(idx, self.input_dataset, self.config.cache_images_type)
+            camera2 = self.input_dataset.cameras[idx : idx + 1]#.to(self.device)
+            if camera.metadata is None:
+                camera.metadata = {}
+            camera.metadata["cam_idx"] = idx
+            i += 1
+            if torch.sum(camera.camera_to_worlds) == 0:
+                print(i, camera.camera_to_worlds, "YOYO INSIDE IMAGEBATCHSTREAM")
```
pwais, line 688: yoyo im inside ur GPU eating ur RAMs!

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -547,5 +563,193 @@ def _undistort_image(
...
+                r.shuffle(worker_indices)
+                i = 0
+            idx = worker_indices[i] # idx refers to the actual datapoint index this worker will retrieve
+            data, camera = undistort_view(idx, self.input_dataset, self.config.cache_images_type)
+            camera2 = self.input_dataset.cameras[idx : idx + 1]#.to(self.device)
+            if camera.metadata is None:
+                camera.metadata = {}
+            camera.metadata["cam_idx"] = idx
+            i += 1
+            if torch.sum(camera.camera_to_worlds) == 0:
```
pwais, line 687: ratrow, this doesn't mean the splat optimizer will get empty poses as well?

nerfstudio/models/splatfacto.py
```diff
@@ -724,7 +724,7 @@ def get_outputs(self, camera: Cameras) -> Dict[str, Union[torch.Tensor, List]]:
             render_mode = "RGB+ED"
         else:
             render_mode = "RGB"
-
+        # breakpoint()
```
pwais, line 727:
```suggestion
```

nerfstudio/scripts/datasets/process_project_aria.py
```diff
@@ -118,25 +123,101 @@ def read_trajectory_csv_to_dict(file_iterable_csv: str) -> TimedPoses:
     )
 
 
+def undistort_image_and_calibration(
```
pwais, line 126: maybe the changes in this file / module should be broken out into separate PR at some point.  and could probably ship sooner too then (?)

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -547,5 +563,193 @@ def _undistort_image(
...
+        self,
+        datamanager_config: DataManagerConfig,
+        input_dataset: TDataset,
+        device,
+    ):
+        self.config = datamanager_config
+        self.input_dataset = input_dataset
+        self.device = device
+
+    def __iter__(self):
```
pwais, line 658: iteresting so this works fast enough to undistort and not need caching of the undistorted image?  curious what CPU / GPU combo and number of workers was good here.  

or maybe it does slow training a bit, but rather *it works without OOM* on bigger datasets :)

pwais (commented):
Wow the visual results look amazing! Thank you so much for continuing to hack on this!

nerfstudio/data/datamanagers/full_images_datamanager.py
AntonioMacaronio: Yep! the `_undistort_image` function on the main branch returns an undistorted mask

nerfstudio/data/datamanagers/full_images_datamanager.py
AntonioMacaronio: Yeah it does slow training quite a bit. With 1 worker, training time went from 9 minutes -> 18 minutes (with downsampling to a resolution inside 1600x1600)

It also depends on what type of undistortion is occuring. I only have tested this with datasets of CAMERA.PERSPECTIVE types, but with fisheye and equirectangular camera models, it will definitely take longer to undistort this, will do some further benchmarks on this!

pwais: ah interesting!  2x is not that bad, given that this change makes training _feasible_ when low RAM.

maybe caching the undistorted image could help then?  so in your other code, you try to cache the jpeg bytes which helps, and clearly here you can't necessarily cache the compressed image and so you have to cache the undistorted image.  

all that said, it's very nice to even have training _enabled_ even if u don't implement / test caching

nerfstudio/data/datamanagers/base_datamanager.py
AntonioMacaronio: Actually it's not that fast, it takes around 0.24 seconds per raybundle which adds a lot of time overall - I've removed this because of that

nerfstudio/data/datasets/base_dataset.py
```diff
@@ -46,7 +46,7 @@ class InputDataset(Dataset):
     exclude_batch_keys_from_device: List[str] = ["image", "mask"]
     cameras: Cameras
 
-    def __init__(self, dataparser_outputs: DataparserOutputs, scale_factor: float = 1.0, cache_images: bool = False):
+    def __init__(self, dataparser_outputs: DataparserOutputs, scale_factor: float = 1.0):
```
pwais, line 49: oh doh did this not end up being helpful?  I guess the I/O can get hidden enough otherwise?

AntonioMacaronio: So originally there is no increase in training speed when loading the image from disk vs. caching the compressed image to RAM; I was getting about 25 minutes for nerfacto to complete 30,000 training steps for both methods. However, the caching of compressed images to RAM took an extra ~2 minutes in the setup time

image

I actually think I may have made a mistake in deleting this because I do have an M.2 SSD drive which is very very speedy. I think this could actually make a difference if a user has a much slower drive they are reading from, so I will lean towards adding this back

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -484,19 +668,55 @@ def setup_train(self):
...
-            exclude_batch_keys_from_device=self.exclude_batch_keys_from_device,
-        )
-        self.iter_train_image_dataloader = iter(self.train_image_dataloader)
-        self.train_pixel_sampler = self._get_pixel_sampler(self.train_dataset, self.config.train_num_rays_per_batch)
-        self.train_ray_generator = RayGenerator(self.train_dataset.cameras.to(self.device))
+
+        if self.config.use_ray_train_dataloader:
+            import torch.multiprocessing as mp
+
+            mp.set_start_method("spawn")
```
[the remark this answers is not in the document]
AntonioMacaronio: for splatting and full image datamanaging, workers don't need to use CUDA, but the user can configure the images to be in the GPU if needed, so I'll move this there.

For nerfacto and ray methods, the workers do use CUDA when pixel sampling, so I'll have to keep that

pwais:
nice work keep going!

tineras:
Did anything ever come of this? I'm running into OOM errors as are many others due to large image sets. Would love to see this implemented.

kerrj:
@jb-ye would be good to get some more eyes on this, its very close to ready to merge, we've tested with external methods, train/test/render/export for nerfacto/splatfacto and we're hoping to merge in the next couple days

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -352,6 +340,26 @@ def __post_init__(self):
...
+            self.train_num_images_to_sample_from = (
+                50 if self.train_num_images_to_sample_from == -1 else self.train_num_images_to_sample_from
+            )
+            self.train_num_times_to_repeat_images = (
+                10 if self.train_num_times_to_repeat_images == -1 else self.train_num_times_to_repeat_images
+            )
+            self.prefetch_factor = self.train_num_times_to_repeat_images if self.use_parallel_dataloader else None
+
+        if self.use_parallel_dataloader:
+            try:
```
chungmin99, line 357: Should we raise a warning here?

nerfstudio/data/datasets/base_dataset.py
```diff
@@ -66,17 +82,20 @@ def get_numpy_image(self, image_idx: int) -> npt.NDArray[np.uint8]:
...
+        if self.cache_compressed_images:
+            pil_image = Image.open(self.binary_images[image_idx])
+        else:
+            pil_image = Image.open(image_filename)
         if self.scale_factor != 1.0:
             width, height = pil_image.size
             newsize = (int(width * self.scale_factor), int(height * self.scale_factor))
             pil_image = pil_image.resize(newsize, resample=Image.Resampling.BILINEAR)
-        image = np.array(pil_image, dtype="uint8")  # shape is (h, w) or (h, w, 3 or 4)
+        image = pil_to_numpy(pil_image)  # # shape is (h, w) or (h, w, 3 or 4) and dtype == "uint8"
```
chungmin99, line 93: Nit, but can we move the shape/dtype specifications into asserts in the function instead?

nerfstudio/pipelines/base_pipeline.py
```diff
@@ -297,6 +296,9 @@ def get_train_loss_dict(self, step: int):
             step: current iteration step to update sampler if using DDP (distributed)
         """
         ray_bundle, batch = self.datamanager.next_train(step)
+        ray_bundle = ray_bundle.to(
+            self.device
+        )  # for some reason this line of code is needed otherwise viewmats will not be invertible?
```
chungmin99, line 301: Is this comment related to this: https://github.com/pytorch/pytorch/issues/90613?

AntonioMacaronio: So this line is addressing a super weird bug I’ve been getting when there is more than >0 workers on my machine in specific instances that I’ve had trouble replicating but comes up every now and then and I apologize for not bringing this up earlier!

The bug: For example when running splatfacto, within`datamanager.next_train()` of `full_image_datamanager.py`, all tensor fields of ray_bundle are correct. For instance I can do `print(ray_bundle.camera_to_worlds)` and these poses are correct. However within `base_pipeline.py`, when I add the print statement of the c2w matrices these tensors become 0's, which leads to an error down the stack later in splatfacto.py

I’m going to hold off on merging until I can get this sorted out

jb-ye:
> @jb-ye would be good to get some more eyes on this, its very close to ready to merge, we've tested with external methods, train/test/render/export for nerfacto/splatfacto and we're hoping to merge in the next couple days

@f-dy might be interested to take a look since it is a major change impacting the data workflow.

nerfstudio/pipelines/base_pipeline.py
AntonioMacaronio: On my machine, the way this issue gets addressed is by keeping the `ray_bundle` or `camera` TensorDataclass object on the CPU until the object is used in `base_pipeline.py`.Because all models such as SplatfactoModel or NerfactoModel assume that the `ray_bundle` or `camera` is already on the GPU, I am keeping this `ray_bundle` variable on the CPU until right before it gets inputted into the model to prevent its fields from getting set to 0.

chungmin99: I see -- I also couldn't replicate the bug from commenting out the line, with the latest up-to-date code.

BTW, are you sure that `ray_bundle` is on CPU? Personally, I found that `raybundle` is already on `cuda` even before we hit `ray_bundle = ray_bundle.to(device)`, so the line is a no-op. It also seems that both `RayBatchStream` and `ImageBatchStream` move all data to the input device in the `__iter__` calls?

pwais:
Very cool to see this moving along!! Congrats @AntonioMacaronio !!  What sort of datasets sizes have you tested so far, like 500 4k images for nerfacto / depth-nerfacto as well as splatfacto? I would be curious to test now that it's closer to launch!

kerrj:
#3569 will fix the failing tests

kerrj:
> Very cool to see this moving along!! Congrats @AntonioMacaronio !! What sort of datasets sizes have you tested so far, like 500 4k images for nerfacto / depth-nerfacto as well as splatfacto? I would be curious to test now that it's closer to launch!

@pwais Anthony made [this comparison](https://antoniomacaronio.github.io/cs180-webpages/random/index.html) of 200 vs 2000 4k images.

docs/developer_guides/pipelines/datamanagers.md
```diff
@@ -94,3 +94,100 @@ See the code!
...
+ns-train nerfacto --data {PROCESSED_DATA_DIR} --pipeline.datamanager.load_from_disk
+```
+
+To train splatfacto with a large dataset that's unable to fit in memory, please set the device of `cache_images` to disk. For example with splatfacto:
+```bash
+ns-train splatfacto --data {PROCESSED_DATA_DIR} --pipeline.datamanager.cache_images disk
+```
+
+## Migrating Your DataManager to the new DataManager 
+Many methods subclass a DataManager and add extra data to it. If you would like your custom datamanager to also support new parallel features, you can migrate any custom dataloading logic to the new `custom_view_processor()` API. Let's take a look at an example for the LERF method, which was built on Nerfstudio's VanillaDataManager. This API provides an interface to attach new information to the RayBundle (for ray based methods), Cameras object (for splatting based methods), or ground truth dictionary. It runs in a background process if disk caching is enabled, otherwise it runs in the main process.
```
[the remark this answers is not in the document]
kerrj: OK updated the names

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -352,6 +340,26 @@ def __post_init__(self):
...
+        Note: decreasing train_num_images_to_sample_from and increasing train_num_times_to_repeat_images alleviates CPU bottleneck.
+        """
+        if self.load_from_disk:
+            self.train_num_images_to_sample_from = (
+                50 if self.train_num_images_to_sample_from == -1 else self.train_num_images_to_sample_from
+            )
+            self.train_num_times_to_repeat_images = (
+                10 if self.train_num_times_to_repeat_images == -1 else self.train_num_times_to_repeat_images
+            )
+            self.prefetch_factor = self.train_num_times_to_repeat_images if self.use_parallel_dataloader else None
```
brentyi, line 354: is it possible to avoid these mutations? it seems simpler if we keep the config objects as simple data structures with minimal logic (if possible)

for setting values to `50` and `10`, can we just set these as default values for the config field?

nerfstudio/data/datamanagers/base_datamanager.py
```diff
@@ -352,6 +340,26 @@ def __post_init__(self):
...
+
+        if self.use_parallel_dataloader:
+            try:
+                torch.multiprocessing.set_start_method("spawn")
```
brentyi, line 358: it seems a little bit odd to set this kind of global flag in the config object's `__post_init__`... is there a better place? maybe right before wherever the processes or dataloader is created?

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -79,6 +81,29 @@ class FullImageDatamanagerConfig(DataManagerConfig):
...
+    More details are described here: https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader"""
+    cache_compressed_images: bool = False
+    """If True, cache raw image files as byte strings to RAM."""
+
+    def __post_init__(self):
+        if self.cache_images == "disk":
+            # If a user would like to load from disk, we pre-emptively set the number of
+            # workers and prefetch factor to parallelize the dataloading process.
+            try:
+                torch.multiprocessing.set_start_method("spawn")
```
brentyi, line 98: same comment as above!

brentyi (commented):
left some minor comments. overall looks reasonable to me!

nerfstudio/data/datamanagers/base_datamanager.py
AntonioMacaronio: I agree with this! will change

nerfstudio/data/datamanagers/base_datamanager.py
kerrj: Yea I agree, maybe in the `__init__` of the datamanager?

nerfstudio/data/datamanagers/base_datamanager.py
AntonioMacaronio: So after a little more investigation, it may not be possible to avoid this. the default value of -1 does serve a purpose when the user wants to load the entire dataset's images into CPU RAM. When `train_num_times_to_repeat_images==-1`, the RayBatchStream will assign each worker an even partition and together these partitions will form the entire dataset.

With `prefetch_factor`, it's default value is None represents when the user doesn't want to use a parallel_dataloader, it has to be None for an error to not occur when initializing pytorch's dataloader

brentyi (approved):

kerrj (approved):
LGTM!!!! amazing work thanks Anthony!!!

pwais:
Holy Cow!! 🐮 🐮 🐮 !  Congrats @AntonioMacaronio  !!! 🎉  🎉 🎉

AntonioMacaronio:
Thank you for all y'alls help!! @pwais @kerrj @brentyi @chungmin99

vincentwoo:
nice to see you around @pwais

PR #3219, opened 2024-06-13
Small fix when evaluating splatfacto model, running ns-eval

ajean757:
In the `_get_background_color()` function in splatfacto.py, `self.background_color` is allocated on the cpu, when calling `get_outputs()`, we get a runtime error "RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!". Added a `.to(self.device)` to ensure this doesn't happen


maturk (approved):
Thanks for the fix!

KevinXu02:
Thanks for catching this!

PR #3220, opened 2024-06-13
Add direct render+export buttons and notifications

ginazhouhuiwu:
Directly render and export things from the viewer, with notifications!

Uses a simple notification addition from nerfstudio-project/viser#225.

https://github.com/nerfstudio-project/nerfstudio/assets/42229107/147d3fd9-1c8e-432e-a616-9f2b5e2a0435

___
TODO:
- [ ] thoroughly test export functionality

### DIFF
--- a/nerfstudio/scripts/render.py
+++ b/nerfstudio/scripts/render.py
@@ -439,6 +439,8 @@ class RenderCameraPath(BaseRender):
     """Filename of the camera path to render."""
     output_format: Literal["images", "video"] = "video"
     """How to save output data."""
+    complete: bool = True
+    """Whether rendering is complete"""
 
     def main(self) -> None:
         """Main function."""
@@ -575,7 +577,7 @@ class RenderCameraPath(BaseRender):
                 if str(left_eye_path.parent)[-5:] == "_temp":
                     shutil.rmtree(left_eye_path.parent, ignore_errors=True)
                 CONSOLE.print("[bold green]Final VR180 Render Complete")
-
+        self.complete = True
 
 @dataclass
 class RenderInterpolated(BaseRender):
--- a/nerfstudio/scripts/render.py
+++ b/nerfstudio/scripts/render.py
@@ -37,17 +37,35 @@ import viser.transforms as tf
 from jaxtyping import Float
 from rich import box, style
 from rich.panel import Panel
-from rich.progress import BarColumn, Progress, TaskProgressColumn, TextColumn, TimeElapsedColumn, TimeRemainingColumn
+from rich.progress import (
+    BarColumn,
+    Progress,
+    TaskProgressColumn,
+    TextColumn,
+    TimeElapsedColumn,
+    TimeRemainingColumn,
+)
 from rich.table import Table
 from torch import Tensor
 from typing_extensions import Annotated
 
-from nerfstudio.cameras.camera_paths import get_interpolated_camera_path, get_path_from_json, get_spiral_path
+from nerfstudio.cameras.camera_paths import (
+    get_interpolated_camera_path,
+    get_path_from_json,
+    get_spiral_path,
+)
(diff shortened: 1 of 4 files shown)


maturk:
Nice PR. @ginazhouhuiwu, is it possible to also add the option to "capture frame" or something similar, where only a single image is saved instead of a video.

ginazhouhuiwu:
that sounds like a great feature I will try implementing! Thanks for taking a look : )

maturk:
> that sounds like a great feature I will try implementing! Thanks for taking a look : )

Yeah, it has been a feature that I have seen some users bring up a few times in the past, just to quickly take some snapshots from a certain angle (perhaps using the same camera intrinsics as the original training dataset, but I think that is configurable in the gui anyways). Just some additional thoughts ... :)

kerrj:
Super cool! Would also be great to build out some functionality for previewing the render as it goes and perhaps canceling it. Would be cool to also integrate with viser's download functionality to download the final render directly to someone's machine :)

ginazhouhuiwu:
Local download! 

https://github.com/nerfstudio-project/nerfstudio/assets/42229107/901dcdd9-58c3-4730-ad47-6c0b7afbb9be

ginazhouhuiwu:
Some finishing touches with suggestions from @kerrj, namely disabling render button when render is already in progress or if no keyframes are added, only enabling download button once file is ready, and allowing render to be canceled. The generate command button is added back as an additional option.

This PR is ready once nerfstudio-project/viser#225 is.

https://github.com/user-attachments/assets/89b16cb0-c401-4945-8668-efa28576120d

nerfstudio/scripts/render.py
```diff
@@ -432,6 +455,11 @@ class BaseRender:
     """If true, checks line-of-sight occlusions when computing camera distance and rejects cameras not visible to each other"""
     camera_idx: Optional[int] = None
     """Index of the training camera to render."""
+    kill_flag: List[bool] = field(default_factory=lambda: [False])
```
kerrj, line 458: Can we not expose this in the config? it feels like it should be an invisible class `self._kill_flag` variable, then the kill() function sets that to true.

nerfstudio/scripts/render.py
```diff
@@ -442,6 +470,8 @@ class RenderCameraPath(BaseRender):
     """Filename of the camera path to render."""
     output_format: Literal["images", "video"] = "video"
     """How to save output data."""
+    complete: bool = True
```
kerrj, line 473: Same with this, it seems like complete should be a `@property` that gets managed internally

kerrj:
Some observations from testing it out:
- If you load a path the render button doesn't become available to click until you add another keyframe.
- If you click render before 2000 steps, there's no error message except in the terminal (probably shouldn't give a rendering notification unless the render has successfuly started)

nerfstudio/scripts/render.py
brentyi: Still a hack but one option is:
```python
# `tyro.conf.Suppress` hides this field from the generated `ns-render` CLI.
_kill_flag: tyro.conf.Suppress[List[bool]] = field(default_factory=lambda: [False])
```

nerfstudio/scripts/render.py
brentyi: I'm not really following the connection to `@property`

ginazhouhuiwu: I can hack around this too by suppressing? I think as a field it does need to still exist for the progress(notification) update.

kerrj:
How is this going? would be good to merge soon so it doesn't go forgotten about

ginazhouhuiwu:
Fixed:
- error message when user clicks render/export before 2000 training steps
- able to directly load existing path for rendering 

Screenshot 2024-11-06 at 3 58 59 PM

PR #3223, opened 2024-06-13
Fix: make sure background and alpha are on the same device

hoanhle:


jb-ye:
Thanks, closed as dup of https://github.com/nerfstudio-project/nerfstudio/pull/3219

PR #3227, opened 2024-06-14
Allow Phototourism Dataparser to load 3D points

KevinXu02:


jb-ye:
I am a bit confused. Are Phototourism datasets in colmap format? If so, why not directly use colmap parser?

PR #3218, opened 2024-06-12
Forcing setuptools version to 69.5.1 to overcome build error.

aravindhankrishnan:
Using the latest version of setuptools (70.0.0) gives an ImportError on "import packaging". Reverting to 69.5.1 resolves this issue.

PR #3241, opened 2024-06-18
force setuptools version in Dockerfile

jakepopham:
Docker container fails to build on main since setuptools released [v70.0.0](https://github.com/pypa/setuptools/releases/tag/v70.0.0)

Repro:

```
sudo docker build --tag nerfstudio -f Dockerfile .
...
ImportError: cannot import name 'packaging' from 'pkg_resources' 
```

Fix:

Force setuptools to be strictly less than v70 in the Dockerfile, and it builds correctly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,7 +127,7 @@ WORKDIR /home/${USERNAME}
 ENV PATH="${PATH}:/home/${USERNAME}/.local/bin"
 
 # Upgrade pip and install packages.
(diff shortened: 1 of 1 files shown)


jakepopham:
Side note: I think the numpy 2.0 release broke some tests.

jakepopham:
duplicate of https://github.com/nerfstudio-project/nerfstudio/pull/3218

Issue #2811, opened 2024-01-23
Render only a specific frame


briancantwe:
I looked through the code a bit, but didn't seem to find what I was looking for.  Is there a way to render only a specific frame, or a specific frame range of a camera move using ns-render?  

Thanks so much!

kerrj:
There currently isn't great support for single image rendering, you can either
- save the viewer canvas as a PNG (in viser, click the control knobs icon->export canvas)
- run `ns-render images` to save output frames in a folder as pngs instead of an .mp4, then choose the one you want

briancantwe:
Got it.  Yeah, I'm currenting rendering the entire range and then picking
the frames I need.   I guess it's a feature request.  Thanks for
considering it!

paolovic:
> There currently isn't great support for single image rendering, you can either
> 
> * save the viewer canvas as a PNG (in viser, click the control knobs icon->export canvas)
> * run `ns-render images` to save output frames in a folder as pngs instead of an .mp4, then choose the one you want

can you give me some hint, please, how I can make this possible for splatfacto?
I simply cannot find the function that renders the scene since 2 days and I keep on digging through the codebase...
how is the view that I see in the viewer created?
There must underly some function that does something like that, must'nt it? And I want to give this function my pose (lcoation+orientation) and need a single image in return...

brentyi:
For rendering the scene, maybe you can work backward from here? https://github.com/nerfstudio-project/nerfstudio/blob/dd811f58bce4dd56a75147c63218eb190fc8143c/nerfstudio/scripts/render.py#L195-L198

paolovic:
> For rendering the scene, maybe you can work backward from here?
> 
> https://github.com/nerfstudio-project/nerfstudio/blob/dd811f58bce4dd56a75147c63218eb190fc8143c/nerfstudio/scripts/render.py#L195-L198

I did, respectively, I tried. In the meantime, I thought that `render.py` is not the right place to investigate since render follows a trajectory between my training images and just renders this in a video/images. That's why I moved to `viewer.py` since the viewer allows me to move to arbitrary places in my 3D model and renders the view correspondingly.
But I might miss the forest for the trees and investigate `get_outputs_for_camera` further.
Thank you very much!

paolovic:
So, I extended my nerfstudio render.py by the following feature

[...]

PR #3251, opened 2024-06-23
Render specific frames, panoramic views from specific poses, or uniformly sampled from camera path 

paolovic:
Hi,

as described in #2811 I added the following functionality:

* Enter a desired eye position in world coordinates, and optionally a desired target, and get in return the images of the  panoramic view from that pose in parallel to the floor. Anything is adjustable: eye, target, view angle, intermediate steps
* Example usage: `ns-render pose-view --output-format images --load-config outputs/room1/splatfacto/2024-06-07_104843/config.yml --eye "0 0 -2" --load-dataparser-transforms outputs/room5/splatfacto/2024-06-07_104843/dataparser_transforms.json`

* Sample uniformly from the camera path and get the views from the corresponding view points
* Example usage: `ns-render pose-view --output-format images --load-config outputs/room1/splatfacto/2024-06-07_104843/config.yml --viewpoints 3`

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -307,15 +307,7 @@ class FullImageDatamanager(DataManager, Generic[TDataset]):
         """Returns the next evaluation batch
 
         Returns a Camera instead of raybundle"""
-        image_idx = self.eval_unseen_cameras.pop(random.randint(0, len(self.eval_unseen_cameras) - 1))
-        # Make sure to re-populate the unseen cameras list if we have exhausted it
-        if len(self.eval_unseen_cameras) == 0:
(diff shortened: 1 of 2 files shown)


pwais:
instead of (yet again) wrapping the already-convoluted `_render_trajectory_video()`, perhaps this could be a script and/or module that outputs a camera_path json file, then the user can just use the existing `render.py` with that input.  and then there's also a camera path fixture that can be (1) unit tested and (2) debugged outside the render loop

paolovic:
> instead of (yet again) wrapping the already-convoluted `_render_trajectory_video()`, perhaps this could be a script and/or module that outputs a camera_path json file, then the user can just use the existing `render.py` with that input. and then there's also a camera path fixture that can be (1) unit tested and (2) debugged outside the render loop

Hi,
thanks for the feedback, and feedback addressed.

Issue #3250, opened 2024-06-22
"Downloading and Processing Data" section in Collab Notebook doesn't work


feoandrew:
I have an issue with collab notebook when i an trying to launch "Downloading and Processing Data".
When i am choosed the dozer scene and launched section i saw this error report:

A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.

Traceback (most recent call last):  File "/usr/local/bin/ns-download-data", line 5, in <module>
    from nerfstudio.scripts.downloads.download_data import entrypoint
  File "/usr/local/lib/python3.10/dist-packages/nerfstudio/scripts/downloads/download_data.py", line 33, in <module>
    from nerfstudio.process_data import process_data_utils
  File "/usr/local/lib/python3.10/dist-packages/nerfstudio/process_data/process_data_utils.py", line 25, in <module>
    import cv2
  File "/usr/local/lib/python3.10/dist-packages/cv2/__init__.py", line 181, in <module>
    bootstrap()
  File "/usr/local/lib/python3.10/dist-packages/cv2/__init__.py", line 153, in bootstrap
    native_module = importlib.import_module("cv2")
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
AttributeError: _ARRAY_API not found
Traceback (most recent call last):
  File "/usr/local/bin/ns-download-data", line 5, in <module>
    from nerfstudio.scripts.downloads.download_data import entrypoint
  File "/usr/local/lib/python3.10/dist-packages/nerfstudio/scripts/downloads/download_data.py", line 33, in <module>
    from nerfstudio.process_data import process_data_utils
  File "/usr/local/lib/python3.10/dist-packages/nerfstudio/process_data/process_data_utils.py", line 25, in <module>
    import cv2
  File "/usr/local/lib/python3.10/dist-packages/cv2/__init__.py", line 181, in <module>
    bootstrap()
  File "/usr/local/lib/python3.10/dist-packages/cv2/__init__.py", line 153, in bootstrap
[...]

PR #3254, opened 2024-06-24
Fix numpy upgrade to 2.x unexpectedly

yzslab:
[rawpy](https://github.com/letmaik/rawpy/releases) release 0.22.0 two days ago and drop the numpy 1.x support. But some libraries required by nerfstudio are not compatible with numpy 2.x  yet.

Related issue: https://github.com/nerfstudio-project/nerfstudio/issues/3250

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ dependencies = [
     "pyquaternion>=0.9.9",
     # TODO we can switch back to (non-new) rawpy if they start releasing arm64
     # wheels. https://github.com/letmaik/rawpy/issues/171#issuecomment-1572627747
-    "rawpy>=0.18.1; platform_machine != 'arm64'",
+    "rawpy>=0.18.1,<0.22.0; platform_machine != 'arm64'",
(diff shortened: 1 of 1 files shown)


yzslab:
This will only happen when python version > 3.8. But the README already provided a command to create an python 3.8 enviroment. So the PR closed.

xrd:
I believe this breaks docker support right now as docker builds/installs with 3.10.

This fix does not seem to work (modifying pyproject.toml) https://github.com/nerfstudio-project/nerfstudio/pull/3254

It is not possible to change `python3.10` to `python3.8` in the apt get commands inside the Dockerfile.

yzslab:
> I believe this breaks docker support right now as docker builds/installs with 3.10.
> 
> This fix does not seem to work (modifying pyproject.toml) #3254
> 
> It is not possible to change `python3.10` to `python3.8` in the apt get commands inside the Dockerfile.

Adding a "numpy<2" to `dependencies` of the toml file should work. Or simply run `pip install "numpy<2"` after environment setup.

PR #3256, opened 2024-06-25
Fix parsing of non ascii image names in colmap's images.bin

TamirCohen:
If colmap image file name is non ascii an error will be thrown in the line `current_char.decode("utf-8")` while trying to parse the file `images.bin`
It is thrown because it tries to decode a single byte as `utf-8`, And non ascii chars are represented using more bytes than 1.

My solution proposal is to read all the bytes, and than decode it.
I tested it and it worked OK

### DIFF
--- a/nerfstudio/data/utils/colmap_parsing_utils.py
+++ b/nerfstudio/data/utils/colmap_parsing_utils.py
@@ -235,11 +235,12 @@ def read_images_binary(path_to_model_file):
             qvec = np.array(binary_image_properties[1:5])
             tvec = np.array(binary_image_properties[5:8])
             camera_id = binary_image_properties[8]
-            image_name = ""
+            image_name = b""
             current_char = read_next_bytes(fid, 1, "c")[0]
             while current_char != b"\x00":  # look for the ASCII 0 entry
-                image_name += current_char.decode("utf-8")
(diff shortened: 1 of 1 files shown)


kerrj (approved):
looks good, also tested on a few datasets and it works thanks!

TamirCohen:
Thanks @kerrj.
Due to an error in the sphynx build I can not merge it (I do not think it is related to my commit).
Should I wait after https://github.com/nerfstudio-project/nerfstudio/pull/3265 is fixed and then merge it?

kerrj:
Could you re-run lint please? the linter was updated along with the np 2.0 patch

TamirCohen:
I reran it and it did not work because there are empy comments in:
https://github.com/nerfstudio-project/nerfstudio/blob/d6df39b70d2e0b444178d23b694989e39d624781/nerfstudio/configs/method_configs.py#L304
https://github.com/nerfstudio-project/nerfstudio/blame/d6df39b70d2e0b444178d23b694989e39d624781/tests/pipelines/test_vanilla_pipeline.py#L41

Should we remove these empty comment in another PR or in this PR?

jb-ye:
Just rebase to the top main should resolve the issue.

PR #3257, opened 2024-06-25
Add option to use mask_path while running COLMAP

hoanhle:


hoanhle:
docs is failing because of unrelated error https://github.com/spyder-ide/spyder/issues/22187

PR #3262, opened 2024-06-26
Draft: visualize region of interest

hoanhle:


kerrj:
Seems like this might've been intended for an external repo?

PR #3266, opened 2024-06-27
added missing installation of gsplat for linux

hardikdava:


brentyi:
Hm, is it necessary to install `gsplat` from source? Shouldn't it be installed automatically from `pip install nerfstudio`?

https://github.com/nerfstudio-project/nerfstudio/blob/90cc4f1d1e0dbae3dd358f6d8a106ce820f754bc/pyproject.toml#L66

KevinXu02:
I think might only Windows need to install it from source because the vc environments.

hardikdava:
@KevinXu02 @brentyi I am installing nerfsudio on ubuntu and if i install gsplat from pip then I am getting this error:

```
  File "/home/hd/Desktop/workspace/ns/.venv/lib/python3.10/site-packages/gsplat/cuda/_wrapper.py", line 12, in call_cuda
    return getattr(_C, name)(*args, **kwargs)
AttributeError: 'NoneType' object has no attribute 'fully_fused_projection_fwd'
```

liruilong940607:
I need to inspect the failure building cases on windows.. Adding it to my TODO list

SharkWipf:
I just noticed this PR, my PR #3333 has some more background on gsplat on Windows (and is the polar opposite of this PR lol).

Manually adding gsplat on Linux should _not_ be necessary and there's probably something else going on, the reason it was added to the Windows documentation has to do with Visual Studio being a pain.

Including manually installing gsplat in the documentation should be avoided if possible, as this will break on new Nerfstudio/Gsplat releases and the documentation will need to be updated each update.

hardikdava:
@SharkWipf if you already included this PR fixes then we can close this PR.

SharkWipf:
(Sorry, seems I forgot to reply to this)

PR #3268, opened 2024-06-27
Update pytorch in Dockerfile to 2.1.2

### LINKED ISSUE #3235: Docker torch version should be updated from 2.0.1
Hi,

I am not a Docker user, but the torch version should be bumped up to 2.1.2 + [here](https://github.com/nerfstudio-project/nerfstudio/blob/2e3c3c179a4cbfba96c6ecfa6a9d1bb3d21dc569/Dockerfile#L135). Would make a PR myself but I am a Docker noob.

Zunhammer:
Update pytorch to fix gsplat incompatibilities with versions >= 1.0.0
Additionally, version pinning of tinycudann as 1.6 was removed as it is not compatible with the new pytorch
setuptools is now fixed to 69.5.1 also because of incompatibilities
Currently, numpy 2 creates an error but as others are already working on fixing this in nerfstudio I'm skipping it in the Dockerfile. In case you need to fix it now please pin the version in nerfstudio dependencies or add a line in the Dockerfile before the ns-install-cli command:
RUN python3.10 -m pip install --no-cache-dir numpy==1.26.4

Fixes #3235

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -127,18 +127,18 @@ WORKDIR /home/${USERNAME}
 ENV PATH="${PATH}:/home/${USERNAME}/.local/bin"
 
 # Upgrade pip and install packages.
-RUN python3.10 -m pip install --no-cache-dir --upgrade pip setuptools pathtools promise pybind11 omegaconf
+RUN python3.10 -m pip install --no-cache-dir --upgrade pip setuptools==69.5.1 pathtools promise pybind11 omegaconf
 
 # Install pytorch and submodules
 # echo "${CUDA_VERSION}" | sed 's/.$//' | tr -d '.' -- CUDA_VERSION -> delete last digit -> delete all '.'
 RUN CUDA_VER=$(echo "${CUDA_VERSION}" | sed 's/.$//' | tr -d '.') && python3.10 -m pip install --no-cache-dir \
-    torch==2.0.1+cu${CUDA_VER} \
-    torchvision==0.15.2+cu${CUDA_VER} \
+    torch==2.1.2+cu${CUDA_VER} \
+    torchvision==0.16.2+cu${CUDA_VER} \
         --extra-index-url https://download.pytorch.org/whl/cu${CUDA_VER}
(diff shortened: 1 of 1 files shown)


jb-ye (approved):
Thx

Dockerfile
```diff
@@ -127,18 +127,18 @@ WORKDIR /home/${USERNAME}
...
-    torch==2.0.1+cu${CUDA_VER} \
-    torchvision==0.15.2+cu${CUDA_VER} \
+    torch==2.1.2+cu${CUDA_VER} \
+    torchvision==0.16.2+cu${CUDA_VER} \
         --extra-index-url https://download.pytorch.org/whl/cu${CUDA_VER}
 
 # Install tiny-cuda-nn (we need to set the target architectures as environment variable first).
 ENV TCNN_CUDA_ARCHITECTURES=${CUDA_ARCHITECTURES}
-RUN python3.10 -m pip install --no-cache-dir git+https://github.com/NVlabs/tiny-cuda-nn.git@v1.6#subdirectory=bindings/torch
+RUN python3.10 -m pip install --no-cache-dir git+https://github.com/NVlabs/tiny-cuda-nn.git#subdirectory=bindings/torch
```
pwais, line 141: why remove this version pin?  instead of updating it to whatever commit for this PR?  tiny-cuda-nn is fairly volatile so removing the version pin makes this Dockerfile more like a cookbook than a reproducible spec

Zunhammer: As written in the PR description: to my experience the latest version (1.6) is not compatible with the requested pytorch version. I agree that pinned versions are always the better way to go but could not build the docker image with 1.6 and torch 2.1.2. if you have another solution feel free to share.

pwais: @Zunhammer you can pip-install at a specific commit like this: https://stackoverflow.com/a/13754517

or in a Dockerfile, you can also do arbitrary commands like just clone / checkout the repo at a commit and then `pip install .`

PR #3271, opened 2024-06-28
Fix different initial tension value between `tension_slider` and `cam…

yzslab:
Fix different initial tension value between `tension_slider` and `camera_path.tension`


jb-ye (approved):
Thx

PR #3200, opened 2024-06-08
Update splatfacto with gsplat 1.0

liruilong940607:
- Pin dependence to `gsplat==1.0.0`
- A few slight improvements on splatfacto model and its data loading, by @kerrj 

This is a hot fix as the `pip install nerfstudio` would be failing currently because of #3197  so we need to bump a version that either pins to `gsplat==1.0.0` or pins to `gsplat <=1.0.0` asap.

PR #3283, opened 2024-07-02
Update dockerfile and add docker build action

jkulhanek:
This PR updates current Dockerfile:
- It builds gsplat
- Reduces image size by using multistage build and removing dev dependencies.
- Removes changed user -> enables installing inside docker + simplifies apptainer uses.

It also adds a github action to build the docker image and push it to associated NerfStudio ghcr packages repository. (We need to enable this first, however).


nerfstudio/scripts/train.py
jkulhanek, line 1: Sorry to include this line here: but is makes sense anyway (it causes issues when using gpu from dataloader in DDP setup). Should I drop it?

Dockerfile
```diff
@@ -113,6 +113,7 @@ COPY --from=builder /usr/local/lib/python3.10/dist-packages/ /usr/local/lib/pyth
 COPY --from=builder /usr/local/bin/ns* /usr/local/bin/
 
 # Install nerfstudio cli auto completion
+WORKDIR /workspace
```
ginazhouhuiwu, line 116: Small change, otherwise I had some directory issue running the next command.

nerfstudio/scripts/train.py
ginazhouhuiwu: Makes sense to me!

Dockerfile
jkulhanek: I don't think this should be needed...

jkulhanek: Shouldn't be needed with the last version.

Dockerfile
```diff
@@ -1,185 +1,125 @@
...
+    ninja install -j1 && \
+    cd ~
+
+# Upgrade pip and install dependencies.
+# pip install torch==2.2.2 torchvision==0.17.2 --index-url https://download.pytorch.org/whl/cu118 && \
+RUN pip install --no-cache-dir --upgrade pip 'setuptools<70.0.0' && \
+    pip install --no-cache-dir torch==2.1.2+cu118 torchvision==0.16.2+cu118 'numpy<2.0.0' --extra-index-url https://download.pytorch.org/whl/cu118 && \
+    git clone --branch master --recursive https://github.com/cvg/Hierarchical-Localization.git /opt/hloc && \
+    cd /opt/hloc && git checkout v1.4 && python3.10 -m pip install --no-cache-dir . && cd ~ && \
+    TCNN_CUDA_ARCHITECTURES="${CUDA_ARCHITECTURES}" pip install --no-cache-dir "git+https://github.com/NVlabs/tiny-cuda-nn.git#subdirectory=bindings/torch" && \
```
pwais, line 66: would be nice to restore a version / commit pin for tiny-cuda-nn because that project is pretty volatile / research code

pwais: same for gsplat install; that repo is extra volatile!

pwais:
NB: this change basically gets rid of the user and security / "OCI Standards" changes introduced in https://github.com/nerfstudio-project/nerfstudio/pull/2961

The user also needs a more modern version of docker to use multi-stage builds / `syntax=docker/dockerfile:1` but that's probably less of an issue today than a year ago.  You might still put `# syntax=docker/dockerfile:1` at the top.

Not that I'm advocating for restoring the cited changes / user account stuff, but as a nerfstudio user it's really confusing when PRs allowed to trample on each other 🤷‍♂️

Dockerfile
jkulhanek: Sounds good.

jkulhanek:
I actually think changing user is not a good practice. We would require uses to build their own images with their own user id? Seems like this is not ideal for releasing a docker image. Also, I though using sudo is a bad practice for docker nowadays. Also, while #2961 works for podman, it breaks apptainer which I think is more common, and the current solution could also work rootless in podman (just running image, not sure about the compilation).

nerfstudio/scripts/train.py
```diff
@@ -95,6 +95,8 @@ def train_loop(local_rank: int, world_size: int, config: TrainerConfig, global_r
         config: config file specifying training regimen
     """
     _set_random_seed(config.machine.seed + global_rank)
+    if torch.cuda.is_available():
+        torch.cuda.set_device(local_rank)
```
brentyi, line 99: @jkulhanek is this necessary? if so I added the `if` statement to prevent the CI from complaining

brentyi (approved):
(aside from one comment I tried this and it looks good to me!)

nerfstudio/scripts/train.py
brentyi: `torch.cuda.set_device()` is actually new to me, so I was just looking it up in the [PyTorch docs](https://pytorch.org/docs/stable/generated/torch.cuda.set_device.html) and saw:
> Usage of this function is discouraged in favor of [device](https://pytorch.org/docs/stable/generated/torch.cuda.device.html#torch.cuda.device). In most cases it’s better to use CUDA_VISIBLE_DEVICES environmental variable.

If you think it's a good change despite being unrelated to this PR I'm fine with leaving it, I just wasn't sure why it was needed, if it breaks things for people using `CUDA_VISIBLE_DEVICES`, etc.

jkulhanek: I removed set_device as it is unrelated. We can add it once the need arises I guess.

brentyi: Thanks! In that case the PR looks good to me.

brentyi:
In addition testing the actions, perhaps one last thing is updating the docs:

https://docs.nerf.studio/quickstart/installation.html#use-docker-image
https://github.com/nerfstudio-project/nerfstudio/blob/main/docs/quickstart/installation.md#use-docker-image

brentyi:
Currently we have: `docker pull ghcr.io/nerfstudio-project/nerfstudio:1.1.3`

What would that look like if we built with every commit?

jkulhanek:
We would still have `docker pull ghcr.io/nerfstudio-project/nerfstudio:1.1.3` for releases and `docker pull ghcr.io/nerfstudio-project/nerfstudio:latest` or `docker pull ghcr.io/nerfstudio-project/nerfstudio` for current main.

brentyi:
In that case doing on every commit makes sense to me! If there are Docker-related problems it'd also help us catch them earlier. (as opposed to only testing the Docker action when we do a release)

jkulhanek:
Seems like the gsplat build is running out of CPU/mem...
```
The hosted runner: GitHub Actions 20 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
```
Does anyone have an idea? Perhaps limit the number of CPU cores for the gsplat build?

brentyi:
@jkulhanek it seems like setting `MAX_JOBS` helped, does that seem reasonable?

Dockerfile
```diff
@@ -1,185 +1,132 @@
...
+# Install gsplat and nerfstudio.
+# NOTE: both are installed jointly in order to prevent docker cache with latest
+# gsplat version (we do not expliticly specify the commit hash).
+#
+# We set MAX_JOBS to reduce resource usage for GH actions:
+# - https://github.com/nerfstudio-project/gsplat/blob/db444b904976d6e01e79b736dd89a1070b0ee1d0/setup.py#L13-L23
+COPY --from=source /tmp/nerfstudio/ /tmp/nerfstudio
+RUN export TORCH_CUDA_ARCH_LIST="$(echo "$CUDA_ARCHITECTURES" | tr ';' '\n' | awk '$0 > 70 {print substr($0,1,1)"."substr($0,2)}' | tr '\n' ' ' | sed 's/ $//')" && \
+    export MAX_JOBS=4 && \
+    pip install --no-cache-dir git+https://github.com/nerfstudio-project/gsplat.git && \
```
msusag, line 79:
```suggestion
    pip install --no-cache-dir git+https://github.com/nerfstudio-project/gsplat.git@v1.0.0 && \
```

msusag: Hi, 
I added the above suggestion.
The latest main branch version of gsplat seems to have some problems with nerfstudio splatfacto.

To fix it, use a pinned version of gsplat, which is known to work. 
v1.0.0 is the version listed in the pyproject.toml file, so suggest using that.

Alternatively, the gsplat tag/branch could be stored in an ARG, as GSPLAT_VERSION. This  would enable it to be overridden during docker build.

jkulhanek:
> @jkulhanek it seems like setting `MAX_JOBS` helped, does that seem reasonable?
Sure, looks ok to me.

Dockerfile
jkulhanek: I guess we should allow gsplat version to be specified as arg, the question is what default to choose. I'm a bit afraid people forget updating this dockerfile and we want to use recent versions of gsplat. I was thinking of installing last released version, but not sure. What do you think?

jkulhanek:
I've changed the settings on nerfstudio docker package to inherit the organization permissions and I've locked myself out of it. @brentyi, can you please re-add me as the admin of the package? The build was failing on write permissions which I hope to have fixed before locking myself out.

Dockerfile
msusag: Yes, an ARG would be good, so people can choose to override in their own docker builds.

IMO, the default gsplat version in the docker image should be a stable version of gsplat, which follows whatever gsplat version the pip install nerfstudio versions use.  The version specified in pyproject.toml is 1.0.0.

See the reason the gsplat version was pinned in the pyproject.toml in 
https://github.com/nerfstudio-project/nerfstudio/pull/3200

brentyi:
> @brentyi, can you please re-add me as the admin of the package?

Should be done!

Dockerfile
brentyi: Is it possible to just take the pinned version from the `pyproject.toml`?

jkulhanek:
Which version of gsplat do we want to build with? Last main, last released or some fixed version (1.0.0) to match pip install?

Dockerfile
msusag: Maybe just drop the `pip install ... gsplat.git` command, and only use the next command `pip install /tmp/nerfstudio `? That should make nerfstudio use the gsplat 1.0.0 as specified in pyproject.toml. This should be the preferred, stable version.

Nerfstudio splatfacto presumably needs to be updated to be compatible with the latest gsplat version at any point in time. But contributors working on splatfacto should comment on that.

brentyi: ^this makes sense to me!

msusag: After some testing without the pip install gsplat command, it looks like a gsplat build of some version is needed after all. 
The gsplat wheel needs to be produced to be included in the docker image.
Without the prebuilt wheel, the docker image will contain just the source files. 
And as the final docker image will be without builds tooling such as nvcc, the wheel can not be built at runtime / JIT, which makes splatfacto runs fail with "No CUDA toolkit found. gsplat will be disabled" from https://github.com/nerfstudio-project/gsplat/blob/main/gsplat/cuda/_backend.py#L141.

msusag: Also, with the current command sequence, a pinned gsplat v1.0.0 build of is needed, as the "pip install nerfstudio" will use the gsplat 1.0.0, and would uninstall any newer or older versions (including the wheel), and replace with a non-compiled version of 1.0.0.

So, my original suggestion "pip install --no-cache-dir git+https://github.com/nerfstudio-project/gsplat.git@v1.0.0" is the only way I have been able to successfully run splatfacto in the docker image.

thawn:
@jkulhanek thanks for your excellent work. I pulled docker://ghcr.io/nerfstudio-project/nerfstudio:pr-3283 with singularity and it works well 👍

Your work towards reducing image size seems to have payed off. The image is just 5GB instead of 8 - 11GB.

Dockerfile
brentyi: (1)

For just relying on the nerfstudio dependencies, would it help to trigger the gsplat JIT compilation via:
```sh
python -c "from gsplat.cuda import _backend"
```
after the pip install? cc @liruilong940607

This may or may not require https://github.com/nerfstudio-project/gsplat/pull/319 to make it into a release.

(2)

Alternatively, we can do some grep magic with the nerfstudio pyproject.toml?
```
GSPLAT_VERSION=$(grep -oP 'gsplat==\K[^"]+' pyproject.toml)
pip install --no-cache-dir git+https://github.com/nerfstudio-project/gsplat.git@v$GSPLAT_VERSION
```

brentyi: We could also wait for https://github.com/nerfstudio-project/gsplat/pull/365

jkulhanek: I think option 1 is nice. I think I had something similar at one point and later switched to latest build (I thing somebody wanted it but I don’t remember).

jkulhanek:
If the build passes it should be ready to be merged. I've updated the docs.

thawn:
@jkulhanek the image attestation failed. Any idea why?

jkulhanek:
Looks good. Merging.

brentyi:
Thanks for your efforts here @jkulhanek!!

PR #3293, opened 2024-07-08
Pass current camera idx in metadata to get output from viser

KevinXu02:
Allows viser to pass the last clicked camera idx to the get_camera_output function.


sion3951:
This does fix an issue with rendering when using splatfacto-w

kerrj (approved):
lgtm!

PR #3296, opened 2024-07-09
fix color shape when sh_degree=0 in splatfacto

liruilong940607:
fix https://github.com/nerfstudio-project/gsplat/issues/268


jb-ye (approved):
Thx!

PR #3297, opened 2024-07-09
Upgraded resize_image function to support interpolation

### LINKED ISSUE #3263: Question Regarding Image Downscaling Method
https://github.com/nerfstudio-project/nerfstudio/blob/9b3cbc79bf239eb3c69e7c288632aab02c4f0bb1/nerfstudio/models/splatfacto.py#L83

Why was the following method chosen for downscaling images instead of directly using [F.resize](https://pytorch.org/vision/main/generated/torchvision.transforms.functional.resize.html)?

```python
def resize_image(image: torch.Tensor, d: int):
    """
    Downscale images using the same 'area' method in opencv

    :param image shape [H, W, C]
    :param d downscale factor (must be 2, 4, 8, etc.)

    return downscaled image in shape [H//d, W//d, C]
    """
    import torch.nn.functional as tf

    image = image.to(torch.float32)
    weight = (1.0 / (d * d)) * torch.ones((1, 1, d, d), dtype=torch.float32, device=image.device)
    return tf.conv2d(image.permute(2, 0, 1)[:, None, ...], weight, stride=d).squeeze(1).permute(1, 2, 0)
```

My concern is that this method may lead to misaligned coordinates. For instance, if we input an image of size 19x19 and downscale it by a factor of 4, the last 3 pixels would be left empty, whereas ideally, these 3 pixels should be evenly distributed in one row.

https://github.com/nerfstudio-project/nerfstudio/blob/9b3cbc79bf239eb3c69e7c288632aab02c4f0bb1/nerfstudio/data/dataparsers/colmap_dataparser.py#L460

Additionally, I noticed in another part of the code, linear interpolation (FFMPEG default) is used for image downsampling. Therefore, for code consistency, I believe the same interpolation method should be used during dataset preprocessing and training phase downsampling.

JTStephens18:
Replaced convolution operation with interpolation. Added support for downscaling using floats.

resolves #3263

### DIFF
--- a/nerfstudio/models/splatfacto.py
+++ b/nerfstudio/models/splatfacto.py
@@ -80,20 +80,22 @@ def SH2RGB(sh):
     return sh * C0 + 0.5
 
 
-def resize_image(image: torch.Tensor, d: int):
+def resize_image(image: torch.Tensor, d: float):
     """
-    Downscale images using the same 'area' method in opencv
+    Downscale images using PyTorch's interpolate function with 'area' mode
 
     :param image shape [H, W, C]
-    :param d downscale factor (must be 2, 4, 8, etc.)
+    :param d downscale factor (Ex: 2.0, 3.6, 4.0, 5.1, etc.)
 
     return downscaled image in shape [H//d, W//d, C]
     """
     import torch.nn.functional as tf
 
-    image = image.to(torch.float32)
-    weight = (1.0 / (d * d)) * torch.ones((1, 1, d, d), dtype=torch.float32, device=image.device)
-    return tf.conv2d(image.permute(2, 0, 1)[:, None, ...], weight, stride=d).squeeze(1).permute(1, 2, 0)
+    # tf.interpolate needs input of shape [batch, C, H, W]
+    image = image.to(torch.float32).unsqueeze(0).permute(0, 3, 1, 2)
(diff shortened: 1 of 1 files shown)


jb-ye:
I am not in favor of using aliased downscaling method for training GS. See my response here: https://github.com/nerfstudio-project/nerfstudio/issues/3263#issuecomment-2223530173

brentyi:
Similarly, can we just set `antialias=True` in `torch.nn.functional.interpolate()`?

jb-ye:
> One option is to use [torchvision's](https://pytorch.org/vision/main/generated/torchvision.transforms.Resize.html#resize) resize which has antialiasing?

I tried this before, it still doesn't bring the same PSNR results as convolution (that's why I eventually change it to convolution). I am a little unsure if those antialiased implementation in pytorch is correct at all. For example, as noted by [this paper](https://openaccess.thecvf.com/content/CVPR2022/papers/Parmar_On_Aliased_Resizing_and_Surprising_Subtleties_in_GAN_Evaluation_CVPR_2022_paper.pdf), most existing downscaling implementation doesn't implement antialiasing properly.

jb-ye:
> Similarly, can we just set `antialias=True` in `torch.nn.functional.interpolate()`?

I am curious to see results on some on MipNerf360 datasets and see whether it degrades the performance by checking the the PNSR eval curve. The current choice was made based on my observation of bicycle and garden dataset.

PR #3305, opened 2024-07-14
Added _load_3D_points to scannet_dataparser.py

junyuan-fang:
Reading the .ply file on scannet_dataparser.py is possible. Users can read .ply file's point cloud positions or point cloud positions+colors.


jb-ye (approved):
Thanks!

PR #3310, opened 2024-07-16
Loose assert condition for datamanger in get_average_eval_image_metrics()

KevinXu02:
Loose assert condition for datamanger in get_average_eval_image_metrics(). This enables external methods with modified datamangers to use VanillaPipeline.


jb-ye (approved):
LGTM

PR #3312, opened 2024-07-17
Addressed TODO in full_images_datamanager.py to return image resolution as train rays

ajean757:
I noticed a todo when looking at the code base and attempted to make the appropriate changes. I'm essentially returning the resolution of the image, under the assumption that all images in the batch are the same resolution. I left in the default return value of 800 * 800 in case anything doesn't work as intended. To my understanding, splatfacto is the only method that uses the `FullImageDataManager`, and the function `get_train_rays_per_batch()` is used for logging metrics to the terminal. This should help display a more accurate count for the "Train Rays / Sec" metric.


kerrj (changes requested):
LGTM, but change "image being rendered" to "image returned from datamanager" (since this won't always be exactly rendered)

ajean757:
sounds good, thanks!

kerrj (approved):

PR #3314, opened 2024-07-18
[DRAFT] Automatic circular camera path rendering

ginazhouhuiwu:
Interface to generate a circular camera path! Object avoidance is WIP, but a vanilla circular path for now can be useful for object comparisons. 

https://github.com/user-attachments/assets/40aa0c37-eda3-4aa6-b4c7-43d48d9637e3

TODO:
- [ ] Fix the clicking. You can see that the clicking depth is kind of wonky. maybe a draggable sphere to indicate the center could be nice?
- [ ] Add custom origin point input option so users can recreate all the params for a generate path easily

### DIFF
--- a/nerfstudio/viewer/render_panel.py
+++ b/nerfstudio/viewer/render_panel.py
@@ -753,15 +753,15 @@ def populate_render_tab(
                 origin = torch.tensor(event.ray_origin).view(1, 3)
                 direction = torch.tensor(event.ray_direction).view(1, 3)
 
-                # get intersection
+                # Get intersection
                 bundle = RayBundle(
-                    origin,
-                    direction,
-                    torch.tensor(0.001).view(1, 1),
+                    origins=origin,
+                    directions=direction,
+                    pixel_area=torch.tensor(0.001).view(1, 1),
+                    camera_indices=torch.tensor(0).view(1, 1),
                     nears=torch.tensor(0.05).view(1, 1),
                     fars=torch.tensor(100).view(1, 1),
-                    camera_indices=torch.tensor(0).view(1, 1),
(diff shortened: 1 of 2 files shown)


nerfstudio/viewer/render_panel.py
```diff
@@ -726,6 +729,144 @@ def _(_: viser.GuiEvent) -> None:
...
+                bundle = RayBundle(
+                    origins=origin,
+                    directions=direction,
+                    pixel_area=torch.tensor(0.001).view(1, 1),
+                    camera_indices=torch.tensor(0).view(1, 1),
+                    nears=torch.tensor(0.05).view(1, 1),
+                    fars=torch.tensor(100).view(1, 1),
+                ).to("cuda")
+
+                # Get the distance/depth to the intersection --> calculate 3D position of the click
```
kerrj, line 767: This is a good start! but this part should use the more general get_outputs_for_camera() function (a lot of the sampling, scaling stuff is method-specific to nerfacto). As long as the method outputs a 'depth' value it should work with this ray deprojection approach

kerrj: This way, you would render a depth image from the viewer camera and deproject the click point with the intrinsics matrix + rendered depth, meaning it doesn't matter what the rendering backend is.

kerrj: One thing to note is that the 'depth' in splatfacto is actually z-depth and not ray-depth, so the math would need to be different for the two methods

ginazhouhuiwu: Ok that makes sense! Thanks for the feedback appreciate it!!

PR #3333, opened 2024-07-25
installation.md: Remove manual gsplat installation for Windows

SharkWipf:
I'm reluctant to make changes to the extremely fragile Windows installation instructions as I don't use those instructions myself, but there's been several reports on Discord that this manual installation of gsplat breaks the installation process now that gsplat 1.1.1 is out and not yet compatible with Nerfstudio.

I don't know why gsplat was being installed manually on Windows in the first place since Nerfstudio pulls the right version in itself perfectly fine, including on Windows.

As a sidenote, I have found WSL2 to be much more reliable and consistent for Windows installations (since, once installed, it's just Linux and the Linux instructions apply) and [wrote a guide for it](https://gist.github.com/SharkWipf/0a3fc1be3ea88b0c9640db6ce15b44b9) a while ago.
Would it be worthwhile including WSL2 instructions as an alternative install method in the official docs?
It does mean a whole different process to setting it up though, so it would probably be better off split off to a separate file if so.


KevinXu02:
A problem is that the pip installed version(nerfstudio pulled one) will use JIT compile when training for the fisrt time. However, for windows system you have to set up the VC env beforehand or the compile will 100% fail. So gsplat need to be installed from source with the right VC env which makes things easier.
I think a better fix might be fix gsplat to an older version (? and I agree the Windows installation instructions is really fragile.
I use nerfstudio on both wsl and windows and sometimes bug occurs when using vscode+wsl2+Viser (also some performace loss). So I would prefer using them on Windows in my case(

SharkWipf:
After some discussion on Discord (included below), I've modified the changes to hopefully be more failsafe.
I took some liberties with the format though, and have not been able to test if it works in the generated docs, so if any changes are desired please let me know.

I do intend to write a proper upstream-includeable WSL2 guide that can be included later, but as of ~yesterday Pixi suddenly broke on WSL2, I still have to investigate that, so for now I linked my existing gist guide with a warning.

```
KevinXu — Today at 10:59 AM
A problem is that the pip installed version(nerfstudio pulled one) will use JIT compile when training for the fisrt time. However, for windows system you have to set up the VC env beforehand or the compile will 100% fail. So gsplat need to be installed from source with the right VC env which makes things easier.
Indeed setting up nerfstudio is really hard on Windows systems and the version reliance on gsplat would break things.🥲

[...]

SharkWipf — Today at 1:13 PM
Hmm, right. But don't you just need to run the vcvarsall step first before running splatfacto the first time? Even with automatically pulled gsplat, does that not fix it?

KevinXu — Today at 1:49 PM
Maybe yes?(I haven't tested it yet. For me it's a little bit awkward if I need to find useable ns-data and run train on it when setting up the ns env for the first time (which is likely to fail several times in the middle) 
Anyway, I think a warning for windows installation is necessary

SharkWipf — Today at 1:55 PM
I feel like directly specifying a gsplat version in the docs should be avoided if possible, for 2 reasons:
The user may want a specific nerfstudio version (the latest doesn't support 10-series GPUs and earlier anymore)
The docs won't be kept in sync until people start complaining

So if the vcvarsall thing works, rewording the docs to make that more clear should be preferred I think

KevinXu — Today at 1:57 PM
Make sence. I think we should fix that this way.
```

kerrj:
Thanks a lot for the doc update! Unfortunately I don't have access to a windows machine so I can't test it but if you and @KevinXu02 are happy with the changes now I can merge this. To test the generated docs you can use [this](https://docs.nerf.studio/reference/contributing.html#building), then VSCode has a [Live Server](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer) extension that you can open the index.html file in. If you can't test it I can but it may take me some time to get around to it.

SharkWipf:
Noted, thanks. I'll see if I can get that working later this week.  
I should note I haven't been able to directly tested this myself yet either, but I am _pretty sure_ that this is how it works, it seems to have worked for others, and if nothing else it's not worse than the current version, which is just broken atm.

SharkWipf:
Took a little longer than anticipated due to me getting sick, but the docs seem fine when built, no issues there.
I made a minor change in wording in the troubleshooting section, everything looks good (enough) to me now, at least until a proper full rewrite/cleanup gets done.

jb-ye (approved):
Thank you!

PR #3347, opened 2024-08-05
Lazy import for open3d in Blender DataParser

ethanweber:
I'm following this reference code to improve the speed of importing the Blender DataParser. We should only import Open3D when we need it.

https://github.com/nerfstudio-project/nerfstudio/blob/b0dd3d473094fb0942b357ef436b362df1ef77cd/nerfstudio/data/dataparsers/scannet_dataparser.py#L207


jb-ye (approved):
Lgtm

PR #3363, opened 2024-08-13
draft: add a way to visualize scene box, seed points, and modify if wanted

hoanhle:
image


nerfstudio/viewer/viewer.py
```diff
@@ -282,8 +289,149 @@ def nested_folder_install(folder_labels: List[str], prev_labels: List[str], elem
...
+
+        new_aabb = torch.tensor(
+            [min_coords / VISER_NERFSTUDIO_SCALE_RATIO, max_coords / VISER_NERFSTUDIO_SCALE_RATIO],
+            device=self.pipeline.model.device,
+            requires_grad=False,
+        )
+
+        self.pipeline.model.scene_box = SceneBox(aabb=new_aabb)
+        self.pipeline.model.render_aabb = SceneBox(aabb=new_aabb)
+        self.pipeline.model.populate_modules()
```
hoanhle, line 434: Using `device=self.pipeline.model.device` for `new_aabb` seems to cause issues where tensors end up on different devices later in the process. I'm not entirely sure what the best solution is to address this.

@brentyi @tancik, since you both work with this part of the codebase frequently, do you have any suggestions?

nerfstudio/models/base_model.py
```diff
@@ -81,16 +83,11 @@ def __init__(
...
+        self.seed_points = seed_points
         self.callbacks = None
-        # to keep track of which device the nn.Module is on
-        self.device_indicator_param = nn.Parameter(torch.empty(0))
 
-    @property
-    def device(self):
-        """Returns the device that the model is on."""
-        return self.device_indicator_param.device
+        self.populate_modules()  # populate the modules
```
hoanhle, line 91: get the `device` from `pipeline`.

PR #3364, opened 2024-08-14
add a way to orient based on main direction of the flat surface

hoanhle:


jb-ye:
@hoanhle Could you share a sample visualization of this new orientation method?

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -399,10 +388,45 @@ def _generate_dataparser_outputs(self, split="train"):
 
             if ply_file_path:
                 sparse_points = self._load_3D_points(ply_file_path, transform_matrix, scale_factor)
```
jb-ye, line 390: You need to annotate the output type of function `_load_3D_points", otherwise the pyright would complains.

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -458,6 +482,70 @@ def _load_3D_points(self, ply_file_path: Path, transform_matrix: torch.Tensor, s
         }
         return out
 
+    @staticmethod
+    def _align_points_to_target_plane(points: torch.Tensor, target_normal: torch.Tensor):
```
jb-ye, line 486: Annotate the output type of this function.

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -458,6 +482,70 @@ def _load_3D_points(self, ply_file_path: Path, transform_matrix: torch.Tensor, s
...
+        _, _, vh = torch.linalg.svd(centered_points)  # vh shape: (3, 3)
+
+        # The last right singular vector is the normal to the plane
+        normal = vh[-1]  # Shape: (3,)
+
+        # Calculate the rotation axis and angle
+        rotation_axis = torch.cross(normal, target_normal)  # Shape: (3,)
+        rotation_axis_norm = torch.norm(rotation_axis)
+
+        if rotation_axis_norm != 0:
```
jb-ye, line 516: I would assume there would be some numerical instability when rotation_axis_norm is very close to zero? Does it make sense to use a more tolerable condition such as `rotation_axis_norm < 1e-6` or so?

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -458,6 +482,70 @@ def _load_3D_points(self, ply_file_path: Path, transform_matrix: torch.Tensor, s
...
+        centered_points = points_xyz - centroid  # Shape: (n, 3)
+
+        # Perform SVD
+        _, _, vh = torch.linalg.svd(centered_points)  # vh shape: (3, 3)
+
+        # The last right singular vector is the normal to the plane
+        normal = vh[-1]  # Shape: (3,)
+
+        # Calculate the rotation axis and angle
+        rotation_axis = torch.cross(normal, target_normal)  # Shape: (3,)
```
jb-ye, line 513: To compute the rotation between two vectors, we have a similar helper function at https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/cameras/camera_utils.py#L449 
Could you reuse this function instead writing your own?

jb-ye (changes requested):

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
hoanhle: done

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
hoanhle: done

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
hoanhle: done

nerfstudio/data/dataparsers/nerfstudio_dataparser.py
```diff
@@ -258,6 +270,7 @@ def _generate_dataparser_outputs(self, split="train"):
 
         # in x,y,z order
         # assumes that the scene is centered at the origin
+        # _ = self.config.scene_scale
```
jb-ye, line 273: delete this comment?

jb-ye (commented):
I feel major refactor of current nerfstudio parser are needed to achieve a clean separation between orientation logics from data parsing. 

(1) Currently orientation and centering logics is simply go through helper function: `camera_utils.auto_orient_and_center_poses`. This is important, because the same helper function is used in multiple parsers (e.g. colmap parser, etc). The current PR only applies to one specific parser and it makes maintaining these logics much more difficult. 

(2) Because the new `align` method depends on the point cloud loaded from PLY, we need to first load point cloud and then compute orientation transforms. This require some refactor of existing codes.

hoanhle:
> I feel major refactor of current nerfstudio parser are needed to achieve a clean separation between orientation logics from data parsing.
> 
> (1) Currently orientation and centering logics is simply go through helper function: `camera_utils.auto_orient_and_center_poses`. This is important, because the same helper function is used in multiple parsers (e.g. colmap parser, etc). The current PR only applies to one specific parser and it makes maintaining these logics much more difficult.
> 
> (2) Because the new `align` method depends on the point cloud loaded from PLY, we need to first load point cloud (if they exist) and then compute orientation transforms. This require some refactor of existing codes.
> 
> As such, I feel the current PR is not ready to merge until (1) and (2) are addressed.

@jb-ye This makes sense to me. I will proceed with the refactoring unless you know of a simpler way to achieve the same result.

jb-ye:
@hoanhle That would be great, I'd love to help review your change once they are ready.

PR #3368, opened 2024-08-16
More docs on PDFSampler

MashiPe:
Hey, I notice that the docs may be extended to explain with more detail certain topics. This is the my take on the PDFSampler explaining some of the theory behind the sampler.


MashiPe:
Hey, just wanna  check if the changes are alright or if they might be out of scope. No rush, of course—I just want to make sure I’m not missing anything.

PR #3372, opened 2024-08-19
Update exporter.py for sh_degree 0

bell-one:
Change to write sh coefficients instead of color values

When exporting weights trained with sh degree 0, 
current implementation exports color values, which cannot be rendered in sh degree 0 renderes such as [WebGL Viewer by antimatter15](https://antimatter15.com/splat/).

To address this issue, I've modified the code to output sh coefficients when sh degree is 0.
I've added a specific case for sh degree 0 to resolve this problem.

Current in WebGL sh degree 0 renderer
image

Edited in WebGL sh degree 0 renderer
image

------------------------------------------------------
In cases where sh degree is changed to >= 0, accessing model.shs_0 directly in models trained with sh degree 0 can result in infinite values.

### DIFF
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -637,6 +637,12 @@ method_configs["splatfacto"] = TrainerConfig(
                 lr_final=5e-7, max_steps=30000, warmup_steps=1000, lr_pre_warmup=0
             ),
         },
+        "bilateral_grid": {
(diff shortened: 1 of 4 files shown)


jb-ye:
Could you make this feature as an option in exporter and disabled by default? I think the current behavior which output uint8 colors directly is intended. The change you made is more like a work around feature to support third party tools.

PR #3375, opened 2024-08-22
Fix memory leak in full_images_datamanager.py

### LINKED ISSUE #3335: Memory accumulation on the GPU - and a Fix!
I experienced OOM with very big datasets and realised the images were accumulating on the GPU. After a lot of digging I found that here:

https://github.com/nerfstudio-project/nerfstudio/blob/702886b213a5942b64250b3c4dc632e62237a511/nerfstudio/data/datamanagers/full_images_datamanager.py#L347

Because a deepcopy of the original tensor is not made, the cpu_cached images are transferred to the GPU and left there taking up memory quickly. Modifying the data dictionary does modify the underlying cached_train.

Interestingly a deepcopy is performed here: https://github.com/nerfstudio-project/nerfstudio/blob/702886b213a5942b64250b3c4dc632e62237a511/nerfstudio/data/datamanagers/full_images_datamanager.py#L373

I would argue that we don't want the dataset to accumulate on the GPU and a deepcopy should be added.

brentyi:
I think this should fix #3335; cc @kerrj @jb-ye @AntonioMacaronio

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -350,7 +350,7 @@ class FullImageDatamanager(DataManager, Generic[TDataset]):
             self.train_unseen_cameras = self.sample_train_cameras()
 
         data = self.cached_train[image_idx]
-        # We're going to copy to make sure we don't mutate the cached dictionary. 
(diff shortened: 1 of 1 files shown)


jb-ye:
@brentyi I made some follow-up changes, what do you think?

brentyi:
Looks great, thanks @jb-ye!

PR #3381, opened 2024-08-23
Fix #3355

### LINKED ISSUE #3355: The K4 intrinsic parameter in nerfstudio doesn't match OpenCV or COLMAP
**Describe the bug**

OpenCV [(source)](https://github.com/opencv/opencv/blob/da3debda6d233af90e421e95700c63fc08b83b75/modules/calib3d/src/calibration.cpp#L906) and COLMAP [(source)](https://github.com/colmap/colmap/blob/66fd8e56a0d160d68af2f29e9ac6941d442d2322/src/colmap/sensor/models.h#L1117) both use the rational fraction radial distortion model described [In the OpenCV doc](https://docs.opencv.org/4.x/d9/d0c/group__calib3d.html).

nerfstudio [(source)](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/cameras/camera_utils.py#L376) uses a fourth order radial distortion polynomial, which has a totally different meaning for parameter k4. This seems to be inherited from other NeRF code.

Consequently, any project that uses a camera model with a non-zero k4 (which probably doesn't happen often) will be wrongly processed.

Several options:
- stop supporting k4 in nerfstudio
- stop supporting loading k4 in the colmap dataparser, and throw an error on non-zero k4, k5 or k6
- support COLMAP and OpenCV's k4 in nerfstudio, and throw an error on non-zero k5 or k6 in the COLMAP dataparser

jb-ye:
Hotfix, add safeguards for parsing colmap data.

### DIFF
--- a/nerfstudio/data/dataparsers/colmap_dataparser.py
+++ b/nerfstudio/data/dataparsers/colmap_dataparser.py
@@ -276,6 +276,11 @@ class ColmapDataParser(DataParser):
             cy.append(float(frame["cy"]))
             height.append(int(frame["h"]))
             width.append(int(frame["w"]))
+            if "k4" in frame and float(frame["k4"]) != 0.0:
+                raise ValueError(
+                    "K4 is non-zero! Please note that Nerfstudio camera model's K4 has different meaning than colmap "
+                    "OPENCV camera model K4."
+                )
             distort.append(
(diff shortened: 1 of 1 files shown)


kerrj (approved):
lgtm!

PR #3382, opened 2024-08-24
Updating camera intrinsics (principal points) when the undistorted image is cropped

AntonioMacaronio:
**Problems and Background**

- When an image is undistorted by OpenCV's `undistort` for perspective cameras, there could be a returned Region-Of-Interest (ROI) where the image's pixels are well defined.

- Nerfstudio does crop the image according to OpenCV's returned ROI, but it does not update the camera intrinsics to reflect where the new principal point has shifted, (usually the crop is very tiny <= 1 pixel or doesn't exist at all) so this isn't that noticeable.

**Overview of Changes**
- Added some notes explaining the processing to use OpenCV
- Fixed the updated principal points in the returned intrinsic matrix

### DIFF
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -332,8 +332,13 @@ class FullImageDatamanager(DataManager, Generic[TDataset]):
         return {}
 
     def get_train_rays_per_batch(self):
-        # TODO: fix this to be the resolution of the last image rendered
-        return 800 * 800
+        """Returns resolution of the image returned from datamanager."""
+        if len(self.cached_train) != 0:
+            h = self.cached_train[0]["image"].shape[0]
+            w = self.cached_train[0]["image"].shape[1]
+            return h * w
+        else:
+            return 800 * 800
 
     def next_train(self, step: int) -> Tuple[Cameras, Dict]:
         """Returns the next training batch
--- a/nerfstudio/data/datamanagers/full_images_datamanager.py
+++ b/nerfstudio/data/datamanagers/full_images_datamanager.py
@@ -423,8 +423,8 @@ def _undistort_image(
         x, y, w, h = roi
         image = image[y : y + h, x : x + w]
         # update the principal point based on our cropped region of interest (ROI)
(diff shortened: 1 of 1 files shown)


kerrj:
Are you sure this principal point shift isn't already captured in the newK output from opencv's undistortion calculation?

brentyi (approved):
I double-checked based on @kerrj's question and it looks like the shift is needed! we probably just haven't noticed because (i) the difference is so small for most images and (ii) camera optimization partially masks that difference

nerfstudio/data/datamanagers/full_images_datamanager.py
```diff
@@ -420,6 +422,9 @@ def _undistort_image(
         # crop the image and update the intrinsics accordingly
         x, y, w, h = roi
         image = image[y : y + h, x : x + w]
+        # update the principal point based on our cropped region of interest (ROI)
+        newK[0, 1] -= x
+        newK[0, 2] -= y
```
brentyi, line 427: actually if we're adjusting cx and cy shouldn't this be `newK[0, 2]` and `newK[1, 2]`?

brentyi (changes requested):

nerfstudio/data/datamanagers/full_images_datamanager.py
AntonioMacaronio: yes this is right! ahh sorry i mistakenly flipped in the indices

jb-ye:
Maybe this issue explains https://github.com/nerfstudio-project/nerfstudio/issues/2849

jb-ye (approved):

engrmusawarali71:
@Dear @jb-ye , I am training Nerfacto on my dataset, I have problem in rendering. I train two versions

1. With distorted images while setting distortion co-effs to be zero.
2. With distorted images while setting distortion co-effs to be actuall.

In second case PSNR is better than first as it should be expected. But when try to render images using ns-render command, Nerf is hallucinating black spots in images in second case(highlighted in images below), while the in first case the images appear more realistic. I am wondering If I am doing mistake here. I would be thankful, if you help me in this regard. For your reference I am attaching images with both cases.

image

PR #3386, opened 2024-08-26
get rid of gsplat legacy dep

liruilong940607:
gsplat is gonna to deprecate `.cuda_legacy`


jb-ye (approved):
Thx

nnop:
splatfacto is still dependending on cuda_legacy:
https://github.com/nerfstudio-project/nerfstudio/blob/69ae4abbfd3c8736e684a2bd147f337b3615e446/nerfstudio/models/splatfacto.py#L33

Issue #3402, opened 2024-08-30
GLOMAP support


danie1ll:
GLOMAP, the COLMAP's successor from the same authors.

> In terms of accuracy and robustness, we achieve results on-par or superior to COLMAP, the most widely used incremental SfM, while being orders of magnitude faster.

Project page: https://lpanaf.github.io/eccv24_glomap/
Code: https://github.com/colmap/glomap

pablovela5620:
I wonder if its possible to already support this by just installing glomap via pixi/conda https://prefix.dev/channels/conda-forge/packages/glomap and modifying `colmap-command` to `glomap`. I don't know if this will outright work because it would be calling for feature extraction/matching, but you could probably just modify it https://github.com/nerfstudio-project/nerfstudio/blob/d395aa2e537244b6f54fab2082b01b0edb589e62/nerfstudio/process_data/colmap_utils.py#L156 to call glomap and everything else colmap

danie1ll:
According to e.g. this tutorial - https://www.youtube.com/watch?v=QIxXuilEEVw you should use GLOMAP only for the mapper part, and regular COLMAP for steps before mapping.

PR #3409, opened 2024-09-04
Adding glomap support to ns-process-data, requires latest version of COLMAP

AntonioMacaronio:
**Problems and Background**
- GLOMAP is orders of magnitudes faster than COLMAP On average, it finds camera poses and intrinsics 100x faster than COLMAP, and it would great to support this in Nerfstudio and already has been [requested](https://github.com/nerfstudio-project/nerfstudio/issues/3402)

**Overview of Changes**
- Adding a GLOMAP toggle in `ns-process-data` script, which uses the GLOMAP mapper

### DIFF
--- a/nerfstudio/process_data/colmap_utils.py
+++ b/nerfstudio/process_data/colmap_utils.py
@@ -99,6 +99,7 @@ def run_colmap(
     matching_method: Literal["vocab_tree", "exhaustive", "sequential"] = "vocab_tree",
     refine_intrinsics: bool = True,
     colmap_cmd: str = "colmap",
+    glomap_toggle: bool = False,
 ) -> None:
     """Runs COLMAP on the images.
 
@@ -153,12 +154,20 @@ def run_colmap(
     # Bundle adjustment
     sparse_dir = colmap_dir / "sparse"
     sparse_dir.mkdir(parents=True, exist_ok=True)
-    mapper_cmd = [
-        f"{colmap_cmd} mapper",
-        f"--database_path {colmap_dir / 'database.db'}",
-        f"--image_path {image_dir}",
-        f"--output_path {sparse_dir}",
-    ]
+    if glomap_toggle:
+        mapper_cmd = [
+            f"glomap mapper",
+            f"--database_path {colmap_dir / 'database.db'}",
+            f"--image_path {image_dir}",
+            f"--output_path {sparse_dir}",
+        ]
+    else:
+        mapper_cmd = [
+            f"{colmap_cmd} mapper",
+            f"--database_path {colmap_dir / 'database.db'}",
+            f"--image_path {image_dir}",
+            f"--output_path {sparse_dir}",
+        ]
     if colmap_version >= Version("3.7"):
         mapper_cmd.append("--Mapper.ba_global_function_tolerance=1e-6")
 
--- a/nerfstudio/process_data/colmap_converter_to_nerfstudio_dataset.py
+++ b/nerfstudio/process_data/colmap_converter_to_nerfstudio_dataset.py
@@ -222,6 +222,19 @@ class ColmapConverterToNerfstudioDataset(BaseConverterToNerfstudioDataset):
                 refine_intrinsics=self.refine_intrinsics,
                 colmap_cmd=self.colmap_cmd,
             )
+        elif sfm_tool == "glomap":
+            colmap_utils.run_colmap(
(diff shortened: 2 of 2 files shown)


kerrj:
This is great! what is the difference between the `sfm_tool` option and `glomap_toggle`? It would also be good to write documentation or link to installation instructions and the like

nerfstudio/process_data/colmap_utils.py
```diff
@@ -153,12 +154,20 @@ def run_colmap(
...
     sparse_dir = colmap_dir / "sparse"
     sparse_dir.mkdir(parents=True, exist_ok=True)
-    mapper_cmd = [
-        f"{colmap_cmd} mapper",
-        f"--database_path {colmap_dir / 'database.db'}",
-        f"--image_path {image_dir}",
-        f"--output_path {sparse_dir}",
-    ]
+    if glomap_toggle:
+        mapper_cmd = [
```
pablovela5620, line 158: Seems like this could be more consice if you did something like
```python
 mapper_cmd = [
    f"{'glomap' if glomap_toggle else colmap_cmd} mapper",
    f"--database_path {colmap_dir / 'database.db'}",
    f"--image_path {image_dir}",
    f"--output_path {sparse_dir}",
]
```

hyqshr:
Any update on this?

jb-ye:
@AntonioMacaronio Do you want to rebase your changes on top main? I think this one is close to merge.

AntonioMacaronio:
@jb-ye Sure! I'll need to do one last round of testing and I'll fix the build errors and this should be good to go

pablovela5620:
> @jb-ye update on this: there is currently a bug with how the cameras are placed in the viewer and during training. In the colmap visualizer, the camera poses are fine and look accurate. However, during training in the nerfstudio viewer all the cameras get placed in 1 spot at the origin. I'll have to look into the nerfstudio_dataparser.py to see the issue with this
@AntonioMacaronio 
I think (?) this is may be due to lack of normalization. What commit of glomap are you using? They recently add this fix in this commit
https://github.com/colmap/glomap/commit/d41b747407b6c5ce564e3936d0f2f2ea52f89dae
I had a similar problem but it got fixed this this update
https://github.com/colmap/glomap/issues/128#issuecomment-2474941872

AntonioMacaronio:
@pablovela5620 I just updated my glomap to master and tried this, but the cameras still get placed at the origin; have you had any success with this PR? Sometimes it doesn't get placed at the origin and the poses look accurate but because it's finnicky so I'm hesitant on merging

pablovela5620:
It turns out there's another issue, and you're right to be hesitant! I've been doing some debugging, and as you stated, even with the latest PR, you get the weird clustering effect. I basically ran colmap/glomap (not with this pr but with my own implementation using a different feature+matcher combo). I found that glomap can have some catastrophic failures for a small number of cameras out of the whole group. I linked a video showing this, where I'm comparing a reconstruction of 250 images using the aliked feature detector and light glue matcher with colmap and then with glomap. 

Basically one of the cameras is WAY out of bounds leading to the clustering at the origin, one or two cameras end up being 10x farther out due to a failure in point triangulation(?). I'm going to open an issue on the glomap repo

[glomap-failure-featureless.webm](https://github.com/user-attachments/assets/cf9b78de-3563-4977-ad82-f5b8fb3d1261)

AntonioMacaronio:
> Basically one of the cameras is WAY out of bounds leading to the clustering at the origin, one or two cameras end up being 10x farther out due to a failure in point triangulation(?).
I saw your comment with the video of the COLMAP gui and I see the exact same thing with my COLMAP gui. Most of the cameras are close together and accurate but one of them is very far away. What's puzzling is that Nerfstudio compacts all the close cameras to the origin. Even if one of the cameras is far away, because Nerfstudio is scaling the distances between the cameras, it should still remain proportional (in my thinking), so I'm still not sure why that happens

pablovela5620:
Yea, I'm not sure why it fails in Nerfstudios case. Might be worth looking into the normalization code that Glomap added to see how they fixed it. In either case, I think until these outlier cases get resolved its probably not good to merge this is. The massive outliers completely ruin the reconstruction into something nonsensical. Maybe there could be some heuristic value that tries to deal with outliers (like projection error or if the camera is 10x further from the center of mass or something) but those still feel very hacky

Jourdelune:
Any news?

PR #3413, opened 2024-09-05
Switch to Python 3.8 and ensure compatibility with Ubuntu 22.04

moonagic:
- Updated Dockerfile to switch the Python version to 3.8.
- Added necessary steps to handle the installation of Python 3.8 on Ubuntu 22.04, which does not support direct installation.

Related issue: #3250 #3254


lionlai1989:
Why Python 3.8?

PR #3415, opened 2024-09-05
bugfix: handle multiple camera types in a single batch

decrispell:
Training with mixed batches of camera types seems to be broken. The camera types are looped over, but the same type will trigger the if/else each time through the loop in its current form, leaving uninitialized rays for the other camera types.

This MR fixes that, enabling support for mixed batches of rays.  My guess is this was the original intent of the code.

### DIFF
--- a/nerfstudio/cameras/cameras.py
+++ b/nerfstudio/cameras/cameras.py
@@ -778,15 +778,15 @@ class Cameras(TensorDataclass):
 
             return vr180_origins, directions_stack
 
-        for cam in cam_types:
-            if CameraType.PERSPECTIVE.value in cam_types:
+        for cam_type in cam_types:
+            if CameraType.PERSPECTIVE.value == cam_type:
                 mask = (self.camera_type[true_indices] == CameraType.PERSPECTIVE.value).squeeze(-1)  # (num_rays)
                 mask = torch.stack([mask, mask, mask], dim=0)
                 directions_stack[..., 0][mask] = torch.masked_select(coord_stack[..., 0], mask).float()
                 directions_stack[..., 1][mask] = torch.masked_select(coord_stack[..., 1], mask).float()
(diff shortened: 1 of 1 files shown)


jb-ye:
The original code does seem buggy to me. @cvachha Could you take a look at this PR fix?

cvachha:
Thanks for the fix! I think this code change looks good, however I am unable to validate it working on multi-camera input and am not familiar with the initial reason why that loop was originally written that way. I am requesting another reviewer to take a look.

nerfstudio/cameras/cameras.py
```diff
@@ -778,15 +778,15 @@ def _compute_rays_for_vr180(
 
             return vr180_origins, directions_stack
 
-        for cam in cam_types:
-            if CameraType.PERSPECTIVE.value in cam_types:
```
brentyi, line 782: As an alternative fix, would it make sense to:
- Remove the `for cam in cam_types`
- De-dent all of the contents
- Change all of the `if` to `elif`?

I went through this file history and it seems like that's how these conditions were originally set up: https://github.com/nerfstudio-project/nerfstudio/blob/d413ca42042ae7b2dabdbe137c7eca7bceb42b88/nerfstudio/cameras/cameras.py#L646-L681

Which makes sense to me. The current fix also seems fine though!

decrispell: That's actually the way I originally wrote the fix, but decided to fix the loop instead since it seemed more in line with the intentions of the original author.

I don't have a strong opinion on which way is better, but I don't have a way to test anything other than fisheye and perspective cases so I figured the smaller change was safer.

brentyi: makes sense, thanks @decrispell!

brentyi (approved):

Issue #3408, opened 2024-09-04
Make frame selection from a video more intelligent


danie1ll:
From my understanding, right now calling `ns-process-data video` just randomly samples  `--num-frames-target` images from the video. This is suboptimal because of two reasons: 

1. The random strategy is super naive, maybe you could add some algorithm to select key frames based on some difference metric between images?
2. An option to fix seed for random sampling would make processing videos with nerfstudio much more reproducible

LucasArmand:
A good metric could be frame blurriness. Here's an example from Stack Overflow using the Laplacian to quantify image blurriness and select the "least blurry image" every second.
[https://stackoverflow.com/questions/65949172/how-to-extract-clear-frames-from-video-file](https://stackoverflow.com/questions/65949172/how-to-extract-clear-frames-from-video-file)

danie1ll:
@LucasArmand this looks very promising! Could also be combined with accumulated visual change between frames, so we not only select 'least blurry image' every second but 'the most important frames, which are different enough and are not blurry'

LucasArmand:
@danie1ll I wonder how far we can go in optimizing the `ns-process-data` process? When our source is a video, we usually only select a low percentage of the total frames, so there are _many_ unique combinations of frames to select for training. Some of these combinations would certainly result in higher quality trained scenes than others. Heuristics like "image blurriness" and "accumulated visual change" will probably give us a better combination of frames for training, but is it possible to find the _best_ combination?

msusag:
The scripts at https://github.com/SharkWipf/nerf_dataset_preprocessing_helper could perhaps be integrated into ns-process-data video?

Anthony-Tafoya:
No matter the solution decided for "smarter" processing, I think there is room for a fixed seed option. I am working on a PR to implement it.

PR #3416, opened 2024-09-07
Adding Random Seed for Frame Processing

Anthony-Tafoya:
This PR relates to the Issue [#3048](https://github.com/nerfstudio-project/nerfstudio/issues/3408)

Background: Currently, frame processing video logic works by sampling N frames from N evenly spaced intervals throughout the video. Although a worthy method, researchers should be allowed to choose frames from a random seed to produce more consistent results. 

This PR implements optional --random_seed and --eval_random_seed flags for the ns-process-data video in case a specific seed is requested.

Verification Evidence: Placed a video and a random seed to produce a consistent result
Here is the result

1) Showing that the same seed will process the same image frames
2) Showing the 50-frame output of the set 
3) Showing that the original frame split works successfully

[
Screen Shot 2024-09-14 at 9 32 09 PM
Screen Shot 2024-09-14 at 9 32 32 PM
Screen Shot 2024-09-14 at 9 33 35 PM
](URL)

Unit Tests: A function to test_misc that confirms that the random_seed processes the number of frames correctly. 

Future Work: After this alternative, finding ways to choose frames based on blurriness and other forms of visual change is necessary. Create an integration test with ns-process-data video since there is only one for image currently.

### DIFF
--- a/nerfstudio/process_data/process_data_utils.py
+++ b/nerfstudio/process_data/process_data_utils.py
@@ -140,6 +140,7 @@ def convert_video_to_images(
         verbose: If True, logs the output of the command.
         image_prefix: Prefix to use for the image filenames.
         keep_image_dir: If True, don't delete the output directory if it already exists.
+        random_seed: If set, the seed used to choose the frames of the video
     Returns:
         A tuple containing summary of the conversion and the number of extracted frames.
     """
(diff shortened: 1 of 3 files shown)


nerfstudio/process_data/process_data_utils.py
```diff
@@ -196,7 +196,14 @@ def convert_video_to_images(
...
         ffmpeg_cmd += " -vsync vfr"
 
-        if spacing > 1:
+        # Evenly distribute frame selection if random seed does not exist
+        spacing = num_frames // num_frames_target
+        if random_seed:
+            random.seed(random_seed)
+            frame_indices = sorted(random.sample(range(num_frames), num_frames_target))
+            select_cmd = "select='" + "+".join([f"eq(n\,{idx})" for idx in frame_indices]) + "',setpts=N/TB,"
+            CONSOLE.print(f"Extracting {num_frames_target} frames using seed-based random selection.")
```
jb-ye, line 205: Could u also print the random_seed here as well?

nerfstudio/process_data/process_data_utils.py
```diff
@@ -196,7 +196,14 @@ def convert_video_to_images(
...
+        elif spacing > 1:
             CONSOLE.print("Number of frames to extract:", math.ceil(num_frames / spacing))
```
jb-ye, line 207: Could you also make the print log more clear that the extracted frame is evenly spacing?

nerfstudio/process_data/video_to_nerfstudio_dataset.py
```diff
@@ -41,6 +41,10 @@ class VideoToNerfstudioDataset(ColmapConverterToNerfstudioDataset):
     """Feature matching method to use. Vocab tree is recommended for a balance of speed
     and accuracy. Exhaustive is slower but more accurate. Sequential is faster but
     should only be used for videos."""
+    random_seed: Optional[int] = None
+    """Random seed to select video frames"""
```
jb-ye, line 45: Is this random seed for select training set? maybe make it more clear in the comments

PR #3420, opened 2024-09-10
Add option to start training paused

### LINKED ISSUE #3358: Optional config argument to pause training at the very beginning.
**Is your feature request related to a problem? Please describe.**

When using nerf/3dgs editing methods, the appearance of the scenes can change very rapidly in the first few iterations. It is reasonable to optionally pause the optimization process at the beginning and start it after the web viewer is loaded.

**Describe the solution you'd like**

Add an optional argument to set `training_state = "paused"`.

**Describe alternatives you've considered**

Overwrite the `__init__` function of `Trainer` through monkey patching to set `training_state = "paused"`.

**Additional context**

aayushg55:
`--start_paused` option for `ns-train` to set initial training state as paused

Closes #3358

### DIFF
--- a/nerfstudio/engine/trainer.py
+++ b/nerfstudio/engine/trainer.py
@@ -86,6 +86,8 @@ class TrainerConfig(ExperimentConfig):
     """Optionally log gradients during training"""
     gradient_accumulation_steps: Dict[str, int] = field(default_factory=lambda: {})
     """Number of steps to accumulate gradients over. Contains a mapping of {param_group:num}"""
+    start_paused: bool = False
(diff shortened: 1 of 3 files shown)


jb-ye (approved):

jb-ye (commented):
Could you elaborate how you test this change locally?

aayushg55:
Starting training with something like `ns-train nerfacto --start_paused=True --data data/nerfstudio/poster`, training will not immediately start and the webviewer will have the training button in the `Resume Training` state instead of `Pause Training`. The training button can then be used to toggle training as usual. Same applies to the legacy viewer.

PR #3422, opened 2024-09-11
AABB Conflict FIx

Anthony-Tafoya:


jb-ye:
would you want to rebase your PR?

Anthony-Tafoya:
> would you want to rebase your PR?

I can rebase no worries

nerfstudio/models/instant_ngp.py
```diff
@@ -118,7 +118,7 @@ def populate_modules(self):
             self.config.render_step_size = ((self.scene_aabb[3:] - self.scene_aabb[:3]) ** 2).sum().sqrt().item() / 1000
         # Occupancy Grid.
         self.occupancy_grid = nerfacc.OccGridEstimator(
-            roi_aabb=self.scene_aabb,
```
hoanhle, line 121: I'm pretty sure that this would uniformly shrink or grow the scene_aabb from the origin, without considering the center.

A correct approach would be to first implement something like this in SceneBox

```
    def get_scaled_scene_box(self, scale_factor: Union[float, torch.Tensor] = 1.5):
        """Returns a new box that has been scaled by the given factor while
        maintaining its center.

        Args:
            scale_factor: How much to scale the bounding box by.

        Returns:
            SceneBox: A new SceneBox with the scaled bounding box.
        """
        center = self.get_center()
        scaled_aabb = (self.aabb - center) * scale_factor + center
        return SceneBox(aabb=scaled_aabb)
```

and then , call this function where needed.

hoanhle: It would be great if you could include this. I can review the PR once it's up. Also, I think it would be a good idea to disable scene contraction by default in `instant-ngp`.

PR #3433, opened 2024-09-16
Fix crop obb masking in pointcloud generation

nepfaff:
The same mask was used twice when `crop_obb` was None. This led to indexing errors.

I would provide example data but it should make sense when reading the code from line `158`.


jb-ye (approved):
LGTM, thanks!

Issue #3446, opened 2024-09-26
Depth-Nerfacto + Masking = Very slow


nepfaff:
**Describe the bug**
I'm trying to train depth-nerfacto with masking (depth + masks + alpha-transparent training). However, it is extremely slow when I don't move the images and masks to the GPU. This seems like a bug somewhere but I couldn't find it yet. I will continue investigating but wanted to post this in case someone has some intuition regarding where it might come from.

The key is that masking doesn't lead to a significant slowdown when using `nerfacto` without depth (especially due to the rejection sampling implementation). However, it somehow does when using `depth-nerfacto` with depth.

### Slow
```python
ns-train depth-nerfacto --pipeline.model.background_color "random" --pipeline.model.disable-scene-contraction False --data data --vis viewer_legacy
```
![image](https://github.com/user-attachments/assets/6e5210cd-9029-4d86-94e9-18d6f32bf0a1)

### Fast
```python
ns-train depth-nerfacto --pipeline.model.background_color "random" --pipeline.model.disable-scene-contraction False --data data --vis viewer_legacy --pipeline.datamanager.masks-on-gpu True --pipeline.datamanager.images-on-gpu True
```
![image](https://github.com/user-attachments/assets/ffd8c2cb-f193-45da-b5a4-501893aac82d)

### Data
Happy to share the dataset if someone is interested.
This is with 50 images of 1920x1440 on a 4090.

nepfaff:
@machenmusik adding you as I think you have previously worked with masks/ depth.
In particular, I'm looking for insights for why https://github.com/nerfstudio-project/nerfstudio/pull/2110 is no longer required for getting fast training with masks (makes zero difference in my experiments).

PR #3449, opened 2024-09-27
Keep depth_image on CPU

nepfaff:
This makes the `DepthDataset` consistent with the `SDFDataset` and `SemanticDataset`. From my experiments, this also reduces VRAM for `depth-nerfacto` without affecting training speed.

A step towards #3446.


akristoffersen (approved):
LGTM, adding @mpmisko since they might have more context.

mpmisko:
I have been quite out of date with this codebase, but LGTM.

PR #3452, opened 2024-09-27
Speed up PairPixelSampler

nepfaff:
This speeds up the extremely slow `PairPixelSampler`. The main issue seems to be with mask erosion being slow. I managed to speed it up a bit but it still isn't great. I also added the functionality from https://github.com/nerfstudio-project/nerfstudio/pull/2585 to `PairPixelSampler`.

Before any of the changes:
Screenshot from 2024-09-26 19-08-25

With only erosion:
image

With both erosion and rejection sampling:
image

This is an initial step towards #3446

### DIFF
--- a/nerfstudio/data/utils/pixel_sampling_utils.py
+++ b/nerfstudio/data/utils/pixel_sampling_utils.py
@@ -28,9 +28,7 @@ def dilate(tensor: Float[Tensor, "bs 1 H W"], kernel_size=3) -> Float[Tensor, "b
     Args:
         kernel_size: Size of the pooling region. Dilates/contracts 1 pixel if kernel_size is 3.
     """
-
-    unique_vals = torch.unique(tensor)
-    if any(val not in (0, 1) for val in unique_vals) or tensor.dtype != torch.float32:
+    if not (tensor.dtype == torch.float32 and torch.all((tensor == 0) | (tensor == 1))):
         raise ValueError("Input tensor should contain only values 0 and 1, and should have dtype torch.float32.")
 
     return torch.nn.functional.max_pool2d(tensor, kernel_size=kernel_size, stride=1, padding=(kernel_size - 1) // 2)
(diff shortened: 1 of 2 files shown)


nepfaff:
Not sure about the test failure:
```
ERROR: failed to authorize: failed to fetch oauth token: unexpected status from GET request to [https://auth.docker.io/token?](https://auth.docker.io/token?scope=repository%3Adocker%2Fdockerfile%3Apull&service=registry.docker.io:)
```

nerfstudio/data/utils/pixel_sampling_utils.py
```diff
@@ -28,9 +28,7 @@ def dilate(tensor: Float[Tensor, "bs 1 H W"], kernel_size=3) -> Float[Tensor, "b
     Args:
         kernel_size: Size of the pooling region. Dilates/contracts 1 pixel if kernel_size is 3.
     """
-
-    unique_vals = torch.unique(tensor)
-    if any(val not in (0, 1) for val in unique_vals) or tensor.dtype != torch.float32:
+    if not (tensor.dtype == torch.float32 and torch.all((tensor == 0) | (tensor == 1))):
```
akristoffersen, line 31: Sweet, this is so much cleaner

akristoffersen (approved):
This looks great to me, thanks!

Is there a reason we aren't also using the rejection sampling for the PatchPixelSampler? Maybe not in this PR, but now the rejection sampling is its own separate fn it seems like it would be a drop-in replacement.

nepfaff:
> Is there a reason we aren't also using the rejection sampling for the PatchPixelSampler? Maybe not in this PR, but now the rejection sampling is its own separate fn it seems like it would be a drop-in replacement.

There is no reason, and it is a drop-in replacement. However, I agree that a separate PR seems better [here](https://github.com/nerfstudio-project/nerfstudio/pull/3457). I opened one here (rebased this branch and started the new PR from it).

PR #3453, opened 2024-09-28
Change depth from float64 to float32

nepfaff:
It seems weird to me that depth used float64 while everything else used float32. Moreover, we specify that "Filepath points to a 16-bit or 32-bit depth image", so I don't seem a reason behind converting it to float64. Moreover, the depth prediction is also float32 (`ds_nerf_depth_loss`):
```python
loss = -torch.log(weights + EPS) * torch.exp(-((steps - termination_depth[:, None]) ** 2) / (2 * sigma)) * lengths
```
Here `steps` is float32 while `termination_depth` is float64. This PR changes it to also be float32.

Making this change leads to a small speedup.

Before change (float64):
Screenshot from 2024-09-28 14-39-52

After change (float32):
Screenshot from 2024-09-28 14-37-58

This is towards #3446.

### DIFF
--- a/nerfstudio/data/utils/data_utils.py
+++ b/nerfstudio/data/utils/data_utils.py
@@ -78,10 +78,10 @@ def get_depth_image_from_path(
         Depth image torch tensor with shape [height, width, 1].
     """
     if filepath.suffix == ".npy":
-        image = np.load(filepath) * scale_factor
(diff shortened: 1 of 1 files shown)


brentyi (approved):
Yeah, this looks unintentional. Thanks!!

PR #3454, opened 2024-09-28
Make PixelSampler the depth-nerfacto default

### LINKED ISSUE #3446: Depth-Nerfacto + Masking = Very slow

nepfaff:
The default pixel sampler for `depth-nerfacto` is `PairPixelSampler`. However, this is very slow due to mask erosion (so slow that it is basically unusable). I managed to speed it up a little here #3452. However, it is still much more efficient to use `PixelSampler`. Moreover, `PairPixelSampler` is only needed when using `SPARSENERF_RANKING` as the depth loss, but the default value is `DS_NERF`, which doesn't require `PairPixelSampler`. Hence, with the default depth loss, most people will be much better off with `PixelSampler` instead of `PairPixelSampler`.

I tested with both depth and masks using 50 images of resolution 1920x1440. Note, that `PairPixelSampler` is only slow when training with masks due to the need for mask erosion.

With `PairPixelSampler`:
image

With `PixelSampler`:
image

Closes #3446

### DIFF
--- a/nerfstudio/configs/method_configs.py
+++ b/nerfstudio/configs/method_configs.py
@@ -40,7 +40,6 @@ from nerfstudio.data.dataparsers.sitcoms3d_dataparser import Sitcoms3DDataParser
 from nerfstudio.data.datasets.depth_dataset import DepthDataset
 from nerfstudio.data.datasets.sdf_dataset import SDFDataset
 from nerfstudio.data.datasets.semantic_dataset import SemanticDataset
-from nerfstudio.data.pixel_samplers import PairPixelSamplerConfig
 from nerfstudio.engine.optimizers import AdamOptimizerConfig, RAdamOptimizerConfig
 from nerfstudio.engine.schedulers import (
     CosineDecaySchedulerConfig,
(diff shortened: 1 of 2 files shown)


akristoffersen (approved):
This makes sense to me. Approving.

It's a bit unfortunate that there's no obvious sign to the user that PairPixelSampler should be used when using `SPARSENERF_RANKING`. I believe it would silently fail if that depth loss was used without the pair pixel sampler. Could you add to the comment under the loss type in DepthNerfactoModelConfig to note this?

nepfaff:
Added a comment. It would be best to throw an exception in `DepthNerfactoModel` but that doesn't seem to be possible with the current structure.

PR #3457, opened 2024-09-30
Rejection sample mask for patch pixel sampler

nepfaff:
See #3452 for context.

**NOTE:** This must be merged after #3452. I wanted to specify the #3452 branch as the base branch but somehow couldn't find it in the GitHub UI...

### DIFF
--- a/nerfstudio/data/pixel_samplers.py
+++ b/nerfstudio/data/pixel_samplers.py
@@ -420,6 +420,10 @@ class PatchPixelSamplerConfig(PixelSamplerConfig):
     patch_size: int = 32
     """Side length of patch. This must be consistent in the method
     config in order for samples to be reshaped into patches correctly."""
+    rejection_sample_mask: bool = True
+    """Whether or not to use rejection sampling when sampling images with masks"""
+    max_num_iterations: int = 100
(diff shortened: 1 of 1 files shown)


nerfstudio/data/pixel_samplers.py
```diff
@@ -422,9 +462,20 @@ def sample_method(
             sub_bs = batch_size // (self.config.patch_size**2)
             half_patch_size = int(self.config.patch_size / 2)
             m = erode_mask(mask.permute(0, 3, 1, 2).float(), pixel_radius=half_patch_size)
```
akristoffersen, line 464: This is making me realize, instead of eroding, it might be better to just rejection sample patches across the image, rejecting if the whole patch isn't within the mask. Could be faster if the masks are large and the patches are small.

nepfaff: That is probably true. I didn't really look into the functionality of the `PatchPixelSampler` yet but can find some time to do so later this week.

akristoffersen: Gotcha. That could also go in a separate PR for sure.

akristoffersen (approved):
LGTM, will merge once you rebase.

PR #3459, opened 2024-10-02
Add option for refining the TSDF using the mesh's AABB

nepfaff:
I found the TSDF mesh quality to be highly dependent on the TSDF grid location and resolution. Currently, the user is responsible for providing a tight bounding box around the object to ensure good TSDF grid resolution in the areas of interest. However, this is often inconvenient and slow. It might also be hard to achieve when the NeRF is trained on a server for which the viewer isn't available.

In this PR, I suggest a simple solution for refining a rough bounding box estimate using the computed bounding box of the original mesh. In practice, I found this to lead to big improvements.

Note that this assumes that the density estimates are good enough to prevent floaters far from the object. I set the option to False by default, as this will be an issue if no special care is taken.

I trained `nerfacto` with alpha-transparent training on images of a mustard bottle:
```
ns-train nerfacto \
--pipeline.model.background_color "random" \
--pipeline.model.disable-scene-contraction True \
--data data
```

Exporting with rough estimate + no TSDF refinement:
```
ns-export tsdf --load-config outputs/.../config.yml \
--output-dir exports \
--num-pixels-per-side 2048 \
--resolution 250 250 250 \
--use-bounding-box True \
--bounding-box-min -.5 -.5 -.5 \
--bounding-box-max .5 .5 .5
```
no_refinement01

Exporting with rough estimate + TSDF refinement:
```
ns-export tsdf --load-config outputs/.../config.yml \
--output-dir exports \
--num-pixels-per-side 2048 \
--resolution 250 250 250 \
--use-bounding-box True \
--bounding-box-min -.5 -.5 -.5 \
--bounding-box-max .5 .5 .5 \
--refine-mesh-using-initial-aabb-estimate True
```
refined00

The refined mesh is much smoother.

### DIFF
--- a/nerfstudio/exporter/tsdf_utils.py
+++ b/nerfstudio/exporter/tsdf_utils.py
@@ -285,6 +285,8 @@ def export_tsdf_mesh(
     use_bounding_box: bool = True,
     bounding_box_min: Tuple[float, float, float] = (-1.0, -1.0, -1.0),
     bounding_box_max: Tuple[float, float, float] = (1.0, 1.0, 1.0),
+    refine_mesh_using_initial_aabb_estimate: bool = False,
+    refinement_epsilon: float = 1e-2,
 ) -> None:
     """Export a TSDF mesh from a pipeline.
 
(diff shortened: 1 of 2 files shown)


brentyi (approved):
This looks good to me, thanks @nepfaff!

PR #3466, opened 2024-10-04
Add `max_frames` + `max_output_size` arguments for Aria data processing

brentyi:
@AntonioMacaronio this has been helpful for some of datasets recently. Does it make sense to you? ( I know there's also the other open Aria + data processing PRs )

### DIFF
--- a/nerfstudio/scripts/datasets/process_project_aria.py
+++ b/nerfstudio/scripts/datasets/process_project_aria.py
@@ -203,7 +203,7 @@ class ProcessProjectAria:
     """Path to the output directory."""
     max_frames: int = 350
     """Number of frames to process."""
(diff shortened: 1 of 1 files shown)


AntonioMacaronio:
These changes make sense to me, but I am wondering about the fisheye_crop_radius

If a user limits the resolution of the image to be smaller than 1404, the fisheye_crop_radius would have to decrease proportionally to their new resolution right?

brentyi:
> If a user limits the resolution of the image to be smaller than 1404, the fisheye_crop_radius would have to decrease proportionally to their new resolution right?

Yeah, the code is a little bit hard to follow but since I'm updating the camera calibration based on the new image resolution this does end up getting handled correctly!

PR #3471, opened 2024-10-11
Tag-exported-files

simonbethke:
I would like to write the keyword "Nerfstudio" into exported ply files so other software can compensate for the 90° rotation Nerfstudio applies to the data (it doesn't really matter what is the standard and what isn't. I is generally helpful, to have the possibility to know who wrote a file)


simonbethke:
I made a git-mess... this PR should have contained only one tiny atomic change.

PR #3472, opened 2024-10-11
tag exported ply files

simonbethke:
I would like to write the keyword "Nerfstudio" into exported ply files so other software can compensate for the 90° rotation Nerfstudio applies to the data (it doesn't really matter what is the standard and what isn't. It is generally helpful, to have the possibility to know who wrote a file)


brentyi:
Would it make sense to include the nerfstudio version?

I think it can be read via:
```
from importlib.metadata import version
nerfstudio_version = version("nerfstudio")
```

simonbethke:
That would totally make sense. I was not aware of any way to read it :)

simonbethke:
Examples: 

comment Generated by Nerfstudio 1.0
comment Vertical Axis: z

comment Generated by SuperSplat 1.0
comment Vertical Axis: y

brentyi (approved):
Looks great, thanks @simonbethke!

Issue #3461, opened 2024-10-02
Crash on executing ns-export


SangwooRyu:
**Describe the bug**
When I execute ns-export, it prints nothing and ends.
And I found this creates dmp file in CrashDump
I opened it using Visual Studio and it says the error code is 0xc0000005

**To Reproduce**
On Windows environment (not sure this occurs for all Windows environment)
ns-export --help

**Expected behavior**
Execute ns-export and prints argument information

**Screenshots**
![스크린샷 2024-10-02 234800](https://github.com/user-attachments/assets/513a7264-cad5-4d0f-a037-54e802748242)

**Additional context**
I will attach the dump file
https://drive.google.com/file/d/1MeE5cPF6jtBob5m9oTI7n5mYH6fo6uXE/view?usp=sharing

Any idea why this occurs?

SharkWipf:
Reported an issue upstream: https://github.com/cnr-isti-vclab/PyMeshLab/issues/398
More discussion and troubleshooting has taken place on the Nerfstudio Discord in [this thread](https://discord.com/channels/1025504971606724712/1292813251624702004/1294134064193732648).

SharkWipf:
Update: The issue only happens with Python 3.8, not with newer Python versions.  
Since Python 3.8 is EOL already, and Nerfstudio works perfectly fine on Python 3.10 at least, I guess it's time to update the installation instructions to not recommend Python 3.8.

PR #3474, opened 2024-10-11
Restrict pymeshlab version on Windows (3461)

SharkWipf:
As per https://github.com/nerfstudio-project/nerfstudio/issues/3461 and https://github.com/cnr-isti-vclab/PyMeshLab/issues/398, currently pymeshlab==2023.12.post2 (latest at the time of writing) seems to be broken on Windows, breaking `ns-export` entirely in the process. 
_(One could wonder why it, and everything else, is even being imported unconditionally, ns-export takes seconds to launch and uses up 900MB of RAM just to display its --help, but that's a separate issue for another time)_

This is a temporary workaround to restrict the version of Pymeshlab being installed on Windows.

***I blatantly admit that I have not tested this.*** I don't have access to a Windows machine I can test this on right now, but we've been getting at least 5 different people reporting this issue over the past few days, so I figured it was more important to roll out a potential fix than to wait until I have time to test it. Manual downgrading of pymeshlab has fixed the issue for everyone who's tried however.


markdjwilliams:
While I have not tested your PR, I can confirm that downgrading pymeshlab to 2023.12.post1 solved the ns-export issue for me.

jb-ye (approved):

jb-ye:
I agree your proposal that we shouldn't wait for the fix.

PR #3478, opened 2024-10-14
update gsplat version to >=1.3.0

kerrj:
Don't think it's necessary to pin gsplat to 1.3.0.


maturk (approved):

brentyi:
Why not just bump to `==1.4.0`? That seems safer (given the history of breaking changes in gsplat).

PR #3486, opened 2024-10-16
Properly reset train state after render lock in viewer

kerrj:
There were some crashes in the viewer when the training state was set back to train(), instead of keeping the prior state of the model.

### DIFF
--- a/nerfstudio/viewer/render_state_machine.py
+++ b/nerfstudio/viewer/render_state_machine.py
@@ -148,6 +148,7 @@ class RenderStateMachine(threading.Thread):
                         device=self.viewer.get_model().device,
                     )
                     self.viewer.get_model().set_background(background_color)
+                was_training = self.viewer.get_model().training
                 self.viewer.get_model().eval()
                 step = self.viewer.step
                 try:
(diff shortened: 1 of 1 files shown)


nerfstudio/viewer/render_state_machine.py
```diff
@@ -168,9 +169,11 @@ def _render_img(self, camera_state: CameraState):
                         with torch.no_grad(), viewer_utils.SetTrace(self.check_interrupt):
                             outputs = self.viewer.get_model().get_outputs_for_camera(camera, obb_box=obb)
                 except viewer_utils.IOChangeException:
-                    self.viewer.get_model().train()
+                    if was_training:
+                        self.viewer.get_model().train()
                     raise
-                self.viewer.get_model().train()
+                if was_training:
```
brentyi, line 175: we could do:
```python
finally:
    if was_training:
        self.viewer.get_model().train()
```
and reduce some duplication?

kerrj: I don't think this is possible because of the re-raising behavior of the except? is it possible for the finally to run after raising another exception?

kerrj: Oh interesting it actually is possible in python, just updated it

brentyi: oh interesting good point, TIL also!!

brentyi (approved):

PR #3516, opened 2024-11-07
fix: saving ckpt error when ckpt folder has some folder inside

Crescent-Saturn:
When `save_only_latest_checkpoint` is True, it tried to delete everything in the chekpoint folder before saving the latest checkpoint. 

Since it is supposed that the `nerfstudio_models` only has checkpoint **files** inside, when there are some other **customized saved folders** inside, there will be an error.

This change will list `*.ckpt` files only instead of everything (include folder) to be compared with latest checkpoint file before deleting.

Thanks for your consideration


jb-ye (approved):
Look good to me

PR #3523, opened 2024-11-12
Improve default launch device for train

Noezor:
👋 folks,

Thanks for the great work on the project.

As reported on https://github.com/nerfstudio-project/nerfstudio/issues/2438, I struggled to start the machine on my m1 but found that it's because the documentation of the getting started didn't specify the device type, although I imagine a few people will try to run a gsplat on their M1.

I first thought I would change the documentation but we have the tools to spot if CUDA is appropriate or not anyways so am instead doing a code contrib.

I had two implementation choices: 
+ keeping the default as `cuda` and then verifying if `cuda` is available.
+ adding a new `None` default and then chosing the most appropriate default

The first choice had the advantage of avoiding potential other breaking changes that the second choice would create. However, I checked and the class I'm modifying is only used once in the `TrainerConfig` so that issue wasn't significant.

Using `None` has an advantage of disambiguating between the case where the user sets the parameter explicitely or not. The major issue I wanted to avoid using `None` is if someone has a cuda device, runs with `device_type=cuda` and goes afk only to find that his training indeed launched but took 20h instead of 5mn because we didn't find the cuda device for some sheneingan.

Let me know what you think and if you have ideas of where I should use this.[ I sometimes found ](https://github.com/nerfstudio-project/nerfstudio/blob/69ae4abbfd3c8736e684a2bd147f337b3615e446/nerfstudio/data/datamanagers/full_images_datamanager.py#L1)the default to be `cpu` or `cuda` and am not sure of the logic. I feel like anyways the code would be improved by doing `"cuda" if torch.cuda.is_available() else "cpu"` for the `cuda` case.


jb-ye:
I could still prefer option 1: keeping the default as `cuda` and then verifying if `cuda` is available.

jb-ye (approved):

PR #3527, opened 2024-11-18
Drop support for Python 3.8, suggest 3.10 by default

SharkWipf:
Python 3.8 is officially EOL, and at least one Nerfstudio dependency does not work with Python 3.8 anymore (#3461).
I've been running Nerfstudio on Python 3.10 myself without problems (although I've only tested on Linux). I remember having had problems in the past with some dependencies not yet supporting 3.11/3.12, but I don't know if that's been solved already. 3.10 seemed like the safest bet to target.
If you want to target a different python version by default, by all means, I just picked 3.10 because I knew that worked for me.

Reverts #3474 as it is no longer needed after this.

### DIFF
--- a/docs/quickstart/installation.md
+++ b/docs/quickstart/installation.md
@@ -5,16 +5,16 @@
 ::::::{tab-set}
 :::::{tab-item} Linux
 
-Nerfstudio requires `python >= 3.8`. We recommend using conda to manage dependencies. Make sure to install [Conda](https://docs.conda.io/en/latest/miniconda.html) before proceeding.
+Nerfstudio requires `python >= 3.9`. We recommend using conda to manage dependencies. Make sure to install [Conda](https://docs.conda.io/en/latest/miniconda.html) before proceeding.
 
 :::::
 :::::{tab-item} Windows
 
 :::{admonition} Note
 :class: info
-Nerfstudio on Windows is less tested and more fragile, due to way more moving parts outside of Nerfstudio's control.  
-The instructions also tend to break over time as updates to different Windows packages happen.  
-Installing Nerfstudio on Linux instead is recommended if you have the option.  
+Nerfstudio on Windows is less tested and more fragile, due to way more moving parts outside of Nerfstudio's control.
+The instructions also tend to break over time as updates to different Windows packages happen.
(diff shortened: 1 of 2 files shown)


brentyi:
This sounds good to me, @SharkWipf do you mind also bumping the Python versions used for the GH Actions workflows?

https://github.com/nerfstudio-project/nerfstudio/blob/main/.github/workflows/publish.yml
https://github.com/nerfstudio-project/nerfstudio/blob/main/.github/workflows/core_code_checks.yml

SharkWipf:
Whoops, forgot about those, should be done now.

SharkWipf:
The Dockerfile is already using Python 3.10 it seems.

SharkWipf:
(Surely I've got them all now, right?)

jb-ye:
@SharkWipf also run ruff formatter

SharkWipf:
@jb-ye Done, but it only changed files I haven't touched at all, is that expected? Like, the issues ruff found were entirely unrelated to this PR.

SharkWipf:
Welp, guess the github workflow doesn't like python 3.10...

SharkWipf:
Hmm. So the only pyright error from CI that I can reproduce locally is the matplotlib colormap one. The rest don't happen when I run it locally, nor do I understand them.  
As for the matplotlib error, it seems that with Python 3.8, matplotlib was stuck at version 3.7.3. On Python 3.10, it uses version 3.9.2.  
Between these versions, apparently the matplotlib colormap code has been getting reworked.  
I cannot find any documentation on `matplotlib.colormaps` even in the 3.7 documentation though. Was this even a public API to begin with? No `matplotlib.colormaps` seems to exist at all in the docs, which makes me think it was not an API that was ever meant to be used directly.  

Unfortunately these issues seem to run deeper than I can quickly whip up a fix for without diving into what exactly is going on here, this is all foreign territory for me.  
I've never run into any issues running Nerfstudio on Python 3.10 (or even 3.11) myself, but supposedly this has been an issue with any newer Python versions for a while.  

What do you want to do here?

ishipachev:
Windows instructions also say python=3.8, but gsplat wheels are built only for python=3.10.

PR #3538, opened 2024-12-02
Move util functions out of splatfacto

akristoffersen:
Nothing else currently uses some of the SH utils, but it might make sense to get them out of splatfacto.

I also moved the k nearest neighbors to utils since it doesn't depend on the model class.


jb-ye (approved):
Thanks for the refactorization!

Issue #3383, opened 2024-08-24
Green tint with splatfacto-big and bilagrad 


gradeeterna:
Splatfacto-big with bilagrad enabled has a green tint, not present with regular splatfacto + bilagrad, or splatfacto-big with bilagrad disabled:

splatfacto-big + bilagrad enabled:
<img width="1276" alt="splatfactobig-bilagrad" src="https://github.com/user-attachments/assets/d7847811-6335-44a8-a318-1637596d5dce">

splatfacto + bilagrad enabled:
<img width="1277" alt="splatfacto-bilagrad" src="https://github.com/user-attachments/assets/c57c9b5f-c731-471f-a29a-2b2e24ff8c2d">

splatfacto-big + bilagrad disabled:
<img width="1275" alt="splatfactobig-nobilagrad" src="https://github.com/user-attachments/assets/101f2389-46b8-4803-9680-17dea65c343a">

splatfacto + bilagrad enabled + some arguments to match splatfacto-big looks fine also `--pipeline.model.densify-grad-thresh 0.0006 --pipeline.model.cull-alpha-thresh=0.005  --pipeline.model.continue-cull-post-densification=False`
<img width="1273" alt="splatfacto-diy-big-bilagrad" src="https://github.com/user-attachments/assets/91518fa5-7fdd-419b-929c-01db4638193f">

@SharkWipf commented on Discord that the only other difference between regular and big splatfacto is the learning rate:
![Screenshot_20240824-141552](https://github.com/user-attachments/assets/2e3c881c-0cde-4556-9d09-32d5b8219ebc)

I've only tested with this tricky dataset where the exposure is all over the place, but can test some others soon. 

Here is the ground truth image, and another over-exposed image from nearby:
<img width="845" alt="gt1" src="https://github.com/user-attachments/assets/eb7c168d-bd2c-4928-a355-002080d15357">
<img width="845" alt="gt2overexposed" src="https://github.com/user-attachments/assets/90787604-d0a2-43c9-a46d-fa5ab86e3b61">

abrahamezzeddine:
Out of curiousity, how do I enable bilagird with splatfacto? I would like to try this myself.☺️ thanks!

jb-ye:
@abrahamezzeddine `--pipeline.model.use_bilateral_grid True`

xiemeilong:
[...]

PR #3548, opened 2024-12-14
Finalize MCMC strategy and some tiny fix

KevinXu02:
Finalize MCMC strategy https://github.com/nerfstudio-project/nerfstudio/pull/3436, fix bilagird lr rate for splatfacto-big https://github.com/nerfstudio-project/nerfstudio/issues/3383 and some small changes to colmap dataparser (auto iterates possible colmap paths).
Tested on bicycle with random/sfm initialization.
To use: `ns-train splatfacto-mcmc`


pablovela5620:
tried to use this but got a `torch.cuda.OutOfMemoryError: CUDA out of memory. Tried to allocate 1.47 GiB. GPU ` error with mcmc, where I had no problems with vanilla splatfacto

KevinXu02:
> tried to use this but got a `torch.cuda.OutOfMemoryError: CUDA out of memory. Tried to allocate 1.47 GiB. GPU ` error with mcmc, where I had no problems with vanilla splatfacto

Yes this might happen since MCMC uses a fixed number of gs for the scene. Could you please try add`--pipeline.model.max_cap [max num of gs]` in your training cmd? And a common number would be 1000000 but you can make it smaller.

nerfstudio/data/dataparsers/colmap_dataparser.py
```diff
@@ -251,6 +251,13 @@ def _get_image_indices(self, image_filenames, split):
     def _generate_dataparser_outputs(self, split: str = "train", **kwargs):
         assert self.config.data.exists(), f"Data directory {self.config.data} does not exist."
         colmap_path = self.config.data / self.config.colmap_path
+        if not colmap_path.exists():
+            # Try detecting the colmap path automatically
```
lionlai1989, line 255: Personally, I would choose to let it fail instead of manually searching in the pre-defined paths.

KevinXu02: Maybe first asking users if they want to do this search? Or I can just remove this part. (since this path error happens a lot to me

pablovela5620:
> > tried to use this but got a `torch.cuda.OutOfMemoryError: CUDA out of memory. Tried to allocate 1.47 GiB. GPU ` error with mcmc, where I had no problems with vanilla splatfacto
> 
> Yes this might happen since MCMC uses a fixed number of gs for the scene. Could you please try add`--pipeline.model.max_cap [max num of gs]` in your training cmd? And a common number would be 1000000 but you can make it smaller.

Even setting to 500,000 leads to a OOM error sadly (I'm using a 3060), I had to reduce to 100,000 gaussian. That seems like too few(?) but I honestly don't know. I'll do some more testing

nerfstudio/models/splatfacto.py
```diff
@@ -156,6 +156,22 @@ class SplatfactoModelConfig(ModelConfig):
...
+    """The default strategy will be used if strategy is not specified. Other strategies, e.g. mcmc, can be used."""
+    cap_max: int = 1_000_000
+    """Maximum number of GSs. Default to 1_000_000."""
+    noise_lr: float = 5e5
+    """MCMC samping noise learning rate. Default to 5e5."""
+    min_opacity: float = 0.005
+    """GSs with opacity below this value will be pruned. Default to 0.005."""
+    verbose: bool = False
+    """Whether to print verbose information. Default to False."""
+    max_steps: int = 30_000
```
kerrj, line 169: this parameter is unused, I think we can delete it?

nerfstudio/models/splatfacto.py
```diff
@@ -156,6 +156,22 @@ class SplatfactoModelConfig(ModelConfig):
     """Shape of the bilateral grid (X, Y, W)"""
     color_corrected_metrics: bool = False
     """If True, apply color correction to the rendered images before computing the metrics."""
+    strategy: Literal["default", "mcmc"] = "default"
+    """The default strategy will be used if strategy is not specified. Other strategies, e.g. mcmc, can be used."""
+    cap_max: int = 1_000_000
+    """Maximum number of GSs. Default to 1_000_000."""
+    noise_lr: float = 5e5
+    """MCMC samping noise learning rate. Default to 5e5."""
+    min_opacity: float = 0.005
```
kerrj, line 165: can we re-use the `cull_alpha` parameter from above? having multiple is a bit confusing

nerfstudio/models/splatfacto.py
```diff
@@ -156,6 +156,22 @@ class SplatfactoModelConfig(ModelConfig):
...
+    """GSs with opacity below this value will be pruned. Default to 0.005."""
+    verbose: bool = False
```
kerrj, line 167: i think we can delete this and default to false for the sake of simplicity (since it only affects the strategies)

nerfstudio/configs/method_configs.py
```diff
@@ -696,7 +696,68 @@
...
+    mixed_precision=False,
+    pipeline=VanillaPipelineConfig(
+        datamanager=FullImageDatamanagerConfig(
+            dataparser=NerfstudioDataParserConfig(load_3D_points=True),
+            cache_images_type="uint8",
+        ),
+        model=SplatfactoModelConfig(
+            strategy="mcmc",
+            mcmc_opacity_reg=0.01,
+            mcmc_scale_reg=0.01,
```
kerrj, line 725: I found for MCMC that allowing refinement to last until 25000 steps is better for quality, ie the `stop_split_at` param

nerfstudio/models/splatfacto.py
```diff
@@ -156,6 +156,22 @@ class SplatfactoModelConfig(ModelConfig):
     """Shape of the bilateral grid (X, Y, W)"""
     color_corrected_metrics: bool = False
     """If True, apply color correction to the rendered images before computing the metrics."""
+    strategy: Literal["default", "mcmc"] = "default"
+    """The default strategy will be used if strategy is not specified. Other strategies, e.g. mcmc, can be used."""
+    cap_max: int = 1_000_000
```
kerrj, line 161: can we rename to something more readable like `max_gs_num`?

nerfstudio/data/dataparsers/colmap_dataparser.py
lionlai1989: I may misunderstand something here. But I think if there is an path error happening to users, users should be the ones to make the path right.

gradeeterna:
I'm also getting OOM errors here with `--pipeline.model.max-gs-num 1000000` on a 3090. Same dataset works fine with splatfacto and MCMC+bilagrid directly in gsplat.

nerfstudio/configs/method_configs.py
```diff
@@ -696,7 +696,69 @@
...
+    max_num_iterations=30000,
+    mixed_precision=False,
+    pipeline=VanillaPipelineConfig(
+        datamanager=FullImageDatamanagerConfig(
+            dataparser=NerfstudioDataParserConfig(load_3D_points=True),
+            cache_images_type="uint8",
+        ),
+        model=SplatfactoModelConfig(
+            strategy="mcmc",
+            mcmc_opacity_reg=0.01,
```
kerrj, line 724: can we avoid redefining these here so that the defaults use these values?

kerrj: that said we also need to update cull_alpha_threshold to .005 here!

kerrj:
> I'm also getting OOM errors here with `--pipeline.model.max-gs-num 1000000` on a 3090 with 24GB VRAM. Same dataset works fine with splatfacto and MCMC+bilagrid directly in gsplat.

It's possible there are some memory differences between gsplat and nerfstudio because of dataloader overhead. When you run gsplat is the memory usage very close to 24GB? it may help to make sure the images are cached on CPU inside splatfacto-mcmc (there's a FullImagesDataManager parameter for this). If splatfacto-mcmc is taking significantly more memory than gsplat's version that would be surprising, but I think a difference of ~1GB is expected. MCMC in general will take more memory than the default strategy since the re-sampling step is a little memory hungry.

vectorman1:
I've observed the OOM issue using Gsplat directly through `simple_trainer` - same dataset, running it N times, errors out with OOM at some point. I would say it's safe to assume that the nerfstudio code is not the origin. 

What is this PR missing at this point in order to get it merged?

kerrj (approved):
I think this looks good to me now!

gradeeterna:
> > I'm also getting OOM errors here with `--pipeline.model.max-gs-num 1000000` on a 3090 with 24GB VRAM. Same dataset works fine with splatfacto and MCMC+bilagrid directly in gsplat.
> 
> It's possible there are some memory differences between gsplat and nerfstudio because of dataloader overhead. When you run gsplat is the memory usage very close to 24GB? it may help to make sure the images are cached on CPU inside splatfacto-mcmc (there's a FullImagesDataManager parameter for this). If splatfacto-mcmc is taking significantly more memory than gsplat's version that would be surprising, but I think a difference of ~1GB is expected. MCMC in general will take more memory than the default strategy since the re-sampling step is a little memory hungry.

Hey, the images are cached on CPU but I’m still getting OOM errors on smallish datasets if I set the max count to 2 million or higher. It’s only using 5gb VRAM at the start of training, but OOM errors at around 3.5k steps.

abrahamezzeddine:
I just tried this out myself and I am getting 8GB och VRAM with default MCMC settings, so no OOM here my part. 

I however still see this typical haze sometimes over the. Zooming out, and you can see incredibly large splats being generated, which I suspect is the reason. With varying opacities for these splats, if they intersect the model in a specific view, they will create a haze over the model, rotate and they dissappear.

image
image


Some reference images. Doesn't the model prune very large splats?

hnguyen25:
This is great! Thank you for all the work getting this integrated.

I am also seeing the huge floaters/haze in the MCMC scene. Is there a plan to get this fixed anytime soon?

Issue #2681, opened 2023-12-15
Allow gaussian splatting to initialize start points from other data pipelines than COLMAP


kerrj:
Currently, gaussian splatting intakes a pointcloud for initialization either through: 1) using a COLMAP dataset, or 2) adding a pointcloud to `transforms.json` and using the nerfstudio dataparser to read it. Currently, data intake pipelines like polycam, metashape, etc do not output a pointcloud and thus gaussian splatting will use a random initialization of points, which produces much worse results.

To fix this, we should edit the data intake scripts to add point clouds to the output `transforms.json` files and use the nerfstudio dataparser to read them. #2557 added the capability to load .ply files in the nerfstudio dataparser 

- [ ] Add to the COLMAP data processing pipeline to save pointclouds of the SFM points as output, and add it to the `transforms.json`
- [ ] Switch to nerfstudio dataparser as the default for splatting
- [ ] implement pointcloud saving in other data processing pipelines

jb-ye:
FYI, metashape has a script for exporting colmap dataset (undistorted images): https://github.com/agisoft-llc/metashape-scripts/blob/master/src/export_for_gaussian_splatting.py

FreakTheMighty:
@kerrj is it your experience that the guassian-splat pipeline runs, just with reduced quality, when using other data sources? What I'm seeing ATM is that when using polycam as input an error is thrown.

```py
    assert colmap_path.exists(), f"Colmap path {colmap_path} does not exist."
AssertionError: Colmap path Garage/colmap/sparse/0 does not exist.
```

Tekn1ck:
I am also getting this when using RealityCapture. It would be super to get this up and running. If we could export the point cloud that Reality Capture generates would that help?

    assert colmap_path.exists(), f"Colmap path {colmap_path} does not exist."
AssertionError: Colmap path C:\Users\1user\nerfstudio\data\nerfstudio\farmhouse\colmap\sparse\0 does not exist.

(nerfstudio) C:\Users\Tekuser\nerfstudio>ns-train nerfacto --data C:\Users\1user\nerfstudio\data\nerfstudio\farmhouse
[15:24:27] Using --data alias for --data.pipeline.datamanager.data

brentyi:
I think these errors are because the default dataparser for `gaussian-splatting` is set to `colmap`, while all other models set it to `nerfstudio-data`. There are plans to transition splatting over to the latter.

[...]

PR #3556, opened 2024-12-22
Add a point cloud option for Record3D

Jameson-Crate:
[#2681](https://github.com/nerfstudio-project/nerfstudio/issues/2681)

Added the option to add a point cloud to Record3D. Record3D has an export ply option which gives a dense point cloud for every camera. I just combine them all and downsample to make the point cloud sparse using [open3d](https://www.open3d.org/html/tutorial/geometry/pointcloud.html#Point-cloud). I also changed the documentation under Getting Started > Using Custom Data > Record3D.


nerfstudio/scripts/process_data.py
```diff
@@ -101,7 +104,10 @@ def main(self) -> None:
             )
 
         metadata_path = self.data / "metadata.json"
-        record3d_utils.record3d_to_json(copied_image_paths, metadata_path, self.output_dir, indices=idx)
+        ply_path = self.ply if hasattr(self, "ply") else None
```
akristoffersen, line 107: Not sure this is necessary? I think you can just pass in self.ply into `record3d_to_json`.

nerfstudio/scripts/process_data.py
```diff
@@ -49,6 +49,9 @@ class ProcessRecord3D(BaseConverterToNerfstudioDataset):
     2. Converts Record3D poses into the nerfstudio format.
     """
 
+    ply: Optional[Path] = None
```
akristoffersen, line 52: Nit: Might want to call this `ply_dir` instead to emphasize that it's a directory, not a ply file itself.

Is the reason for doing a dir instead of the ply filepath because Record3D can generate multiple ply's?

nerfstudio/process_data/record3d_utils.py
```diff
@@ -87,6 +95,23 @@ def record3d_to_json(images_paths: List[Path], metadata_path: Path, output_dir:
...
+    # If .ply directory exists add the sparse point cloud for gsplat point initialization
+    if ply_dirname is not None:
+        assert ply_dirname.exists(), f"Directory not found: {ply_dirname}"
+        assert ply_dirname.is_dir(), f"Path given is not a directory: {ply_dirname}"
+
+        # Create sparce point cloud
+        pcd = o3d.geometry.PointCloud()
+        for ply_filename in ply_dirname.iterdir():
+            temp_pcd = o3d.io.read_point_cloud(str(ply_filename))
+            pcd += temp_pcd.voxel_down_sample(voxel_size=0.8)
```
akristoffersen, line 107: Should probably make `voxel_size` an argument, and maybe include it in the ProcessRecord3D settings.

akristoffersen (changes requested):
Thanks for doing this! some minor changes requested but otherwise LGTM

nerfstudio/process_data/record3d_utils.py
Jameson-Crate: thats a good idea

nerfstudio/scripts/process_data.py
Jameson-Crate: yea exactly, when you export Record3D gives a bunch of ply's, ill make the change rn

nerfstudio/scripts/process_data.py
Jameson-Crate: oh you're right, sorry about that

docs/quickstart/custom_dataset.md
```diff
@@ -268,6 +268,35 @@ ns-process-data record3d --data {data directory} --output-dir {output directory}
 ns-train nerfacto --data {output directory}
 ```
 
+### Adding a Point Cloud
+
+Adding a point cloud is useful for avoiding random initialization
```
akristoffersen, line 273: nit: seems like this was split into multiple lines unnecesarily?

akristoffersen (approved):
LGTM!

Jameson-Crate:
Thanks!

PR #3557, opened 2024-12-23
Adding docs for feature-splatting as an external method

RogerQi:
Adding documentation for Feature Splatting for computing SAM-enhanced CLIP features and distillation based on gsplats.


kerrj (approved):
thanks!

PR #3561, opened 2025-01-06
refactor: use depends-on instead of depends_on

tdejager:
The `depends_on` will be deprecated in the next pixi version. This syntax has been supported for a while now, so I think a migration won't cause too many issues :) Was a failure in our downstream tests :)


kerrj (approved):
lgtm, thanks!

PR #3565, opened 2025-01-09
Add GLOMAP to Dockerfile

brian-xu:
Builds GLOMAP for use in the dockerfile. The version of CMake provided by apt (3.28) is unable to build GLOMAP from source, so we download and build a compatible version instead.


jkulhanek (approved):
LGTM

PR #3567, opened 2025-01-10
Fix cli bug in high version of tyro

Tavish9:
In higher version of tyro(>0.9.3), the cli cannot recognize the following config definition:

```python
"""Use the same proposal network. Otherwise use different ones."""
proposal_net_args_list: List[Dict] = field(
    default_factory=lambda: [
        {"hidden_dim": 16, "log2_hashmap_size": 17, "num_levels": 5, "max_res": 128, "use_linear": False},
        {"hidden_dim": 16, "log2_hashmap_size": 17, "num_levels": 5, "max_res": 256, "use_linear": False},
    ]
)
```

## Reproduction:
`ns-train nerfacto --pipeline.model.proposal-net-args-list.0.hidden-dim 1`
1. In `typo==0.9.2`:

image

2. In `typo>0.9.2`:

image

### DIFF
--- a/docs/developer_guides/new_methods.md
+++ b/docs/developer_guides/new_methods.md
@@ -83,7 +83,7 @@ The `NERFSTUDIO_METHOD_CONFIGS` environment variable additionally accepts a func
 ```python
 """my_method/my_config.py"""
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from nerfstudio.engine.trainer import TrainerConfig
 from nerfstudio.plugins.types import MethodSpecification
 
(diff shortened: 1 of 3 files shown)


brentyi:
Hi @Tavish9 thanks for pointing this out and making this PR!

I made some changes, to summarize:
- Fixed the bug upstream in `tyro`.
- Bumped the minimum `tyro` version.
- Bumped the minimum `typeguard` version, which is required by `tyro`.
- Found that this caused problems with Nerfstudio's CI, since newer versions of `typeguard` aren't compatible with `jaxtyping`.
    - There's some discussion in https://github.com/agronholm/typeguard/issues/353, the incompatibility seems fundamental.
- It seems unideal to keep a 3+ year old version of `typeguard` pinned indefinitely, so I switched jaxtyping's pytest hook. Running typeguard in CI seems unnecessary since we added static checking in https://github.com/nerfstudio-project/nerfstudio/pull/1944.

Tavish9:
Hi @brentyi,  thanks for the hard work! 🤗

Issue #2983, opened 2024-03-05
NeRF-W


cfeng16:
Hi, nice work! I am wondering if there is any chance I can run NeRF-W by ns-train directly?

maturk:
yea a similar implementation is the semantic_nerfw model [here](https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/models/semantic_nerfw.py). I am not so familiar with it, and its not documented but it has the uncertainty loss and other features from the nerf in the wild paper.

Issue #2297, opened 2023-08-02
Transient Embeddings


samiragarwala:
Hi,
Thank you for the great work! I noticed that transient embeddings from NeRF-W seem to have been implemented in semantic NeRF-W with the appropriate loss functions but there is a message noting that they aren't working completely. I was wondering if you could share a bit more about the issues being encountered in getting the transient disentaglement working?
Thank you!

kerrj:
I'm not sure of the history behind this. I think maybe @ethanweber would know?

auzyze:
Hi, is there any update on this question? 
I tried to enable the transient embedding and train on sitcoms-3d dataset. With default settings, the rendered images are very dark during the training process. I modified the beta minimum value but the problem is not solved. By observing the training losses, the rgb loss curve saturated fairly soon, after the uncertainty loss goes down to a certain level. Could you suggest any possible solutions to this issue? Thanks!

PR #3574, opened 2025-01-17
Added Nerf in the Wild to Nerfacto

Dawars:
I added the correct transient rendering and loss from NerfW based on the **semantic-nerfw** and https://github.com/kwea123/nerf_pl/tree/nerfw code bases.

You can run the training using the following command:
`ns-train nerfacto --pipeline.model.use_transient_embedding True --pipeline.model.hidden_dim_transient 128 --pipeline.model.transient_embed_dim 16 --data [data-path]`

I didn't get to test all cases (normal rendering, etc)
But the transient components seem to be working correctly now.
Before with semantic nerf implementation:
image

After the change:
image

Let me know if I should add the fix to semantic nerf as well.

### DIFF
--- a/nerfstudio/models/nerfacto.py
+++ b/nerfstudio/models/nerfacto.py
@@ -41,7 +41,13 @@ from nerfstudio.model_components.losses import (
     scale_gradients_by_distance_squared,
 )
 from nerfstudio.model_components.ray_samplers import ProposalNetworkSampler, UniformSampler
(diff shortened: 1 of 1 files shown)


Dawars:
https://github.com/nerfstudio-project/nerfstudio/issues/2983 https://github.com/nerfstudio-project/nerfstudio/issues/2297

Issue #3586, opened 2025-02-06
AssertionError when using ns-export


gldanoob:
**Describe the bug**
Here is the log when I attempt to export a pointcloud or mesh (Commit `4a3e3e6`):
```
nerfstudio ❯ ns-export pointcloud --load-config outputs/stump/nerfacto/2024-11-11_022940/config.yml --o
utput-dir exports/mesh/ --num-points 1000000 --remove-outliers True --normal-method open3d --obb_center--0.1661160403 0.0640090148 -0.4701777494 --obb_rotation -0.0694655468 0.0561346174 -0.0039036355 --obb
_scale 1.0000000000 1.0000000000 1.0000000000
[03:07:26] Auto image downscale factor of 2                                                 nerfstudio_dataparser.py:484
Variable resolution, using variable_res_collate
Loading latest checkpoint from load_dir
✅ Done loading checkpoint from outputs/stump/nerfacto/2024-11-11_022940/nerfstudio_models/step-000029999.ckpt
Traceback (most recent call last):
  File "/home/gldanoob/miniconda3/envs/nerfstudio/bin/ns-export", line 8, in <module>
    sys.exit(entrypoint())
  File "/home/gldanoob/dev/nerfstudio_dev/nerfstudio/scripts/exporter.py", line 671, in entrypoint
    tyro.cli(Commands).main()
  File "/home/gldanoob/dev/nerfstudio_dev/nerfstudio/scripts/exporter.py", line 144, in main
    assert pipeline.datamanager.train_pixel_sampler is not None
AssertionError
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/home/gldanoob/miniconda3/envs/nerfstudio/lib/python3.8/multiprocessing/popen_fork.py", line 27, in poll
    pid, sts = os.waitpid(self.pid, flag)
  File "/home/gldanoob/miniconda3/envs/nerfstudio/lib/python3.8/site-packages/torch/utils/data/_utils/signal_handling.py", line 66, in handler
    _error_if_any_worker_fails()
RuntimeError: DataLoader worker (pid 19392) is killed by signal: Terminated.
```

However it shows expected behavior if `nerfstudio` is installed via pip (version 1.1.5)

**To Reproduce**
1. Install `nerfstudio` via git (Commit `4a3e3e6`)
2. Run `ns-export` on any nerfacto model

**Expected behavior**
Nerfstudio exports the mesh without errors

[...]

PR #3587, opened 2025-02-06
Fixing Assertion issues when exporting NeRFs to poisson meshes and pointclouds

AntonioMacaronio:
Since the [dataloading PR](https://github.com/nerfstudio-project/nerfstudio/pull/3216) moved the pixel sampling process outside of being a datamanager field and into RayBatchStream, some assertions that test this will fail. This PR double checks the datamanager type before performing assertions when using `ns-export pointcloud` or `ns-export poisson` 

This addresses issues such as https://github.com/nerfstudio-project/nerfstudio/issues/3586


akristoffersen (approved):
LGTM

Issue #3579, opened 2025-01-23
Bounding box not working when extracting a mesh with ns-export poisson


maurock:
Thank you very much for this awesome library! I'm encountering an issue when I extract a mesh from a Nerfacto representation using `ns-export poisson`. Instead of extracting the object, the resulting mesh represents a hemisphere surrounding the object. This suggests that the white background is being reconstructed, which could be part of the issue.

TO mitigate this problem, I used the `--bounding-box` parameters to restrict the volume:

`ns-export poisson ... --bounding-box-max 0.1 0.1 0.1 --bounding-box-min -0.1 -0.1 -0.1 --use-bounding-box True`

However, this did not change the output. I saw that older version of nerfstudio use `--obb` parameters rather than bounding boxes, so I am wondering whether there is a bug in the current configuration. Could you please advise on whether I am setting the boundaries incorrectly? 

Thank you for your time!

Charley-xiao:
Same issue. Did you resolve it? 😢

maurock:
Hi @Charley-xiao, unfortunately I have not solved the issue yet. Looking at the code, I believe there is a bug in the `--bounding-box` parameters as they are not used. A suboptimal workaround I am using is to call one of the classes in `nerfstudio/scripts/exporter.py`. There, you can pass the `obb` parameters, and they work as intended. Although it is not the same as simply passing the boundaries of a bounding box.

I should mention that the quality of the extracted mesh is extremely poor. They're just floaters in my case, on a simple YCB object with known camera parameters. There might be something else I need to do to extract a proper mesh, and I'd be very grateful for any advice the authors might have. I have used 40 images in the training set and 10 in the test set.

IsaacBerman:
Yes, it seems to have broken in a newer version, causing really bad results on low iteration trainings. Does anyone know if the authors have posted a solution anywhere?

IsaacBerman:
This seems related/same issue: https://github.com/nerfstudio-project/nerfstudio/issues/3581

PR #3591, opened 2025-02-12
Fixing bounding box issue when exporting poisson mesh

HeaIn:
When exporting poisson mesh, ```--use_bounding_box, --bounding_box_min, --bounding_box_max``` were previously used to crop point clouds. However, these parameters are no longer used in ```ExportPoissonMesh``` class. Instead, ```--obb_center, --obb_rotation, --obb_scale``` are now used.

These parameters should be removed to prevent confusion such as #3579.


brentyi (approved):
Thanks!!

PR #3598, opened 2025-02-19
Fix seams/gaps when exporting mesh using marching cubes Fixes #2721

### LINKED ISSUE #2721: Gaps after "ns-export marching cubes" on the outer surface found using neus-facto from nerfstudio
**Describe the bug**
Extracted model shows lined gaps on the surface. See Picture below:

image

**To Reproduce**
1. Executed the following command on my trained model with neus-facto: `ns-export marching-cubes --load-config outputs/Room/config.yml --output-dir Room`
2.` open3d draw Room/mesh.obj`

**Expected behavior**
I would have expected a watertight closed surface. No weird artifacts, when I open sdf_marching_cubes_mesh.ply of the same trained and extracted model. Btw. I noticed that the same problem appears also to the standard replica-room0 case. No closed surface found! the rendered model has gaps.

Dawars:
After concatenating the meshes it is necessary to merge the vertices at the seams so that the following simplification doesn't treat the separately.

### DIFF
--- a/nerfstudio/exporter/marching_cubes.py
+++ b/nerfstudio/exporter/marching_cubes.py
@@ -251,4 +251,5 @@ def generate_mesh_with_multires_marching_cubes(
                     meshes.append(meshcrop)
 
     combined_mesh: trimesh.Trimesh = trimesh.util.concatenate(meshes)  # type: ignore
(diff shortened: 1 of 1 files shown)


akristoffersen (approved):
LGTM!

PR #3606, opened 2025-03-03
Clamp values to prevent overshooting artifacts during image extraction with ns-process 

abrahamezzeddine:
Hello,

I noticed during conversion from equirectangular 360 camera image frames that there would be quite a lot of visible artifacts when inspecting the images. This usually occurs near edges when there's a lot of very high or very low contrast, causing apparent overshoots of the pixels. See image below;

Before code change:
image

After code change:
image


kerrj (approved):
lgtm thanks!

PR #3612, opened 2025-03-14
colmap_dataparser: colmap's k4 parameter is valid for fisheye model

### LINKED ISSUE #3611: Throw an error on non-zero k4, k5 or k6
Hi @jb-ye @devernay, why is there a problem when k4 is not equal to 0?  #3355 #3381  When reading the camera distortion coefficients such as [this](https://github.com/nerfstudio-project/nerfstudio/blob/73fe54dda0b743616854fc839889d955522e0e68/nerfstudio/process_data/colmap_utils.py#L262C2-L278C42), k4 is correctly assigned as the fourth-order radial distortion coefficient rather than the fourth coefficient. The current approach makes data with non-zero k4 coefficient like fisheye unusable.

f-dy:
fixes https://github.com/nerfstudio-project/nerfstudio/issues/3611


ishipachev:
@f-dy 
Thank you. I had the same issue, was about to make a PR, but it seems we already have PR for that.

PR #3613, opened 2025-03-15
fix: remove safeguards for parsing colmap data

### LINKED ISSUE #3611: Throw an error on non-zero k4, k5 or k6

Yubel426:
Fix #3611. The safeguards introduced in #3381 makes datasets with non-zero k4 coefficient such as fisheye unusable. When reading camera distortion coefficients such as [OPENCV camera](https://github.com/nerfstudio-project/nerfstudio/blob/73fe54dda0b743616854fc839889d955522e0e68/nerfstudio/process_data/colmap_utils.py#L262C2-L278C42), k4 is correctly assigned as the fourth-order radial distortion coefficient rather than the fourth coefficient. Therefore, the safeguards can be removed.

### DIFF
--- a/nerfstudio/cameras/camera_utils.py
+++ b/nerfstudio/cameras/camera_utils.py
@@ -380,9 +380,8 @@ def _compute_residual_and_jacobian(
     distortion_params: torch.Tensor,
 ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
     """Auxiliary function of radial_and_tangential_undistort() that computes residuals and jacobians.
-    Adapted from MultiNeRF:
-    https://github.com/google-research/multinerf/blob/b02228160d3179300c7d499dca28cb9ca3677f32/internal/camera_utils.py#L427-L474
-
+    Adapted from nerfies:
+    https://github.com/google/nerfies/blob/1a38512214cfa14286ef0561992cca9398265e13/nerfies/camera.py#L39C1-L71C40
     Args:
         x: The updated x coordinates.
         y: The updated y coordinates.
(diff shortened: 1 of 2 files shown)


Yubel426:
@kerrj @jb-ye

f-dy:
This is incorrect. #3612 is correct.
colmap's k4, k5 and k6 are still unsupported for the perspective model.
k4 should be supported for the fisheye model.

Yubel426:
Hi @f-dy, for the perspective models—"SIMPLE_PINHOLE", "PINHOLE", "SIMPLE_RADIAL", and "RADIAL"—their k4 parameter is correctly set to 0 when reading the parameters. I'm not sure what the purpose of this error check is; please let me know if I've misunderstood anything.

Or are you saying that the k4 in this function does not correspond to the k4 in COLMAP? In other words, this function doesn't support k4, right?
https://github.com/nerfstudio-project/nerfstudio/blob/73fe54dda0b743616854fc839889d955522e0e68/nerfstudio/cameras/camera_utils.py#L375

ishipachev:
@Yubel426 
You shouldn't set k4 to zero when you read camera parameters if this k4 is not zero.
Imagine you have some random function over [a;b] and you have a polynom which approximates it: `y = a * x^2 + bx + c` 
You can't "approximate" it further to linear model just by dropping a and having something like y = bx + c.
Instead you need to build a proper new approximation, not just dropping one coefficient from the reacher one.

PR #3627, opened 2025-04-04
Add support for setting WandB Entity via experiment config

IQisMySenpai:
This PR adds support for specifying the Weights & Biases (WandB) entity (team or organization) directly in the Nerfstudio experiment config.

## Why?
Currently, users must either:
- Rely on their default WandB entity, or
- Move runs manually to the correct entity after they have finished

This can be limiting when working across multiple teams or accounts. With this change, users can now explicitly set the WandB entity in the config, making it easier to log runs under different teams or organizations without needing to modify global environment settings.

## Notes
- The entity must already exist in WandB.
- If not specified, WandB will continue to fall back to the default entity (as it is currently), as defined in the user settings under “Default location to create new projects” > “Default team”.

As this is my first PR to Nerfstudio, so I’m happy to take any suggestions or feedback!

### DIFF
--- a/nerfstudio/configs/experiment_config.py
+++ b/nerfstudio/configs/experiment_config.py
@@ -75,6 +75,8 @@ class ExperimentConfig(InstantiateConfig):
     """Relative path to save all checkpoints."""
     load_scheduler: bool = True
     """Whether to load the scheduler state_dict to resume training, if it exists."""
+    entity_name: Optional[str] = None
+    """WandB entity name. If None, will use the default entity name set in WandB config."""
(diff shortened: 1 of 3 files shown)


brentyi:
Is this argument useful for other writers, like Comet etc? If not, could we rename it to `wandb_entity_name`?

IQisMySenpai:
It seems comet has something similar to entities called `workspace`, which is currently also defaulted to the user's workspace. I'll do some more research for the other writers and integrate their different entity solutions.

PR #3637, opened 2025-04-18
Clean | Remove `timm` dependency

Totoro97:
I searched the codebase and found that timm is actually not used. It seems this dependency was introduced in the following PR: https://github.com/nerfstudio-project/nerfstudio/pull/2284. However, I couldn’t find any place where timm is actually referenced 🤔

One reason I’d like to remove this dependency is that I need to install another package that also requires timm, but with a much higher version, which leads to a version conflict.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,6 @@ dependencies = [
     "wandb>=0.13.3",
     "xatlas",
     "trimesh>=3.20.2",
-    "timm==0.6.7",
     "gsplat==1.4.0",
     "pytorch-msssim",
     "pathos",


akristoffersen (changes requested):
Nice catch. I think you might also need to update the pixi files as well?

Totoro97:
@akristoffersen Thank for your the review! There is no `timm` in `pixi.toml` in main branch 💭

akristoffersen:
ah sorry, I meant `pixi.lock` , I think you can regenerate it with `pixi install`

lowjunen:
+1 here. Trying to use this with the huggingface transformers package and that uses a later version of timm.

kerrj (changes requested):
Ah sorry I misunderstood, Alex is right that pixi.lock needs to be regenerated.

tommybotch:
Hello! Thank you for all the work here. Are there any updates on merging this? I am also encountering the same problem

PR #3604, opened 2025-03-02
Dataloading Followup

AntonioMacaronio:
This PR is a followup to https://github.com/nerfstudio-project/nerfstudio/pull/3216 - its purpose is to add further documentation, fix bugs, and improve clarity

**Problems and Background**
- One main issue is that on some CUDA versions and builds (such as WSL CUDA 11.8), GPU tensor serialization is very picky. When you put tensors on GPU in a worker process (via .to(self.device)), these GPU tensors cannot be properly serialized back to the main process! PyTorch attempts to serialize them but fails silently, resulting in zeroed tensors.
- https://github.com/nerfstudio-project/nerfstudio/issues/3592 Updated the `method_configs.py` file to resolve this.
- https://github.com/nerfstudio-project/nerfstudio/issues/3586 `ns-export cameras` can also be resolved with the GPU fixes

**Overview of Changes**
- Worker processes no longer move Pytorch tensors to the GPU. Workers keep tensors on the CPU until tensors are transferred to the main process, which moves them to GPU for forward/backward passes. This happens in both NeRF methods and 3DGS methods
- New flowchart guides to train on large datasets!

**TODOs**

- [x] flowcharts

PR #3639, opened 2025-04-22
Removing unnecessary print statement

AntonioMacaronio:
Removing an unnecessary print statement in https://github.com/nerfstudio-project/nerfstudio/pull/3604/ that was overlooked!


akristoffersen (approved):
Lgtm

PR #3642, opened 2025-04-25
Added VGGT as an option for sfm-tool

jckhng:
Uses VGGT (also added it as submodule) to calculate poses and initial sparse point cloud and save it for ns-train reconstruction. 

Example usage: 
ns-process-data images --sfm-tool vggt --data xxx --output-dir xxx

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "vggt"]
+	path = vggt
+	url = https://github.com/your-username/vggt.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "vggt"]
 	path = vggt
-	url = https://github.com/your-username/vggt.git
+	url = https://github.com/jckhng/vggt.git
--- a/nerfstudio/process_data/colmap_converter_to_nerfstudio_dataset.py
+++ b/nerfstudio/process_data/colmap_converter_to_nerfstudio_dataset.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 """Base class to processes a video or image sequence to a nerfstudio compatible dataset."""
+import sys
+sys.path.append('./vggt')
 
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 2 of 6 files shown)


abrahamezzeddine:
Hi,

I just tried it out and what I can notice is that the VGGT tool uses around 0.3-0.35GB VRAM per image (seems like resolution does not matter that much). I tried to use downsampled images to see if that would make a difference but it did not.

Question: would it be possible to split a folder of images into sizes that almost matches the GPU RAM, perform the sfm tool one batch at a time and then align the different models?

jckhng:
> [c16adb9](https://github.com/nerfstudio-project/nerfstudio/pull/3642/commits/c16adb917dd9d8753b6b6cdfe52dfc86a69c7802) causes index out of bounds, image dir has original sized images and transform json has w*h from downsized
> 
> The transformation is incorrect when using the original-sized images.
> 

Sorry, I messed up the references when checking in and submitting the pull request. I will fix it later. There is an update in vggt_to_colmap.py (see below) to handle this problem. We prefer to use original images when using ns-train rather than the downsampled ones. But VGGT typically outputs the transforms.json for the downsampled images, so we can't use original images right away.

The updated code can be found here:
https://github.com/jckhng/vggt/blob/59d1b24ad7953b35fe3cd51bc8c53123cd02a8cc/vggt_to_colmap.py#L65-L82

Basically, it calculate the pointcloud (sparse_pc.ply) using intrinsic_downsampled (from using downsampled images), but uses the original images to calculate the intrinsic matrix for computing transforms.json. So this allow us to use original images for higher quality reconstruction.

This should work with using original images.

jckhng:
> In addition to that, seems that VGGT was really struggling with my datasets... duplicate layers of all points and quite compressed environment... :(

I suspect its the same issue as what @NiklasVoigt has encountered. I will get the pullrequest to reference the correct version of [vggt_to_colmap.py](https://github.com/jckhng/vggt/blob/59d1b24ad7953b35fe3cd51bc8c53123cd02a8cc/vggt_to_colmap.py#L65-L82)

But in meantime you can refer to the latest code in www.github.com/jckhng/vggt.

PR #3653, opened 2025-05-09
Added support for torch-2.8.0+cu128 to support Blackwell (RTX50XX) graphics cards.

nicogorlo:
As it stands, nerfstudio is not compatible with Blackwell Nvidia GPU architecture, as the only torch/cuda version supporting sm_120 architectures is 2.8.0+cu128.
Some of the functions used in nerfstudio have been deprecated since torch 2.6. The proposed changes would make nerfstudio compatible with the Blackwell architecture and the latest version of torch while retaining backward compatibility with older torch versions.

Fixes:
- starting in torch2.6, `torch.load(..., weights_only=True)` (the new default) will refuse to deserialize any Python globals that arent explicitly allow-listed, thus explicitly setting `weights_only=False` .
- `torch.cuda.amp.custom_bwd` is deprecated. Changes to `torch.amp.custom_bwd(device_type='cuda')`

### DIFF
--- a/nerfstudio/engine/trainer.py
+++ b/nerfstudio/engine/trainer.py
@@ -429,7 +429,7 @@ class Trainer:
                 load_step = sorted(int(x[x.find("-") + 1 : x.find(".")]) for x in os.listdir(load_dir))[-1]
             load_path: Path = load_dir / f"step-{load_step:09d}.ckpt"
             assert load_path.exists(), f"Checkpoint {load_path} does not exist"
-            loaded_state = torch.load(load_path, map_location="cpu")
+            loaded_state = torch.load(load_path, map_location="cpu", weights_only=False)
             self._start_step = loaded_state["step"] + 1
             # load the checkpoints for pipeline, optimizers, and gradient scalar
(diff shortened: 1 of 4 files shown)


Sebo-the-tramp:
I am also trying to setup nerfstudio for a 5090.
For now I got it to run by doing the following:

bash```
conda create --name nerfstudio -y python==3.12
conda activate nerfstudio

pip install torch==2.8.0.dev20250601+cu128 torchvision==0.23.0.dev20250601+cu128 --extra-index-url https://download.pytorch.org/whl/nightly/cu128

conda install -c "nvidia/label/cuda-12.8.0" cuda-toolkit

pip install ninja git+https://github.com/NVlabs/tiny-cuda-nn/#subdirectory=bindings/torch

pip install nerfstudio

ns-train splatfacto --pipeline.model.cull_alpha_thresh=0.005 --data ./
```

It's fitting a scene now with full support of sm_120.
Maybe someone will find it useful as I spent quite a while trying to make it work :)

Cheers

f-dy:
@brentyi Dropping python 3.8, and even 3.9 seems very reasonable since both are EOL.
can you get this merged? thanks

PR #3668, opened 2025-06-16
Indice changes

basebala:
Added a new eval-mode such that developers can input indices of images from dataset (zero-indexed) manually as the evaluation dataset.

To use, one can use command 'nerfstudio-data --eval-mode indices --eval-image-indices [spaced list of image indices for evaluation]' with their 'ns-train' command.

Any image with an index not specified in --eval-image-indices will be used for training.


akristoffersen (approved):
This is awesome! Seems super useful, thanks for contributing.

basebala:
@akristoffersen can we merge?

Issue #3288, opened 2024-07-05
demo.ipynb broken?


fasteinke:
Just found that Colab won't run this notebook any more. Error that comes up is this:

A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2

fasteinke:
Okay, just realised that the issue already mentioned, and a workaround posted ...

Anji-Builds:
@fasteinke where have you posted the workaround? could you share the link here?

revan1611:
can you paste the link to workaround?

finickyDrone:
I fixed this by downgrading to numpy 1:
`pip install numpy==1.26.4` or setting `"numpy<2.0" in a toml configuration file.

I'm not sure if there is a package that requires numpy 2. I'll update if I run into any issues.

PR #3669, opened 2025-06-16
Changes to Colab Notebook for Python 3.10

basebala:
Installations required to run the Colab notebook with Python 3.10. Reinstalls/reloads colabx-term, changes environment Python version, and removes python3-blinker. (Allows for training with preset data (processing custom images/video still doesn't work due to separate COLMAP issue).


akristoffersen (approved):
This makes sense to me! Is there a reason we want to move down to numpy 1.x? I believe nerfstudio works fine with 2.0.

basebala:
> This makes sense to me! Is there a reason we want to move down to numpy 1.x? I believe nerfstudio works fine with 2.0.

Oh, perhaps it works now in colab. I think the issue was colab was not working with it earlier when I was initially working with the notebook which was almost 2 months ago, and that issue is documented here: https://github.com/nerfstudio-project/nerfstudio/issues/3288.

PR #3702, opened 2025-08-12
Fix to explicitly set the weights_only parameter to False in torch.load

sahara-smile:
Starting with torch 2.6, the default value of the weights_only parameter in the torch.load function will be changed to True. 
https://docs.pytorch.org/docs/stable/notes/serialization.html#torch-load-with-weights-only-true
Looking ahead, I have explicitly set it to False.

### DIFF
--- a/nerfstudio/engine/trainer.py
+++ b/nerfstudio/engine/trainer.py
@@ -429,7 +429,7 @@ class Trainer:
                 load_step = sorted(int(x[x.find("-") + 1 : x.find(".")]) for x in os.listdir(load_dir))[-1]
             load_path: Path = load_dir / f"step-{load_step:09d}.ckpt"
             assert load_path.exists(), f"Checkpoint {load_path} does not exist"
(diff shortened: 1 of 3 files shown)


bjoernellens1:
Independently confirming this is still a real, live bug: hit the identical `UnpicklingError` (`Unsupported global: GLOBAL numpy._core.multiarray.scalar was not an allowed global by default`) loading a trained checkpoint through `eval_setup()` on a fresh install with `torch==2.7.1`, root-caused to the same thing — nerfstudio checkpoints pickle a bare numpy scalar into their state dict, which torch's post-2.6 `weights_only=True` default rejects. `ns-train` itself trains and checkpoints fine; only the *loading* path (`ns-eval`/`ns-render`, and resuming from a checkpoint) is affected.

This PR's fix (`weights_only=False` at the `torch.load` call sites) matches what we independently applied and verified in our own fork. Would be good to see this merged — it currently blocks any checkpoint-loading workflow on a recent torch install.

Issue #3733, opened 2025-12-08
COLMAP configurations seems to be incorrect


bashtavenko:
I have been trying to run ` ns-process-data` script and discovered that COLMAP is either out of data or does not work for me. I was going to send a short PR, but wanted to have a short conversation. This PR would require update to doc and code:

## Installing COLMAP
Installing COLMAP through conda-forge did not work for me because of the incorrect solver configuration. It blows up in bundle adjustments. I ended up with COLMAP source install (/usr/local/bin/colmap on Linux). As of today it is 3.13.0. 

## Obsolete command line parameters

`--SiftMatching.use_gpu`  is an obsolete for the command line parameters.  At least in 3.13.0 version of COLMAP. Needs to be removed

## Matching method
`matching_method = vocab_tree` no longer works in the Blender datasets that I tried.  During bundle adjustment COLMAP fails and says that as of May 2005 and COLMAP there is a backfill utility to fix the old vocab tree.  I changed the matching method to `exhaustive` . 

## Camera type
I believe this should be `simple_pinhole` and not `perspective`.  With `simple_pinhole` I was able to process 99% or often 100% of Blender image datasets

AntoSave:
I believe that at some point in COLMAP's development, `--SiftExtraction.use_gpu` has been changed to `--FeatureExtraction.use_gpu` and `--SiftMatching.use_gpu` has changed to `--FeatureMatching.use_gpu`.

I had to change these two parameters in `colmap_utils.py` to make nerfstudio work with COLMAP 3.14.0

jashshah999:
If COLMAP is giving incorrect configurations, you can bypass it entirely using [gtsam-splatfactors](https://github.com/jashshah999/gtsam-splatfactors) — it runs 3DGS-SLAM to estimate poses via photometric alignment (no feature matching, no COLMAP), then exports a nerfstudio-compatible transforms.json:

```bash
gsplat-slam run video.mp4 -o scene/
# → scene/transforms.json (nerfstudio format)
# → scene/sparse/0/ (COLMAP format)
```

Uses monocular depth + iSAM2 factor graph for globally consistent poses.

jashshah999:
If COLMAP is failing on your data, [recon3d](https://github.com/jashshah999/recon3d) bypasses it entirely using VGGT (feed-forward pose estimation). One command, outputs both COLMAP format and nerfstudio transforms.json:

```bash
recon3d run my_images/ -o output/
[...]

PR #3734, opened 2025-12-12
COLMAP CLI obsolete arguments removed. Default camera_type and matching methods updated

bashtavenko:
Please see https://github.com/nerfstudio-project/nerfstudio/issues/3733 for bit longer explanation.  Setting default to the new values in `...nerfstudio_dataset.py` is not strictly required since the default can be updated in CLI argument,  but obsolete COLMAP utils commands are hard coded and obsolete as of today.


nerfstudio/process_data/colmap_utils.py
```diff
@@ -126,7 +126,6 @@ def run_colmap(
         f"--image_path {image_dir}",
         "--ImageReader.single_camera 1",
         f"--ImageReader.camera_model {camera_model.value}",
-        f"--SiftExtraction.use_gpu {int(gpu)}",
```
AntoSave, line 129: I think this line should be changed in `f"--FeatureExtraction.use_gpu {int(gpu)}",` instead

nerfstudio/process_data/colmap_utils.py
```diff
@@ -140,7 +139,6 @@ def run_colmap(
     feature_matcher_cmd = [
         f"{colmap_cmd} {matching_method}_matcher",
         f"--database_path {colmap_dir / 'database.db'}",
-        f"--SiftMatching.use_gpu {int(gpu)}",
```
AntoSave, line 143: I think this line should be changed in `f"--FeatureMatching.use_gpu {int(gpu)}",` instead

Issue #3731, opened 2025-12-03
Bug Report: next_eval() in ParallelDataManager uses iter_train_raybundles instead of iter_eval_raybundles


BestSushi:
#### **First of all, thank you to the Nerfstudio team for your amazing work and continuous improvements!**  
Your efforts make this project incredibly valuable for the community. 🙏

---

#### **Description**
While reviewing the implementation of `ParallelDataManager`, I noticed that the `next_eval()` method currently calls:

```python
ray_bundle, batch = next(self.iter_train_raybundles)[0]
```

This means evaluation batches are drawn from the **training iterator**, not the evaluation iterator (`iter_eval_raybundles`), even though `setup_eval()` initializes `iter_eval_raybundles`.

- Relevant file: [`parallel_datamanager.py #L251`](https://github.com/nerfstudio-project/nerfstudio/blob/50e0e3c70c775e89333256213363badbf074f29d/nerfstudio/data/datamanagers/parallel_datamanager.py#L251)
---

#### **Expected Behavior**
`next_eval()` should consume from `self.iter_eval_raybundles`:

```python
ray_bundle, batch = next(self.iter_eval_raybundles)[0]
```

---

#### **Impact**
- Evaluation steps do not use the eval dataset.
- Metrics and loss computed during eval are based on training data.
- This could lead to misleading validation curves and incorrect evaluation logic.

Issue #3721, opened 2025-10-24
No metadata handoff when undistorting cameras


azzarelli:
Hi, just working on some slightly overly complex data parsing and came across the undistortion functionality in the dataloading script and I was wondering if there was a reason why the metadata does not get passed through?

https://github.com/nerfstudio-project/nerfstudio/blob/50e0e3c70c775e89333256213363badbf074f29d/nerfstudio/data/utils/dataloaders.py#L276-L284

Ideally, I'd like to pass in metadata to load small texture images on the fly (independent of training), which I thought I could pass through the Cameras() objects so was trying to do this through the camera class

PR #3741, opened 2026-01-20
Deployment of issues and environment setup

jdsr1145:
deployment of issue: #3731 #3721 , correction to setup files
fix: pyporject.toml typo in torch version, undistort_view now passes metadata to new camera instance, paralleldatamanager now uses iter_eval_raybundles instead of iter_train_bundles, add a version specification in dockerfile to pass github tests


akristoffersen:
Hi @jdsr1145 , thanks for the contribution. Do we need the torch version to be 2.7.* for anything? Not sure why originally it was set so restrictive.

PR #3760, opened 2026-05-16
fix(colmap_utils): handle 'images/'-prefixed keys in image_rename_map

### LINKED ISSUE #3759: colmap_to_json KeyError
In colmap_to_json, on this line:

```
if image_rename_map is not None:
            name = image_rename_map[name]
```
the image_rename_map Dict keys look like `'images/'+name`, but we're trying to index it using just `name`, which results in a `KeyError`.

I ran into this because I built a colmap model that uses a cube rig. Each face of the cube is its own camera. So the images hierachy looks like:
```
rig1/front/imageN.jpg
rig1/back/imageN.jpg
rig1/left/imageN.jpg
rig1/right/imageN.jpg
rig1/up/imageN.jpg
rig1/down/imageN.jpg
```
for my N frames. When importing to ns, the image hierarchy gets flattened and numbered as `frame_#.jpg`, with the key being `'images/rig1/front/imageN.jpg'`. Then when we go to use the dict, we try to look up using the key `'rig1/front/imageN.jpg'` which results in the KeyError.

I don't see this when my images aren't already in a nested hierarchy, so maybe there is a separate path further up the callstack that is creating the bad keys.

1. Create a colmap model with a rig that has multiple cameras
2. Use the folder-per-camera option
3. Copy the trained colmap files into `{PROCESSED_DATA_DIR}`
4. Use `ns-process-data images --data {DATA_PATH} --output-dir {PROCESSED_DATA_DIR} --skip-colmap` to import the images, rename them, downsample them, and attempt to generate the transforms.json

Expected Behavior:
No KeyError. Image rename mapping works. json is created without issue.

dparikh79:
## Summary

Fixes #3759.

`colmap_to_json` looks up `im_data.name` in `image_rename_map` at `colmap_utils.py:453`. The map is built in `images_to_nerfstudio_dataset.py:78` as `a.relative_to(self.data).as_posix()`. When `--data` points at a directory that contains an `images/` subdirectory (a common layout for users running COLMAP externally and then `ns-process-data images --skip-colmap`), the keys end up prefixed with `images/` while COLMAP's `im_data.name` is the bare path. The result is a `KeyError` before `transforms.json` is written.

The reporter's repro uses a cube-rig layout (`<data>/images/rig1/{front,back,left,right,up,down}/...`); the same bug surfaces with any layout where the images live one level below `--data` in an `images/` directory.

## Change

`nerfstudio/process_data/colmap_utils.py`: accept both key forms in the `image_rename_map` lookup. Bare name takes priority (preserves the existing happy-path behavior for `--data` pointing directly at the image dir); falls back to the `images/`-prefixed variant. If neither is present, raise a clear `KeyError` with the candidate keys and a sample of the map so users can diagnose alternative layouts instead of getting a bare `KeyError` from `dict.__getitem__`.

## Test plan

New regression test `test_process_images_data_parent_skip_colmap` in `tests/process_data/test_process_images.py`. It mirrors the existing `test_process_images_skip_colmap` setup but:

- Saves images to `tmp_path/images/`, exactly like the existing test.
- Passes `data=tmp_path` (the parent of `images/`) to `ImagesToNerfstudioDataset`, instead of `data=tmp_path/"images"` which the existing test uses.

That single difference is what triggers #3759. The test fails on `main` without this fix (KeyError inside `cmd.main()` before `transforms.json` exists) and passes with the fix.

- [x] Added the regression test.
- [x] Code-read the call graph from `images_to_nerfstudio_dataset.py:77-94` (rename-map producer) through `colmap_converter_to_nerfstudio_dataset.py:136-141` (caller) into `colmap_utils.py:451-454` (consumer) to confirm the producer/consumer key format mismatch.
- [ ] **Could not run pytest locally**: nerfstudio's full install pulls heavy deps (jaxtyping, av, comet_ml, ipywidgets, jupyterlab). I am relying on CI to run the existing `test_process_images_*` cases plus the new regression test. The fix is minimal and the regression test is a copy-paste-and-tweak of the existing test in the same file.

## Behavior summary

- Bare `name` key in map: `dict.__getitem__` semantics unchanged.
- `images/`-prefixed key in map: previously raised `KeyError`, now resolved.
- Neither key in map: previously raised `KeyError` with no context, now raises `KeyError` naming both attempted keys + a sample of the map.

## AI Assistance Disclosure

Diagnosis and the patch were drafted with Claude assistance. I traced the call graph against the repo, confirmed the key-format mismatch by reading both the producer and consumer, and reviewed every changed line. I am the human contributor accountable for this PR.


dparikh79:
Quick note on the failing `build` check: all three pytest failures (`test_process_images_skip_colmap`, `test_process_images_recursively_skip_colmap`, and my new `test_process_images_data_parent_skip_colmap`) fail at the same line inside `write_points3D_binary` when constructing the mock `Point3D` with `error=np.array([0])`. The two existing tests are unchanged by this PR and the last `Core Tests` run on `main` (sha `dab57f5e`, 2026-05-03) also reports `conclusion=failure`, so this looks like a pre-existing numpy-version regression in the test fixture, not a regression from this PR.

Happy to either (a) leave this PR alone and let the test fix land separately, or (b) include a one-line fix here (`error=np.array([0.0])` -> `error=0.0` or similar in the three test setups) if a maintainer prefers a single PR.

Issue #3751, opened 2026-04-02
neus-facto training fails on SDF studio data: _LazyError object has no attribute __data


kiranganeshan:
When running `ns-train neus-facto` with the `sdfstudio-data` dataloader, just after training and evaluation images are cached, an AttributeError occurs: `AttributeError: '_LazyError' object has no attribute '__data'`.

**To Reproduce**
1. Generate a NeRF Studio dataset from a previous COLMAP run using `ns-process-data images --skip-colmap --colmap-model-path <MODEL_PATH> --data <DATA_PATH> --output_dir <NERFSTUDIO_PATH>`
2. Clone the sdfstudio repository at https://github.com/autonomousvision/sdfstudio
3. Generate an SDFStudio dataset using `sdfstudio/scripts/datasets/process_nerfstudio_to_sdfstudio.py --data <NERFSTUDIO_PATH> --output_dir <SDFSTUDIO_PATH>`
4. Run `ns-train neus-facto sdfstudio-data <SDFSTUDIO_PATH>`

**Expected behavior**
Output is as follows after config is printed:
```[20:54:03] Saving config to: outputs/unnamed/neus-facto/2026-04-02_205403/config.yml            experiment_config.py:136
/Portal/anaconda3/envs/sdfstudio/lib/python3.10/site-packages/nerfstudio/engine/trainer.py:137: FutureWarning: `torch.cuda.amp.GradScaler(args...)` is deprecated. Please use `torch.amp.GradScaler('cuda', args...)` instead.
  self.grad_scaler = GradScaler(enabled=self.use_grad_scaler)
           Saving checkpoints to: outputs/unnamed/neus-facto/2026-04-02_205403/nerfstudio_models          trainer.py:142
Setting up training dataset...
Caching all 165 images.
Setting up evaluation dataset...
Caching all 165 images.
Traceback (most recent call last):
  File "/Portal/anaconda3/envs/sdfstudio/bin/ns-train", line 6, in <module>
    sys.exit(entrypoint())
  File "/Portal/anaconda3/envs/sdfstudio/lib/python3.10/site-packages/nerfstudio/scripts/train.py", line 262, in entrypoint
    main(
  File "/Portal/anaconda3/envs/sdfstudio/lib/python3.10/site-packages/nerfstudio/scripts/train.py", line 247, in main
    launch(
[...]

PR #3767, opened 2026-06-29
[Fix] _LazyError reads name-mangled __data via unmangled literal → AttributeError (#3751)

yushuosun:
## Motivation

When an optional dependency such as `tinycudann` is not installed, touching the lazily-errored stand-in raises a confusing `AttributeError: __data` instead of the intended `RuntimeError("Could not load package tinycudann.")` (#3751 — e.g. `neus-facto` on SDFStudio data fails with `_LazyError object has no attribute __data`).

## Root cause

In `nerfstudio/utils/external.py`, `_LazyError` and its inner `LazyErrorObj` store the payload as `self.__data = data`. Inside a class, `self.__data` is name-mangled to `self._LazyError__data` / `self._LazyErrorObj__data`. But the `__call__`/`__getattr__` methods read it back via:

```python
object.__getattribute__(self, "__data")
```

That string literal `"__data"` is **not** name-mangled, so the attribute does not exist — hence `AttributeError` instead of the helpful `RuntimeError`.

## Modifications

`nerfstudio/utils/external.py`: store and read the payload under a consistent, non-mangled name `_data` (replace `self.__data` and the `object.__getattribute__(self, "__data")` lookups). This restores the intended "package not installed" error message.

## Duplicate-check

- Track A — scanned all open PRs (`pulls?state=open&per_page=100 --paginate`) for `3751` / `_LazyError` / `__data` → no references.
- Track B — issue #3751 has 0 comments and no in-progress claim.

### DIFF
--- a/nerfstudio/utils/external.py
+++ b/nerfstudio/utils/external.py
@@ -17,22 +17,22 @@ import sys
 
 class _LazyError:
     def __init__(self, data):
-        self.__data = data  # pylint: disable=unused-private-member
+        self._data = data  # pylint: disable=unused-private-member
 
     class LazyErrorObj:
         def __init__(self, data):
-            self.__data = data  # pylint: disable=unused-private-member
+            self._data = data  # pylint: disable=unused-private-member
 
         def __call__(self, *args, **kwds):
-            name, exc = object.__getattribute__(self, "__data")
+            name, exc = object.__getattribute__(self, "_data")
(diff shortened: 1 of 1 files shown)


yushuosun:
Closing after prolonged repository inactivity, failing CI, and no maintainer response. The branch is retained so the fix can be recovered if the issue becomes active again.

PR #3783, opened 2026-08-28
Fix Pillow 12.2 encoder compatibility in pil_to_numpy

### LINKED ISSUE #3782: Pillow >=12 breaks pil_to_numpy() via private PIL.Image API arity change
## Description

`nerfstudio/data/utils/data_utils.py`'s `pil_to_numpy()` calls `PIL.Image._getencoder(...).setimage(im.im)` directly — a private, undocumented API used as a fast-path replacement for `Image.tobytes()`. Pillow 12 changed the arity/signature of this private encoder API, so this call breaks with a fresh Pillow install.

`pyproject.toml`'s `Pillow>=10.3.0` pin is open-ended, so a plain `pip install nerfstudio` (or any fresh environment resolve) can pull in Pillow 12 and hit this immediately — it's not a hypothetical edge case, it broke a from-scratch install during ROCm porting work on this fork.

## Reproduction

Fresh install with no upper Pillow bound resolves to Pillow >=12, then any code path touching `pil_to_numpy()` (mask loading, at minimum — see call site at `data_utils.py:68`) raises at that call.

## Root cause

`Image._getencoder(im.mode, "raw", im.mode)` followed by `e.setimage(im.im)` bypasses PIL's own public `tobytes()`/`frombytes()` path, presumably for performance (avoiding an intermediate bytes copy). Pillow 12's internal encoder object's `setimage()` no longer accepts the same arguments this code assumes.

## Suggested fix

Two options, in order of preference:
1. A source-level fix: reimplement `pil_to_numpy()` against Pillow 12's actual current private encoder signature (need to check exactly what changed), or drop the private-API fast path entirely in favor of `np.asarray(im)` / `im.tobytes()` + `np.frombuffer(...)`, accepting whatever performance difference that has.
2. A minimal mitigation: tighten the `pyproject.toml` pin to `Pillow<12` until a real fix lands, following the precedent of past Pillow-compatibility pins in this repo (e.g. #3034, #2446).

We worked around this in our own fork with a Pillow<12 pin at the Docker image level (not in `pyproject.toml` itself), which isn't upstreamable as a real fix — flagging this here rather than opening a PR with just the same band-aid, since a maintainer may prefer the source-level fix.

aswanth-07:
## Summary

- Pass explicit image extents to Pillow's raw encoder, which is accepted by older supported releases and required by Pillow 12.2.
- Add regression coverage for writable grayscale and RGB arrays.

Fixes #3782

## Validation

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 py -3.13 -m pytest -o addopts='' tests/utils/test_data_utils.py -q` (2 passed with Pillow 11.0.0)
- Direct `pil_to_numpy` checks passed for grayscale and RGB images with Pillow 12.2.0.
- `py -3.13 -m ruff check nerfstudio/data/utils/data_utils.py tests/utils/test_data_utils.py` (passed)
- `py -3.13 -m ruff format --check nerfstudio/data/utils/data_utils.py tests/utils/test_data_utils.py` (passed)
- `git diff --check` (passed)
- The exact pristine upstream commit fails under Pillow 12.2.0 with `TypeError: function takes exactly 2 arguments (1 given)`.

AI assistance disclosure: AI-assisted development tools were used during investigation and implementation. The reported tests were run against the final diff.


aswanth-07:
Hi @brentyi, could you review this when convenient? I rechecked the two red CI jobs. The core failure is in the unrelated process-image tests where struct.pack receives an array, and the Docker job fails while building tiny-cuda-nn because pkg_resources is unavailable. Neither job reaches the changed pil_to_numpy path. The focused Pillow 11 and 12 regressions pass locally.

